PR #13, opened 2023-07-10
Wrap all peer-loop execution in `catch_unwind`

Sword-Smith:
This should ensure that the peer map is always up-to-date even if a peer-loop panics, which shouldn't be possible.

### DIFF
--- a/src/connect_to_peers.rs
+++ b/src/connect_to_peers.rs
@@ -1,5 +1,5 @@
 use anyhow::{bail, Result};
-use futures::{SinkExt, TryStreamExt};
+use futures::{FutureExt, SinkExt, TryStreamExt};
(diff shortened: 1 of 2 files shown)


ps1dr3x:
Are you aware of specific non unwind safe cases in the code?

I probably wouldn't have been so cautious to do it otherwise, but the code looks good to me and I don't think it would cause any harm, so why not (especially since it's done already 😄).

ps1dr3x (approved):

Sword-Smith:
> Are you aware of specific non unwind safe cases in the code?

I actually can't answer that as I don't really know what it means :/

I was just thinking that I wanted to provide a callback function in case the `peer_loop` thread panicked. Such that the part of the global state that keeps track of connected peers can be correctly updated.

PR #34, opened 2023-08-09
improve continuous integration

jan-ferdinand:


Sword-Smith:
CI works on all three OSes when you exclude benches, which apparently use too much RAM for GitHub's Windows build machines.

jan-ferdinand:
Ah, that's what that obscure error was! Thanks for figuring that out. :blush:

Issue #43, opened 2023-09-17
Impl tab completion (shell completion) for neptune-cli


dan-da:
It would be great if tab completion worked for commands.

I could possibly get this working and submit a PR...

PR #44, opened 2023-09-19
add shell tab completions feature to neptune-cli

dan-da:
addresses #43

Add `completions` sub-command that prints out completion code for present shell. Eg in bash do:

    neptune-cli completions > ~/.bash_neptune_cli
    echo "source ~/.bash_neptune_cli" >> ~/.bashrc

Supports bash, elvish, fish, powershell, zsh

Adds dependency on crate [clap_complete](https://docs.rs/clap_complete/latest/clap_complete/).

Notes:
* The present implementation is intended to make neptune-cli more ergonomic during development.
* `clap_complete` also supports compile time generation of completion shell scripts using build.rs.  Useful when installing binary for end user.  That solution is a little more complex, perhaps in the future and/or we can discuss if its worth doing in this PR.
* The `clap` project is working on a completion overhaul that would enable dynamic shell completion by invoking the binary from the shell with a special hidden flag.  That should be a better solution once available.  See:
  https://github.com/clap-rs/clap/issues/1232 and 
  https://github.com/clap-rs/clap/issues/3166
* We could possibly hide this behind a `completions` feature flag to enable building without the clap_complete crate.   Please let me know if this is desired.
* I did not add completions for other binaries such as neptune-core.  I figured I would see how this PR is received first.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -472,6 +472,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete"
+version = "4.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4110a1e6af615a9e6d0a36f805d5c99099f8bab9b8042f5bc1fa220a4a89e36f"
+dependencies = [
+ "clap",
+]
+
+[[package]]
 name = "clap_derive"
 version = "4.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1303,6 +1312,7 @@ dependencies = [
  "cargo-husky",
  "chrono",
  "clap",
+ "clap_complete",
  "crossterm",
  "directories",
  "field_count",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,6 +57,7 @@ triton-vm = "0.33.0"
 twenty-first = "0.32.1"
 tui = "0"
 unicode-width = "0"
+clap_complete = "4.4.1"
 
 [dev-dependencies]
 pin-project-lite = "0.2.12"
--- a/install-bash-completions.sh
+++ b/install-bash-completions.sh
@@ -1,0 +1,9 @@
+SCRIPT="$HOME/.bash_neptune_cli"
+LINE="source $SCRIPT"
+FILE="$HOME/.bashrc"
+
+cargo run --bin neptune-cli -- completions > $SCRIPT && \
+grep -qF -- "$LINE" "$FILE" || echo "$LINE" >> "$FILE" && \
+source $SCRIPT && \
+echo "completions installed to $SCRIPT and added to $FILE"
+
--- a/src/bin/neptune-cli.rs
+++ b/src/bin/neptune-cli.rs
@@ -1,5 +1,6 @@
-use anyhow::Result;
-use clap::Parser;
+use anyhow::{bail, Result};
+use clap::{CommandFactory, Parser};
+use clap_complete::{generate, Shell};
 
 use neptune_core::config_models::network::Network;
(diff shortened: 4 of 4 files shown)


Sword-Smith:
I tested it out and got it to work on my machine that runs bash. Thank you for this!

One question though: There was a process of manual setup to get the tab-completion to work (the two commands you provide above), and I guess you would have to rerun that command if the CLI interface changes. I'm not much of a Unix/bash expert but is this how it's normally done? Is there a way to avoid this manual step?

Just to clarify, I consider this PR an improvement despite the manual setup process and am happy to merge.

dan-da:
Yeah, I find the present situation with clap completions sub-optimal.   Hopefully they will finish up the dynamic completions by the time neptune launches.  That provides both more functionality as well as it should always stay current with the binary without any install step.   Otherwise, the build.rs method provides a way to generate the shell scripts at compile time without the `completions` runtime sub-command.  They could then be checked in git, installed for end user, etc.   

Still, this method was easy to build with few loc and for us devs install can easily be automated with a script like this:

install-completions.sh
```
LINE='source ~/.bash_neptune_cli '
FILE="$HOME/.bashrc"

cargo run --bin neptune-cli -- completions > ~/.bash_neptune_cli && \
grep -qF -- "$LINE" "$FILE" || echo "$LINE" >> "$FILE"
```

I can add this script to the PR if you like.   Note of course that it is bash specific, so doesn't help for the other supported shells.

Sword-Smith:
> Yeah, I find the present situation with clap completions sub-optimal. Hopefully they will finish up the dynamic completions by the time neptune launches. That provides both more functionality as well as it should always stay current with the binary without any install step. Otherwise, the build.rs method provides a way to generate the shell scripts at compile time without the `completions` runtime sub-command. They could then be checked in git, installed for end user, etc.
> 
> Still, this method was easy to build with few loc and for us devs install can easily be automated with a script like this:
> 
> install-bash-completions.sh
> 
> ```
> LINE='source ~/.bash_neptune_cli '
> FILE="$HOME/.bashrc"
> 
> cargo run --bin neptune-cli -- completions > ~/.bash_neptune_cli && \
> grep -qF -- "$LINE" "$FILE" || echo "$LINE" >> "$FILE"
> ```
> 
> I can add this script to the PR if you like. Note of course that it is bash specific, so doesn't help for the other supported shells.

I'm all in favor of adding scripts to the code base. Please do that. In case you're wondering what @aszepieniec and I are currently working on, it's (the compiler)[https://github.com/TritonVM/tasm-lang] and the (standard library)[https://github.com/TritonVM/tasm-lib] that is required to define the consensus logic that must run in the Triton-VM virtual machine.

Once the compiler is where I want it to be, I'll most likely give neptune-core more attention.

dan-da:
I added the bash install script.

I also took a quick look at what would be involved to use the compile-time generation of completion scripts.   It appears that neptune-cli.rs would need to be split into separate files so that the CLI commands can be included from build.rs without redefining main().  Not really a big deal, but then maybe neptune-cli should get its own directory, etc.  So that's a road we could pursue if you wish and maybe will happen naturally later on anyway as neptune-cli grows, but I didn't feel right about just diving into it on my own for dubious benefit.

here's a writeup about it.
https://kbknapp.dev/shell-completions/

Sword-Smith (approved):
LGTM

Sword-Smith:
I approved this PR. Feel free to merge, @dan-da.

dan-da:
@Sword-Smith I think you will need to merge as I don't have perms.

> Only those with write access to this repository can merge pull requests.

Sword-Smith:
I merged, and then I put the script into a `scripts` directory. One nice-to-have thing would be if the script reported an error if `neptune-core` is not running.

Issue #45, opened 2023-09-19
clippy warnings.


dan-da:
While working on #44 I ran `cargo clippy` and noticed a lot of warnings.   Specifically:

```
$ cargo clippy 2>&1 | grep ^warning | wc -l
20
$ cargo clippy 2>&1 | grep ^warning
warning: useless use of `vec!`
warning: useless use of `vec!`
warning: useless use of `vec!`
warning: useless use of `vec!`
warning: useless use of `vec!`
warning: useless use of `vec!`
warning: useless use of `vec!`
warning: useless use of `vec!`
warning: useless use of `vec!`
warning: useless use of `vec!`
warning: useless use of `vec!`
warning: useless use of `vec!`
warning: useless use of `vec!`
warning: useless use of `vec!`
warning: useless use of `vec!`
warning: `neptune-core` (lib) generated 15 warnings (run `cargo clippy --fix --lib -p neptune-core` to apply 10 suggestions)
warning: usage of an `Arc` that is not `Send` or `Sync`
warning: usage of an `Arc` that is not `Send` or `Sync`
warning: usage of an `Arc` that is not `Send` or `Sync`
warning: `neptune-core` (bin "neptune-dashboard") generated 3 warnings
```

I can work on a PR to fix these.   Hopefully all are simple.

I was surprised to see the warnings as previously I've worked on projects where CI does not allow PRs to be accepted until all clippy warnings are resolved and also cargo fmt --check is clean.

Is that a policy we could consider implementing here?    Keeps things nice and tidy for everyone.

Sword-Smith:
The `warning: useless use of `vec!`` usually means that we are concatenating vectors like this:
`vec![vec![a, b], vec![c]].concat()`. With `rustc` version 1.72.0 this became a clippy warning, as the outermost `Vec<T>` can be replaced by an array. So clippy wants you to write `[vec![a, b], c].concat()` instead.

This issue arose in several repos since the above, deprecated pattern was used a lot in our code.

I would have to do a little more digging into the `Arc`/`Send` issue. I don't know the solution to that from the top of my head.

dan-da:
thx.   I'll take a look at cleaning these up tmw.   Maybe learn a bit more about the codebase in the process.

Sword-Smith:
> thx. I'll take a look at cleaning these up tmw. Maybe learn a bit more about the codebase in the process.

To quickly get an overview of the code, you can start here:
[...]

PR #46, opened 2023-09-20
Fix clippy warnings pr

dan-da:
Addresses #45

* Fixes clippy warnings about unnecessary vec![] and Arc<RefCell<..>>
* Applies cargo fmt

With this patch applied, both these commands run cleanly:

```
$ cargo clippy --all-features --all-targets
$ cargo fmt --check
```

Notes:
* running cargo version `1.72.0 (103a7ff2e 2023-08-15)`
* Changed all occurences of Arc<RefCell<..>> to Arc<Mutex<..>>.  These were only in the Dashboard.

Testing:
* Verified that Dashboard runs and navigated all screens
* ran `cargo test` without error

### DIFF
--- a/src/bin/dashboard_src/history_screen.rs
+++ b/src/bin/dashboard_src/history_screen.rs
@@ -1,5 +1,4 @@
 use std::{
-    cell::RefCell,
     cmp::{max, min},
     sync::{Arc, Mutex},
     time::{Duration, UNIX_EPOCH},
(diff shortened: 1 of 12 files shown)


Sword-Smith:
Yeah. Since the dashboard has a polling thread and is apparently multi threaded, I guess we need `Arc` and not just `Rc` there.

The rest of the commits I consider uncontroversial. Thanks for fixing this though.

Issue #57, opened 2023-10-05
"regtest" hard-coded in log output


dan-da:
I've been seeing lines like this in output of neptune-core:

```
2023-10-05T21:56:38.383155149Z  INFO ThreadId(05) neptune_core::peer_loop: Updated block info by block from peer. block height 600
2023-10-05T21:56:39.474829162Z  INFO ThreadId(05) neptune_core::mine_loop: Mining thread got message from main
2023-10-05T21:56:39.474974901Z  INFO ThreadId(05) neptune_core::mine_loop: Miner thread received regtest block height 600
```

neptune-core is started with the command:

```
RUST_BACKTRACE=1 RUST_LOG='info,tarpc=warn' cargo run --bin neptune-core -- --mine --throttled-mining --peers 51.15.139.238:9798 --peers 139.162.193.206:9798
```

Given that there are peers, I would not think that regtest should not be used.   So I grepped in the code and found:

```
$ rg regtest mine_loop.rs 
361:                        info!("Miner thread received regtest block height {}", latest_block.header.height);
405:                info!("Found new regtest block with block height {}. Hash: {}", new_block_info.block.header.height, new_block_info.block.hash.emojihash());
```

I think that either `regtest` should be removed from these, or replaced with a string variable identifying the current network.

I plan to submit a PR for this soon.

PR #61, opened 2023-10-10
fix hard-coded "regtest" in log output

dan-da:
Addresses #57

Now uses state.cli.network instead of hard-coded string for the two occurrences previously identified in #57.

New (tested) log output:

```
2023-10-10T03:44:50.540462149Z  INFO ThreadId(05) neptune_core::mine_loop: Miner thread received alpha block height 00000000000000005102

```

### DIFF
--- a/src/mine_loop.rs
+++ b/src/mine_loop.rs
@@ -358,7 +358,7 @@ pub async fn mine(
                             mt.abort();
                         }
                         latest_block = *block;
-                        info!("Miner thread received regtest block height {}", latest_block.header.height);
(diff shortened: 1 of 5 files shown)


src/mine_loop.rs
```diff
@@ -358,7 +358,7 @@ pub async fn mine(
                             mt.abort();
                         }
                         latest_block = *block;
-                        info!("Miner thread received regtest block height {}", latest_block.header.height);
```
Sword-Smith, line 361: This looks more like a `debug` log message than an `info` log message to me :)

Sword-Smith (approved):

PR #69, opened 2023-11-18
Multi val db access pr

dan-da:
In general, this PR makes locking around DB accesses (both read and write) less granular.

There is a tradeoff here that became more apparent the further I dug into things.

This *may* make functions that perform a lot of DB accesses a bit more performant.  It *might* lessen occurrence of deadlocks. 

For certain, it:
  1. forces all other threads to block until these functions complete. 
  2. requires allocating memory for all values retrieved from get_many(), get_all().

I have commented the locking issue in the code at each point of usage.   

Our previous approach processing each value at a time might be a bit slower and acquire many more locks, but it is much more concurrent and works with arbitrary number of values without exploding memory for large datasets, eg wallets with thousands of monitored utxo.   For this reason, I am making this a draft PR, and not recommending it be merged until further testing/benchmarking/analysis is conducted.

HOWEVER:  the elephant in the room is the nature of the DB accesses.   The `rusty-leveldb` crate we are using [does not support concurrent accesses](https://docs.rs/rusty-leveldb/latest/rusty_leveldb/):

> "As opposed to the original, this implementation is not concurrent (yet):

Whereas the [leveldb](https://crates.io/crates/leveldb) crate (wrapping C++ leveldb) [does support concurrency](https://github.com/google/leveldb/blob/main/doc/index.md#concurrency).   Also the [DB::get()](https://skade.github.io/leveldb/leveldb/database/struct.Database.html) fn in `rusty-leveldb` requires (&self mut), so we cannot use an `RwLock`, but must instead use a `Mutex`.  Readers block eachother as well as writers.   I believe that if we were to move to the `leveldb` crate we could get rid of the locking around DB calls and the code in this PR would work with good concurrency.  Note also that [`get()` method takes &self](https://docs.rs/leveldb/0.8.4/leveldb/database/kv/trait.KV.html#tymethod.get).  We should not need *any* locks around these db calls (greatly simplifying our code), but if we wanted to make our own RwLock, we could.

Also, there is a [draft PR for twenty-first](https://github.com/Neptune-Crypto/twenty-first/pull/160) that provides iterators for the DB types.  This would enable usage of the get many/all set-based semantics in this PR without blowing up memory for large data sets.  So that is a possibility as well and would work best with the concurrent version of leveldb.

Finally, bolded on the rusty-leveldb [repo index page](https://github.com/dermesser/leveldb-rs):

> **NOTE: I do not endorse using this library for any data that you care about.**

one last clarification:  I find this code a bit cleaner and more readable than what came before, so I hope we can merge it, or something close if/when the above issues are resolved.

commit messages follow:

-----

Author: danda <dan-da@users.noreply.github.com>
Date:   Fri Nov 17 21:01:57 2023 -0800

    feat: use get/set_many, get/set_all() for DB
    
    Use multi-item DB operations instead of locking DB per item
    
    Cargo.toml is modified to override twenty-first dep with
    master from github.
-----

Author: sword-smith <thor@neptune.cash>
Date:   Thu Nov 9 21:57:36 2023 +0100

    Improve lock-handling in method to prune abandoned MUTXOs
    
    Co-authored-by: Dan <danda@neptune.cash>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -54,13 +54,14 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.8.3"
+version = "0.8.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2c99f64d1e06488f620f932677e24bc6e2897582980441ae90a671415bd7ec2f"
+checksum = "91429305e9f0a25f6205c5b8e0d2db09e0708a7a6df0f42212bb56c32c8ac97a"
 dependencies = [
  "cfg-if",
  "once_cell",
  "version_check",
+ "zerocopy",
 ]
 
 [[package]]
@@ -155,9 +156,9 @@ checksum = "a4668cab20f66d8d020e1fbc0ebe47217433c1b6c8f2040faf858554e394ace6"
 
 [[package]]
 name = "arbitrary"
-version = "1.3.1"
+version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a2e1373abdaa212b704512ec2bd8b26bd0b7d5c3f70117411a5d9a451383c859"
+checksum = "7d5a26814d8dcb93b0e5a0ff3c6d80a8843bafb21b39e8e18a6f05471870e110"
 dependencies = [
  "derive_arbitrary",
 ]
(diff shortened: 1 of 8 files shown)


dan-da:
closing in favor of #89, already merged.

PR #72, opened 2023-11-19
feat: support for tokio-console

dan-da:
Adds `--tokio-console` boolean cli arg to neptune-core.

When this flag is present, logging is configured to emit tokio tracing information and send it to localhost:6669 for processing by tokio-console.

This requires building tokio with the `tokio_unstable` build flag. This is specified in the new `.cargo/config.toml`

see: https://github.com/tokio-rs/console

Here's a screenshot of the `resources` view:

screenshot

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,4 @@
+# tokio_unstable compile flag is necessary for tokio-console usage.
+# see: https://docs.rs/console-subscriber/latest/console_subscriber/
+[build]
+rustflags = ["--cfg", "tokio_unstable"]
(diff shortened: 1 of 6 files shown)


Sword-Smith (approved):
This looks exciting :)

Hope we can get rid of unstable down the line though. I hope unstable will work on Windows. If not, then maybe one could handle this with a environment variable that is only set on Unix/Linux.
cf. https://docs.rs/tokio/latest/tokio/#unstable-features

dan-da:
I have my doubts that it will move out of unstable any time soon.   I'm viewing this as a temporary thing and maybe we disable or remove it or improve build configurability before mainnet.

dan-da:
instructions for running tokio-console with neptune-core:

```
# install tokio-console app
cargo install --locked tokio-console

# start neptune-core  (listens on localhost:6669)
neptune-core --tokio-console

# run tokio-console  (attempts to connect to localhost:6669)
tokio-console
```

See https://github.com/tokio-rs/console for more details, screenshots, etc.

PR #78, opened 2024-01-03
Listen to IPv4 and IPv6 ports as default

Sword-Smith:
This should add IPv6 support. That works, but I haven't tested that IPv4 still works, as I haven't had the chance to run it on a machine with a static IP address which is the only way to connect via IPv4.


Sword-Smith:
Merged through da457e23fb4fad84faa60e8cce4d22d0c6c37af9

PR #82, opened 2024-01-09
Tokio locks pr

dan-da:
This PR adds `crate::util_types::sync::tokio::{AtomicRw, AtomicMutex}`.   These are wrappers around `tokio::sync::{RwLock, Mutex}` and complement `twenty_first::sync::{AtomicRw, AtomicMutex}` which wrap the std library lock types.

This PR just introduces the locks without using them in the App yet.

This is extracted from, and is a building block for, a much larger neptune-core PR to follow that reworks lock handling substantially. 

These async tokio locks are very similar to the sync counterparts except that:
1. calls to `lock()` and `lock_mut()` are async, so would typically be followed by `.await`.
2. additional methods `lock_async()` and `lock_async_mut()` are provided.   These enable the caller to pass in an async callback fn.  Unfortunately the usage/ergonomics is a bit awkward (see doctest examples) and has a runtime cost so in practice I have tended to just use `lock_guard()` and `lock_guard_mut()` instead. 

Like the sync types in twenty-first, these types support named locks with lock-event callbacks, which enables detailed loggining/tracing for detecting deadlocks, and lock usage by thread or tokio task.


dan-da:
In the interest of saving time, I merged without a review.    Anyway, this doesn't really change anything, just adds new, as-yet unused types.

Issue #75, opened 2023-12-17
Make File Operations async.  (ArchivalState, etc)


dan-da:
Related to #74.

Presently many, if not all, file operations use synchronous APIs, called from async functions.   For example async fn [get_latest_block()](https://github.com/Neptune-Crypto/neptune-core/blob/45d2ca7d5b7ceedf7a85ff6e885b88f323fd2b9c/src/models/state/archival_state.rs#L388) eventually calls [get_block_from_record](https://github.com/Neptune-Crypto/neptune-core/blob/45d2ca7d5b7ceedf7a85ff6e885b88f323fd2b9c/src/models/state/archival_state.rs#L340) which opens a file and reads it in via (unsafe) mmap calls.

These sync operations will block the tokio executor which can result in poor performance/concurrency.  So in general, it should be more performant to convert the file ops to [async counterparts provided by tokio](https://docs.rs/tokio/latest/tokio/fs/index.html), or at least wrap them with spawn_blocking.

note: I'm unsure about making the mmap calls async; an area for research.

dan-da:
The only remaining blocking I/O that I'm aware of is in twenty-first storage.    To fix that, we need to add a tokio dep in twenty-first, or move storage out of twenty-first somehow.

It's also possible there is blocking I/O in triton_vm or tasm-lib, but I'm hoping not....    ;-)

dan-da:
closed by #124

Issue #74, opened 2023-12-14
Make RustyLevelDB API async.


dan-da:
LevelDB provides a synchronous API.

RustyLevelDB is a thin wrapper around LevelDB API.   It also provides [sync methods](https://github.com/Neptune-Crypto/neptune-core/blob/45d2ca7d5b7ceedf7a85ff6e885b88f323fd2b9c/src/database/rusty.rs#L61).

In neptune-core, we have async API like [`get_block()`](https://github.com/Neptune-Crypto/neptune-core/blob/45d2ca7d5b7ceedf7a85ff6e885b88f323fd2b9c/src/models/state/archival_state.rs#L463), `get_block_header()`, etc that *appear* to be async.  However internally they are calling these blocking DB APIs.

This creates a hidden source of blocking/contention that may not be obvious until operating under load, and suddenly performance becomes very bad one day.  [See this description](https://internals.rust-lang.org/t/warning-when-calling-a-blocking-function-in-an-async-context/11440/5).

note: The RustyLevelDb is presently wrapped in a tokio lock, which requires await to acquire the lock.   This gives an appearance of async, however the actual DB calls such as `get()`, `set()`, `delete()` remain blocking calls and may potentially block the calling task for database disk I/O.

note: I'm in the process of removing this tokio lock, which is no longer necessary because the C++ leveldb does its own locking.   While refactoring, this issue made itself clear.

A simple solution could be to make the RustyLevelDb methods async.   Internally, they would wrap the DB calls inside [spawn_blocking](https://docs.rs/tokio/1.35.0/tokio/task/fn.spawn_blocking.html).  This enables the RustyLevelDb method to await the DB operation (and return a Future immediately).

> In general, issuing a blocking call or performing a lot of compute in a future without yielding is problematic, as it may prevent the executor from driving other futures forward. This function runs the provided closure on a thread dedicated to blocking operations.

dan-da:
I've created a `RustyLevelDbAsync` struct that provides async versions of all methods in `RustyLevelDB`, even `new()`, which creates/opens a DB.

For now this lives in neptune-core, on a branch in my forked repo.   But that's probably not where it belongs, because...

This async issue also affects `twenty_first::storage`.   The LevelDB wrapper in there and all the `StorageVec`, `DbtSchema` stuff is synchronous, and will need to be made async.

[...]

PR #89, opened 2024-01-13
feat: global state lock, async leveldb, etc.

dan-da:
First off, I apologize for the size of this PR.   I normally keep PRs as small/focused as I can.

I started a branch while working on twenty_first storage in order to test those changes in neptune.   As I iterated on storage that led to the AtomicRw lock in twenty-first, which led to tokio AtomicRw in neptune.  At first I was replacing and refactoring (encapsulating) the existing locks and I wanted to be able to log/trace their activity, so I added a facility for that.   As I dug deeper I found that we still couldn't really guarantee concurrent write serialization at the highest level ie across Blockchain, Mempool, Wallet state.  So I experimented with a lock around GlobalState, and it appears to work well and also simplify code in the other State types as well.  Most other changes were sort of incidental along the way, as I was deep into refactoring.

Here are the high level changes:

 * create `GlobalStateLock` and use it for all state locking.
 * remove other locks (except those in storage layer, still a todo)
 * add tokio `AtomicRw` and `AtomicMutex` async lock types.
 * trace (log) all lock events (TryAcquire, Acquire, Release)
 * use tokio Builder when spawning tasks to provide a task name for tracing.
 * add `RustyLevelDbAsync` async-friendly DB wrapper
 * simplify `LightState`, now just a type alias for Block
 * add enum `BlockChainState` to distinguish Archival vs Light mode
 * use twenty_first storage iterators where appropriate
 * improve some doc-comments, `cargo doc` runs without warnings
 * general cleanups and refactors
 * log timestamp of each mined block, converted to localtime
 * display date fields using user's localtime in `neptune-dashboard` 

Many of these changes are inter-related.  However, if requested I could pull out some things into smaller, easier to review PRs.

The GlobalState lock is the most significant architectural change and so may need the most justification.  Here's a comment from the code that describes and makes a case for it:

```
`GlobalStateLock` holds a [`tokio::AtomicRw`] ([`RwLock`]) over [`GlobalState`].

Conceptually** all reads and writes of application state
require acuiring this lock.

Having a single lock is useful for a few reasons:
 1. Enables write serialization over all application state.
    (blockchain, mempool, wallet, global flags)
 2. Readers see a consistent view of data.
 3. makes it easy to reason about locking.
 4. simplifies the codebase.

The primary drawback is that long write operations can
block readers.  As such, every effort should be made to keep
write operations as short as possible, though
correctness/atomicity have first priority.

Using an `RwLock` is beneficial for concurrency vs using a `Mutex`.
Readers do not block eachother.  Only a writer blocks readers.
See [`RwLock`](std::sync::RwLock) docs for details.

** At the present time, storage types in twenty_first::storage
implement their own locking, which means they can be mutated
without acquiring the `GlobalStateLock`.  This may change in
the future.

Usage conventions:

// property naming:
struct Foo {
    global_state_lock: GlobalStateLock
}

// read guard naming:
let global_state = foo.global_state_lock.lock_guard().await;

// write guard naming:
let global_state_mut = foo.global_state_lock.lock_guard_mut().await;

These conventions make it easy to distinguish read access from write
access when reading and reviewing code.

When using a read-guard or write-guard, always drop it as soon as possible.
Failure to do so can result in poor concurrency or deadlock.

Deadlocks are generally not hard to track down.  Lock events are traced.
The app log records each `TryAcquire`, `Acquire` and `Release` event
when run with `RUST_LOG='info,neptune_core=trace'`.

If a deadlock has occurred, the log will end with a `TryAcquire` event
(read or write) and just scroll up to find the previous `Acquire` for
write event to see which thread is holding the lock.
```

So yeah I think the GlobalStateLock is a good change.  If nothing else, it makes a lot of code easier to read/maintain without all the interior locks.   But if others don't agree, I am willing to do some more work to inform and hopefully justify it, such as creating/running some perf benchmarks... (though that would take time that could be spent on other things).     Alternatively, some of the other changes could be split out into separate PRs.   I also have some older code from the branch that encapsulates existing locks rather than using `GlobalStateLock`.

Notes:

* I have squashed the 70+ original commits since early December into a single commit because rebasing all of those onto present tip would have taken ages.
* I had some twenty-first dep difficulties:
    * This relies on some twenty-first changes made after 0.36.0.   So there is a Cargo.toml patch section.   The patches are in branch `0_36_backports` in twenty-first repo.
   * I couldn't just use twenty-first master because it appears to contain some changes that are incompatible with triton-vm and/or tasm.
   * triton-vm now exports twenty_first (yay!) but tasm does not yet export triton-vm.   If/when that happens, we could/should change all use statements in neptune-core, and that would enable mixed versions of twenty-first for times like this.    But for now the 0_36_backports branch is a workaround.
   * If twenty-first::storage and twenty-first::sync were in their own crates, that would have solved this issue also.  Perhaps eventually.

Testing:

* ran two nodes on local subnet overnight, each starting with no wallet or blockchain, connected only to eachother, with dashboard running.   Both found blocks and reported no significant errors.
* synced with alphanet-v5 peers.
* passes all unit tests and doc tests
* cargo doc generates docs without warnings.
* fmt and cargo are clean, including tests

### DIFF
--- a/src/database/rusty.rs
+++ b/src/database/rusty.rs
@@ -103,6 +103,12 @@ where
             Err(err) => panic!("database failure: {}", err),
         }
     }
+
+    fn flush(&self) {
+        self.database
+            .write(&WriteBatch::new(), true)
+            .expect("Database flushing to disk must succeed");
+    }
 }
 
 /// `RustyLevelDbAsync` provides an async-friendly and clone-friendly wrapper
@@ -213,6 +219,12 @@ where
         task::spawn_blocking(move || inner.delete(key))
             .await
             .unwrap()
+    }
+
+    /// Delete database value asynchronously
+    pub async fn flush(&self) {
+        let inner = self.0.clone();
+        task::spawn_blocking(move || inner.flush()).await.unwrap()
     }
 }
 
--- a/src/database/rusty.rs
+++ b/src/database/rusty.rs
@@ -1,12 +1,17 @@
 use super::leveldb::LevelDB;
 use anyhow::Result;
-use leveldb::batch::WriteBatch;
-use leveldb::iterator::Iterable;
-use leveldb::options::ReadOptions;
 use serde::de::DeserializeOwned;
 use serde::Serialize;
 use std::marker::PhantomData;
 use std::path::Path;
+use std::sync::Arc;
+use tokio::task;
+use twenty_first::leveldb::{
+    batch::WriteBatch,
+    iterator::Iterable,
+    options::{Options, ReadOptions},
+};
+use twenty_first::leveldb_sys::Compression;
 use twenty_first::storage::level_db::DB;
 
 pub struct RustyLevelDB<Key, Value>
@@ -35,8 +40,8 @@ where
     }
 }
 
-pub fn create_db_if_missing() -> leveldb::options::Options {
-    let mut opts = leveldb::options::Options::new();
+pub fn create_db_if_missing() -> Options {
+    let mut opts = Options::new();
     opts.create_if_missing = true;
     opts
 }
@@ -47,7 +52,7 @@ where
     Value: Serialize + DeserializeOwned,
 {
     /// Open or create a new or existing database
-    fn new(db_path: &Path, options: &leveldb::options::Options) -> Result<Self> {
+    fn new(db_path: &Path, options: &Options) -> Result<Self> {
         let database = DB::open(db_path, options)?;
         let database = Self {
             database,
@@ -57,30 +62,30 @@ where
         Ok(database)
     }
 
-    fn get(&mut self, key: Key) -> Option<Value> {
+    fn get(&self, key: Key) -> Option<Value> {
         let key_bytes: Vec<u8> = bincode::serialize(&key).unwrap();
         let value_bytes: Option<Vec<u8>> = self.database.get(&key_bytes).unwrap();
         value_bytes.map(|bytes| bincode::deserialize(&bytes).unwrap())
     }
 
-    fn put(&mut self, key: Key, value: Value) {
+    fn put(&self, key: Key, value: Value) {
         let key_bytes: Vec<u8> = bincode::serialize(&key).unwrap();
         let value_bytes: Vec<u8> = bincode::serialize(&value).unwrap();
         self.database.put(&key_bytes, &value_bytes).unwrap();
     }
 
-    fn batch_write(&mut self, entries: &[(Key, Value)]) {
+    fn batch_write(&self, entries: impl IntoIterator<Item = (Key, Value)>) {
         let batch = WriteBatch::new();
-        for (key, value) in entries.iter() {
-            let key_bytes: Vec<u8> = bincode::serialize(key).unwrap();
-            let value_bytes: Vec<u8> = bincode::serialize(value).unwrap();
+        for (key, value) in entries.into_iter() {
+            let key_bytes: Vec<u8> = bincode::serialize(&key).unwrap();
+            let value_bytes: Vec<u8> = bincode::serialize(&value).unwrap();
             batch.put(&key_bytes, &value_bytes);
         }
 
         self.database.write(&batch, true).unwrap();
     }
 
-    fn delete(&mut self, key: Key) -> Option<Value> {
+    fn delete(&self, key: Key) -> Option<Value> {
         let key_bytes: Vec<u8> = bincode::serialize(&key).unwrap(); // add safety
         let value_bytes: Option<Vec<u8>> = self.database.get(&key_bytes).unwrap();
         let value_object = value_bytes.map(|bytes| bincode::deserialize(&bytes).unwrap());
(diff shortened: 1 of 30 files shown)


src/models/state/mod.rs
```diff
@@ -709,16 +811,34 @@ impl GlobalState {
...
+            .archival_mutator_set
+            .set_sync_label(hash);
+        self.chain
+            .archival_state_mut()
+            .archival_mutator_set
+            .persist();
+
+        debug!("Persisted all databases");
 
         Ok(())
```
Sword-Smith, line 843: It seems to me that there are some databases that are not being persisted here: `block_index_db` in `ArchivalState` and `peer_databases` in `NetworkingState`. Shouldn't they be flushed here as well?

src/database/rusty.rs
```diff
@@ -93,54 +98,174 @@ where
...
+///     driving other futures forward.
+///
+/// See:
+///  * <https://github.com/Neptune-Crypto/neptune-core/issues/74>
+///  * <https://internals.rust-lang.org/t/warning-when-calling-a-blocking-function-in-an-async-context/11440/5>
+///  * <https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html>
+///
+/// Therefore `RustyLevelDbAsync` wraps the sync methods with `spawn_blocking()`
+/// so that the tokio runtime can run the blocking IO on a thread where blocking
+/// is acceptable
```
Sword-Smith, line 121: I'm a bit unsure about why this is needed. The general struct seems to be 

```rust
pub async fn <method>(&self, other_args) {
    let inner = self.0.clone();
    task::spawn_blocking(move || inner.<method>(other_args)).await.unwrap()
}
```

So you're just running the database operation in a new thread and waiting for that thread to finish instead of just running it in the original thread? I don't fully understand why that is beneficial. Maybe you can elaborate on what the documentation from `tokio` says as I'm not sure I fully get it?

dan-da:
I just pushed a fix for the `prune_abandoned_monitored_utxos()` issue discovered in testing.

Here are my notes about the investigation.

[debug_prune_abandoned.txt](https://github.com/Neptune-Crypto/neptune-core/files/13946882/debug_prune_abandoned.txt)

src/database/rusty.rs
dan-da: I will try.   You probably know most of this, but anyway...

So the purpose of async/await functions (futures) is that they can return immediately to the caller without blocking.  The executor then polls the function to see if it is ready yet or not.   By doing this with multiple functions, it can effectively "multitask" performing multiple operations concurrently, even from a single OS thread.  Unlike native threads that can be pre-empted by the OS, this is a cooperative execution model.   It requires voluntary participation.    So an async function that calls blocking code is basically not cooperating.  It is blocking the entire *native/OS* thread, which prevents the executor from polling (aka driving forward) other tasks.  So then, concurrency suffers.    This is why crates like tokio provide async replacements for many std lib items, especially the filesystem APIs which are written in a blocking manner.

There is some further nuance here.  The default tokio runtime (which we are using) uses a multi-threaded executor.   By default (again) it creates one thread per CPU, eg on my laptop that's 4 native OS threads.   So even if one is blocked, some tasks can still proceed, which may alleviate the problem to a degree.   I do not know the details beyond that.... it may be possible for a single blocking function to still stall some other task(s).  But let's say we have 4 async functions executing simultaneously, one on each CPU and each is calling a blocking function.  Then all other tasks are blocked.   oops.

So, what it boils down to is:  async best-practice is to avoid calling blocking functions, especially file I/O.  File I/O takes a really long time and most of it does not involve the CPU, so that's just wasted capacity.

So that's why I filed #74 and #75 a while back.

Now, with regards to the spawn_blocking() calls in `RustyLevelDBAsync` that is basically a way to inform the tokio executor "hey I'm going to do some blocking stuff, please prepare accordingly".  So tokio may create a new OS thread, or move my task to an existing thread in a thread pool.  Anyway, it arranges to keep its regular OS threads freed up for polling/advancing well-behaving futures that return quickly.

I may be off on a detail here or there, but I believe the overall gist is correct.

Sword-Smith:
Regarding that commit [9b9ff2c](https://github.com/Neptune-Crypto/neptune-core/pull/89/commits/9b9ff2cd8c60ada42e75256843eff9d62faa086f) that solves the deadlock issue in `prune_abandoned_monitored_utxos`, I found [this comment](https://github.com/tokio-rs/tokio/discussions/2653#discussioncomment-35875) to be helpful:
> I see the issue. You shouldn’t be using futures::executor alongside with Tokio as neither executor (Tokio or futures) are aware of one another, which explains the blocking behavior you’re seeing.

I’ll also note that #[tokio::main] expands into tokio::Runtime::block_on, which means this code has two nested block_ons.

Good job solving this problem! It's been a thorn side for quite a while.

Sword-Smith:
> I can get my node running this version to crash like so:
> ...

I reset the node, with prunable UTXOs, and could not reproduce the problem. This is likely due to a corrupted database, and I'm not sure we should pursue this further for now.

src/models/state/mod.rs
dan-da: writing down what was discussed in conversation:

yeah, the flush calls are not necessary since `RustyLevelDB::batch_write()` sets the sync flag to true when calling `twenty_first::storage::leveldb::DB::write()`.    LevelDB guarantees that the data has been fully written to disk (synced/flushed) when the write call returns.

In fact, the C++ levelDB API does not offer a flush() method at all.   For that reason, I had removed it from RustyLevelDB.

I see though that you kept RustyLevelDB::flush() and modified it to perform an extra, empty write.

afaict, this extra write is un-necessary disk I/O.

That said, I think we are not calling flush() so frequently, so it probably won't be an issue for performance for the foreseeable future.    So I have added the flush calls back into this PR.

I also updated RustyLevelDB so that it defaults the sync flag to true, in case someone calls a setter other than batch_write() ie (put, delete).   Searching the code we don't seem to use these presently.  It might be best just to remove them RustyLevelDB, so that any future devs are not tempted to call them, and bypass batch writes.

Sword-Smith (approved):

src/models/state/mempool.rs
```diff
@@ -324,15 +180,23 @@ impl MempoolInternal {
...
+    ///
+    /// Computes in O(1)
+    pub fn is_empty(&self) -> bool {
         self.tx_dictionary.is_empty()
     }
 
-    fn get_transactions_for_block(&self, mut remaining_storage: usize) -> Vec<Transaction> {
+    /// Return a vector with copies of the transactions, in descending order, with
+    /// the highest fee density not using more than `remaining_storage` bytes.
+    /// Typically a block is about 0MB, meaning that the return value of this function is also less than 1MB.
```
Sword-Smith, line 198: This last line of the comment seems incorrect.

Sword-Smith:
LGTM

PR #94, opened 2024-01-19
fix: reduce dashboard CPU usage via fewer draws

dan-da:
The dashboard was constantly using around 15% CPU on my system while doing basically nothing.  This drops usage to under 1% for all screens.

The high usage was caused by constantly re-drawing the screen in the main loop.

The implemented fix is to add a `RefreshScreen` event which is sent by sub-screens when data or key events occur that require a redraw.

In this way, redraws only occur when required.

----

I am not certain this is the best or most elegant way to fix it, but it seems to works well on my system.

Testing: I tested functionality in all screens and compared with a dashboard instance running the previous code, and they appear to function identically, including the send screen which has some tricky stuff grabbing user input and moving control focus.

### DIFF
--- a/src/bin/dashboard_src/dashboard_app.rs
+++ b/src/bin/dashboard_src/dashboard_app.rs
@@ -97,6 +97,7 @@ pub enum ConsoleIO {
 /// Events that widgets can pass to/from each other
 #[derive(Debug, Clone)]
 pub enum DashboardEvent {
+    RefreshScreen,
     ConsoleEvent(Event),
     ConsoleMode(ConsoleIO),
     Shutdown(String),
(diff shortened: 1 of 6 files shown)


dan-da:
@Sword-Smith pls review if u wish, or else I can just merge it.

src/bin/dashboard_src/dashboard_app.rs
```diff
@@ -266,20 +273,25 @@ impl DashboardApp {
...
+                app.handle(
+                    &mut terminal,
+                    DashboardEvent::ConsoleMode(ConsoleIO::InputSupplied(string)),
+                )
                 .await?;
             }
 
-            terminal.draw(|f| app.render(f))?;
-
+            // note: setting a low duration like 100 can cause high CPU usage
```
Sword-Smith, line 289: What does 100 mean here (in the comment), microseconds, ticks?

Sword-Smith (approved):
LGTM

src/bin/dashboard_src/dashboard_app.rs
dan-da: milliseconds.  so 100 = 1/10 second.

The comment applies to the following line is, which calls `::from_millis()`
`if event::poll(Duration::from_millis(100))?`

PR #97, opened 2024-01-24
refactor public announcement

### LINKED ISSUE #40: Refactor: Public Script --> Public Announcement
Neptune has three types of scripts:

 - lock scripts, which capture the policy under which a UTXO can be spent;
 - type scripts, which capture the evolution rules of state-carrying UTXOs;
 - public scripts, which are attached to transactions as opposed to UTXOs.

The point of public scripts (as was understood at the time the [white paper](https://neptune.cash/whitepaper/) was written) is to mediate disputes. For instance, to close a lightning channel unilaterally, the closer needs to provide (by means of public script) cryptographic data that would allow the counterparty to contest the closure if it were malicious.

However, while *data broadcasting* is certainly a necessary feature, it is by no means clear that *public computation* is also. If it turns out that data boardcasting suffices, then we should drop the program part of the public scripts and rename the notion and structs to variants of "announcement" instead.

Questions:

 1. How could lightning channels work, in more detail, assuming only public announcements are available and not public scripts?
 2. Is there a meaningful more complex overlay protocol that requires public scripts and for which public announcements do not suffice?

aszepieniec:
The refactor replaces {*public script*, *public script hash*, *public script input*} with *public announcement*. It closes #40.


aszepieniec:
Note that these changes are not backwards-compatible, and so this requires bumping the testnet version (to alphanet v6).

Issue #100, opened 2024-01-30
flaky test: `allow_multiple_inputs_and_outputs_in_block`


aszepieniec:
Test `allow_multiple_inputs_and_outputs_in_block` in `archival_state.rs` fails some of the time.

To decrease the probability of failure:
 - run the test with a single thread: `cargo test allow_multiple_inputs_and_outputs_in_block -- --test-threads 1`

To increase the probability of failure:
 - uncomment the print statements in the test body in between 
```        block_2.accumulate_transaction(tx_from_alice);```
  and
```
        block_2.accumulate_transaction(tx_from_bob);
```

Commit c239e362683126149b3e25d360e485eb8a03a4c7 turns the failing `unwrap` into an `unwrap_or_else` with a `panic!` so as to display debug information.

I'm not sure how to fix this bug. I decided to shelve this bug hunt for the time being and devote attention to things that I can do. Feel free to take it off my plate.

**Comment by aszepieniec** [comment]
To increase also try: `cargo test allow_multiple_inputs_and_outputs_in_block -- --nocapture`.

**Comment by Sword-Smith** [comment]
Got this error when running the `allocate_sufficient_input_funds_test` test as well.

This happened while my machine was running a node.

```
thread 'models::state::wallet::wallet_tests::allocate_sufficient_input_funds_test' panicked at src/util_types/mutator_set/mutator_set_kernel.rs:191:21:
Can't get chunk index 2 from dictionary! dictionary: {}
stack backtrace:
   0: rust_begin_unwind
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/panicking.rs:72:14
   2: neptune_core::util_types::mutator_set::mutator_set_kernel::MutatorSetKernel<H,M>::remove_helper::{{closure}}
             at ./src/util_types/mutator_set/mutator_set_kernel.rs:191:21
```

**Comment by dan-da** [comment]
Okay, so the panic is occuring in mutator_set_kernel.rs, remove_helper() method:

```
thread 'models::state::archival_state::archival_state_tests::allow_multiple_inputs_and_outputs_in_block' panicked at src/util_types/mutator_set/mutator_set_kernel.rs:195:21:
Can't get chunk index 0 from dictionary! dictionary: {}
```

which is this this call to get_mut():
```
[...]

PR #101, opened 2024-01-30
Revisit Block Structure

aszepieniec:
This PR:

 - introduces a skeleton for verifying blocks using TritonVM
 - modifies the `Block` data structure
 - redefines the block's hash in [MAST](https://www.mit.edu/~jlrubin/public/pdfs/858report.pdf) terms
 - refactors public scripts into public announcements
 - prefers the block hash over its header as index by which to fetch blocks from the database

Note that (some of) these changes are not backwards-compatible, meaning that this essentially corresponds to a new alphanet release.


aszepieniec:
Note that the unsuccessful test runs (in both cases) boil down to a single failing test: the flaky one from #100.

Sword-Smith (approved):
Excellent breakdown of the validity logic. I look forward to working with these traits `SecretWitness` and `ValidationLogic`. Also great that the block hash is stored or transferred but always calculated.

I think we can avoid the manual implementation of `discriminant`. If the variants do not carry data, you can just cast the values to `usize`. I would keep the `HasDiscriminant` trait though as it's nice to have a uniform way of getting the discriminants for these field selectors.

PR #114, opened 2024-02-29
feat: Amounts

aszepieniec:
Amounts and amount logic are quite important in the design of a currency. This PR:

 - Implements (in compiler-friendly rust) a type script `NativeCurrency` that verifies that transactions are balanced, *i.e.*, do not generate new coins or have negative outputs.
  - Upgrades the dependency on `tasm-lib`.
  - Simplifies the internal representation of `NeptuneCoins` amounts: it now uses a `u128` rather than a `U32s<NUM_LIMBS>`.

### DIFF
--- a/src/models/blockchain/type_scripts/native_currency.rs
+++ b/src/models/blockchain/type_scripts/native_currency.rs
@@ -111,10 +111,13 @@ impl ConsensusProgram for NativeCurrency {
                 if input_salted_utxos.utxos[i as usize].coins[j as usize].type_script_hash
                     == self_digest
                 {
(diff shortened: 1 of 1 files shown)


aszepieniec:
- Implement (in compiler-friendly rust) a type script `NativeCurrency`
   that verifies that transactions are balanced, *i.e.*, do not generate
   new coins or have negative outputs.
 - Upgrades the dependency on `tasm-lib`.
 - Simplifies the internal representation of `NeptuneCoins` amounts:
   it now uses a `u128` rather than a `U32s<NUM_LIMBS>`.

PR #124, opened 2024-04-02
make storage async, attempt 3.

dan-da:
addresses #108, #75
obsoletes #120, #121

co-authored with @aszepieniec.

This PR builds on #120.  @aszepieniec was able to remove the `Mmr` trait which resolves [this headache](https://github.com/Neptune-Crypto/neptune-core/pull/120#issuecomment-2010302460) and I added some further cleanups and fixes to get all tests passing afterwards.

Unlike #120, this PR does not spawn an OS thread for each method call in `ArchivalMmr`(or any method).

All tests are passing.  <strike>however it is important to note that f3384fbe6f642de9ca4474b4e5b4ead37a4da36c contains a questionable fix to `ArchivalMutatorSet::batch_remove()` that should be carefully reviewed by @aszepieniec prior to merging this.</strike>.   edit: fixed properly by @aszepieniec in d602858f3d77933119153d9dc0a60e512e52a099

I should also note that this PR requires much less code to become async than #121.

dependencies: this code depends on github builds of `twenty-first` and `tasm-lib`.  The required changes are already in `master` of each crate, so once new releases are published we can use those instead.

Also, I have another PR in the pipeline that depends on this one.


src/models/blockchain/block/mod.rs
```diff
@@ -557,20 +575,16 @@ mod block_tests {
...
     };
-    use tasm_lib::twenty_first::{
-        storage::level_db::DB,
-        util_types::{
-            emojihash_trait::Emojihash,
-            mmr::archival_mmr::ArchivalMmr,
-            storage_schema::{SimpleRustyStorage, StorageWriter},
-        },
-    };
+    use tasm_lib::twenty_first::util_types::emojihash_trait::Emojihash;
```
aszepieniec, line 587: Note that trait Emojihash has been killed in twenty-first v39.

src/models/state/mod.rs
```diff
@@ -711,26 +722,36 @@ impl GlobalState {
...
-                    let msmp = match msmp_res {
-                        Some((_blockh_hash, msmp_val)) => msmp_val,
-                        None => continue 'inner,
-                    };
-
-                    // If UTXO matches, then check if the AOCL index is also a match.
-                    // If it is, then the UTXO is already in the wallet database.
-                    if msmp.auth_path_aocl.leaf_index == incoming_utxo.aocl_index {
-                        continue '_outer;
+
```
aszepieniec, line 725: The previous version handled all DB IO outside of the loop (referring to line 735 in the new version). IIuc, this is no longer true. Isn't that a performance degradation?

src/models/state/mod.rs
```diff
@@ -801,11 +826,9 @@ impl GlobalState {
         // loop over all monitored utxos
         let monitored_utxos = self.wallet_state.wallet_db.monitored_utxos_mut();
 
-        // note: iter_mut_lock holds a write-lock, so it should be dropped
-        // immediately after use.
-        let mut iter_mut_lock = monitored_utxos.iter_mut();
-        'outer: while let Some(mut setter) = iter_mut_lock.next() {
-            let monitored_utxo = setter.value();
+        'outer: for i in 0..monitored_utxos.len().await {
```
aszepieniec, line 829: Same as previous comment: looks like DB IO inside of a loop.

src/models/state/wallet/wallet_state.rs
```diff
@@ -338,40 +339,46 @@ impl WalletState {
...
-                        };
-                        warn!(
-                        "Unable to find valid membership proof for UTXO with digest {utxo_digest}. {confirmed_in_block_info} Current block height is {}", new_block.kernel.header.height
-                    );
-                        // panic!("Unable to find valid membership proof.");
+        {
+            let stream = monitored_utxos.stream().await;
+            pin_mut!(stream); // needed for iteration
+
+            while let Some((i, monitored_utxo)) = stream.next().await {
```
aszepieniec, line 347: Same as above: Aren't we reading from the DB `monitored_utxos.len()` many times here, whereas we used to just be reading once? Note that monitored UTXOs is unlikely to be more than a few thousand instances and can thus safely fit into RAM.

src/models/state/wallet/wallet_state.rs
```diff
@@ -466,34 +473,42 @@ impl WalletState {
...
-        assert_eq!(
-            mutxo_with_valid_mps,
-            valid_membership_proofs_and_own_utxo_count.len(),
-            "Monitored UTXO count must match number of managed membership proofs"
-        );
+        {
+            let stream = monitored_utxos.stream_values().await;
+            pin_mut!(stream); // needed for iteration
+
+            let mutxo_with_valid_mps = stream
```
aszepieniec, line 488: Consider renaming this variable to something that reflects that it counts something.

src/util_types/mutator_set/removal_record.rs
```diff
@@ -139,14 +139,14 @@ impl RemovalRecord {
     /// Update a batch of removal records that are synced to a given mutator set, given
     /// that that mutator set will be updated with an addition. (The addition record
```
aszepieniec, line 140:
```suggestion
    /// Update a batch of removal records that are synced to a given mutator set, in anticipation
    /// of one addition to that mutator set. (The addition record
```

aszepieniec (approved):
In some places we `.await` inside a loop, which sounds bad performance-wise especially if it resolves through a database IO. Could you comment on how that affects performance and what the idiomatic way to do things asyncly is?

Also worth noting that this PR gave rise to issues #126 and #127 which are about avoiding repeated IO calls (and other optimizations) in the course of archival MMR and mutator set logic.

src/models/blockchain/block/mod.rs
dan-da: yeah, so that will affect RPC and dashboard display, at least.  I'm not sure what else.    seems outside the scope of this PR.

I'm thinking its best to merge this PR as-is, which depends on a specific twenty-first github revision prior to emojihash removal.

Then we can remove it from neptune-core in a much smaller followup commit/pr.

src/models/state/mod.rs
dan-da: I will review.

dan-da: Analogy: This is basically the difference between reading a file in one call vs reading it line by line.  Reading the entire file in one call is fine if it is guaranteed to be a small file and/or you have low concurrency.   However, if you are reading in a 1 gigabyte file and you have much concurrency, your process will choke.  

Specifics:
Ok, in this case the previous code was calling monitored_utxos.get_all() before the loop.   This means it must read all available monitored_utxos into memory, all rows are collected into a Vec.  The collect itself takes some time.    I'm unclear how many utxos there could be, but I imagine that for some wallets (exchanges, frequent traders, multi-user wallets etc) it could be quite large with possibly tens of thousands of utxos.  So this could use significant amount of memory.   And the memory usage increases by the number of concurrent tasks performing same/similar operations.  Also, internally get_all() is reading records from the DB one-by-one anyway.  

The new code reads utxos one by one, so there is no need to allocate a bunch of memory up-front.  There is no collect() into a Vec.  We begin processing the first utxo right away.

In the past I benchmarked get_all (wraps get_many) vs get() in a loop.  iirc, get_all() was marginally faster but not enough to warrant increased memory, etc.   And it may be that it was only faster when it had pending writes, which meant that some items were in cache.   I would need to review the benches to be certain.

So yeah, mainly when I made this change I was thinking about avoiding the collect() and saving memory and generally being more concurrent.

related:  the new code is using an async `Stream` ([Stream primer](https://www.qovery.com/blog/a-guided-tour-of-streams-in-rust/)) instead of an iterator.  A `Stream` can potentially provide items in any order as each becomes available and the caller doesn't need to care.  At present, the storage layer that provides the `Stream` is not doing this, it calls get().await on each index in the order received.  But it could be modified to fire off all the gets and join_all() the results, so they come back in order of completion.  That *might* be faster, or it might create disk contention and be slower, I dunno.  Would need testing, and doesn't seem a priority now.

src/models/state/mod.rs
dan-da: In this case, both the previous code and new code are retrieving and setting elements one-by-one inside the loop.  They are equivalent.

The old code is using a mutating iterator which provides a setter, which holds the value and allows mutation.

Mutating streams do not exist afaik, so I was unable to duplicate the mutating iterator in the storage layer.  Instead, the new code uses regular get() and set() with indexes, which is logically the same.

src/models/state/wallet/wallet_state.rs
dan-da: The new logic is equivalent to the old code.  Both are reading line-by-line.   Same general comments apply as above.

Detail:

The old code for the for loop was: 

```
   for (i, monitored_utxo) in monitored_utxos.iter() {
       ...
   }
```

this is equivalent:

```
    while let Some((i, monitored_utxo)) = stream.next().await {
        ...
    |
```


With regards to len(), it is called exactly once, in StorageVecStream::stream(), which looks like:

```
    async fn stream<'a>(&'a self) -> impl Stream<Item = (Index, T)> + 'a
    where
        T: 'a,
    {
        self.stream_many(0..self.len().await).await
    }
```

and stream_many looks like:

```
    async fn stream_many<'a>(
        &'a self,
        indices: impl IntoIterator<Item = Index> + 'a,
    ) -> impl Stream<Item = (Index, T)> + 'a
    where
        T: 'a,
    {
        stream! {
            for i in indices.into_iter() {
                yield (i, self.get(i).await)
            }
        }
    }
```

With regards to memory usage and concurrency, see my comment above.

dan-da:
> In some places we `.await` inside a loop, which sounds bad performance-wise especially if it resolves through a database IO. Could you comment on how that affects performance and what the idiomatic way to do things asyncly is?
> 
> Also worth noting that this PR gave rise to issues #126 and #127 which are about avoiding repeated IO calls (and other optimizations) in the course of archival MMR and mutator set logic.

dan-da:
> > In some places we `.await` inside a loop, which sounds bad performance-wise especially if it resolves through a database IO. Could you comment on how that affects performance and what the idiomatic way to do things asyncly is?

The await just means "pause until this call completes, and meanwhile process other pending tasks".   Whereas in the old code with blocking storage layer, the same DB call meant "pause this entire OS thread while we wait on disk IO, blocking all pending tasks, until this call completes".

So awaiting inside a loop is not bad per-se.  And indeed compared to the old code, we are much more concurrency-friendly.

For illustration, let's examine the same scenario with non-async DB calls vs async calls.

non-async:
We get an RPC call that performs a lot of DB operations, say 1000 ops, and each op takes 10 ms.  (just for example).  Ok, so that's 100 seconds spent processing that RPC call, just for the DB ops.   And the tokio thread executing that task is just sitting there twiddling its thumbs waiting for the db/disk during those 100 seconds.  It may well have 30 other tasks that it should be processing, but it can't because it is waiting on blocking I/O.   More tasks are coming in from the network (peers, rpc, etc) and they pile up.

async:
We get the same RPC call that in aggregate takes 100 secs in the DB (primarily disk i/o).   Each call to the DB is followed by `await` which immediately returns control to the tokio executor.  So as the DB begins processing, tokio's executor has begun processing the next pending task, and when it hits an await, the executor moves on to the next one, and so on.    The CPU remains utilized whenever a task needs attention and thus no pile-up occurs.

The tokio executor is very similar to an OS scheduler, except that it requires cooperation (via calling await) whereas OS schedulers are pre-emptive and can rip execution away from any thread at any time.  Since the tokio executor relies on cooperation (by calling await) its performance degrades rapidly when tasks do not cooperate (by executing for a long time without any call to await).

I'm likely stating things you already know, but I hope there's a helpful nugget in there somewhere.   ;-)

src/models/state/mod.rs
Sword-Smith: The function in questions is invoked by the main thread through `global_state_lock.resync_memebership_proofs()` it holds a write lock over the global state until it's done, i.e. until all monitored UTXOs are in an updated state. So any concurrency would have to be in threads that don't need to read from the global state.

The same is true for `prune_abandoned_monitored_utxos` in this same file: This is done by the main thread, and all other threads pretty much have to wait for that to finish anyway.

The most critical point of operation for this software is when the global state is updated with a new block. The main thread grabs a write lock to the global state, and most operations (mining, sharing blocks with peers, answering RPC calls, creating new transactions) should not and cannot function until a new valid, consistent state is reached. And the main thread will communicate that this is the case by releasing its lock on global state.

This means that the most important things to aim for for this software is
1. Correctness and atomicity of state updates (the lock on global state achieves this)
2. Fast state updates, so other tasks can continue
3. Concurrency in the case where a write lock is *not* held on the global state.

**Reading all monitored UTXOs into memory**
I calculated how big a monitored UTXO is, assuming 1bn UTXOs, the number I expect after 5-10 years:
```
Size of `MonitoredUtxo`, in `BFieldElement`
UTXO: 14
  digest: 5
  coins: 9
    type_script_hash: 5
    state: 4
spent/confirmed/abandoned_at:  3*8 = 24
blockhash_to_membership_proof (3x): 47980
  digest: 5
  MmrMembershipProof: (~30 * 5) = 150
  MsMembershipProof:
    2x digest: 10
    ChunkDictionary (max 45 of these): 45 * (1 + 150 + 200) = 15795
    	u64: 1
    	MmrMembershipProof: (~30 * 5)
    	Chunk: (~0.10 * 2048) ~= 200
```

Rounding up, that's 50.000 `BFieldElement`s per monitored UTXO, which is 400.000 bytes, 400KB. Normal users will have between 0 and 100 monitored UTXOs which is 40MB. Loading that into memory is not a problem on any computer, not even a smartphone.

People running exchanges can have up to 10.000 monitored UTXOs, so around 4GB. If you're running an archival node in a setting that handles high volume, I think 4GB of RAM is absolutely within reason.

Sword-Smith (changes requested):
See question/comment about main thread/locks held on global state, and RAM consumption.

src/models/state/mod.rs
dan-da: Thank-you for the memory usage calculations.  I'm surprised at the 400Kb and 40Mb figures for 100 mutxo.  That's even higher than I expected.

Of course I agree we want fast state updates.  That is something we should always strive for.  So then, I do not see any reason to believe that fetching all records from the database in a for loop and collecting into a Vec (which is what get_all() does) and then reading them in a second loop would be faster than fetching each from the database in a single for loop and never collecting into a Vec, which is what the code in this PR does.  Indeed, on the face of it, it seems like this version should be faster.

Rather than guessing and theorizing, it is better to benchmark.   There is an existing benchmark for DbtVec I made last year.  It doesn't yet build for the new async storage, but I just ran the version from a week ago, which should be largely equivalent for comparing get() vs get_all().  Note that get_all() calls get_many() internally.  Here are the results:

```
db_dbtvec                      fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ read_100_entries                          │               │               │               │         │
│  ├─ get_each_entry_1_time                  │               │               │               │         │
│  │  ├─ get_many_persisted    88.17 µs      │ 159.2 µs      │ 116.8 µs      │ 107.4 µs      │ 100     │ 100
│  │  ├─ get_persisted         82.15 µs      │ 130.5 µs      │ 103.7 µs      │ 95.3 µs       │ 100     │ 100
│  │  ├─ get_many_unpersisted  20.47 µs      │ 26.95 µs      │ 20.95 µs      │ 21.1 µs       │ 100     │ 100
│  │  ╰─ get_unpersisted       10.83 µs      │ 11.54 µs      │ 10.91 µs      │ 10.92 µs      │ 100     │ 100
│  ╰─ get_each_entry_20_times                │               │               │               │         │
│     ├─ get_many_persisted    1.686 ms      │ 2.345 ms      │ 1.717 ms      │ 1.806 ms      │ 100     │ 100
│     ├─ get_persisted         1.495 ms      │ 1.682 ms      │ 1.541 ms      │ 1.542 ms      │ 100     │ 100
│     ├─ get_many_unpersisted  304.4 µs      │ 388.6 µs      │ 314.1 µs      │ 318.2 µs      │ 100     │ 100
│     ╰─ get_unpersisted       156.3 µs      │ 195.8 µs      │ 165.8 µs      │ 163.3 µs      │ 100     │ 100
```

So we see here that for 100 records it is faster to call get() directly than to call get_many().  This may be due to additional logic we have in get_many or it may be due largely to the collect().  I would need to re-code dbtvec get_many() to isolate that.  Anyway, it is clear that get_many() and by extension get_all() is not faster, but actually slower both for persisted and unpersisted vecs.

With regards to concurrency and memory usage:

In the case of  restore_monitored_utxos_from_recovery_data() that is operating on locked global state, it is correct that only one loop will be executing at a time.   However, if we look at for example get_latest_balance_height_internal() which also iterates over all mutxo, that only acquires a read lock.  This means that N number of operations can occur concurrently.  And that means that memory usage is linear with the number of concurrent operations.  So our busy exchange node using 4Gb ram per operation that may be processing many concurrent operations hits 32Gb with just 8 concurrent operations.  That is of course an extreme example, but it is useful to consider edge cases.

bottom line:   The code in this PR reflects my understanding of the most performant way to retrieve these DB values, and it is my recommendation that we leave it as-is.  In my opinion, get_all() would be allocating memory unnecessarily and be slower.  So I hope that puts your mind(s) at ease.   Nevertheless, I will change it to use get_all() for all these loops if you should ask for it again.

Sword-Smith: It does put my mind somewhat at ease. Going forward, I think we should take two initiatives:
1. Measure and (log-info) output the time it takes to apply a new block to the archival state. I want to track that number and see how it scales with the number of MUTXOs.
2. Minimize duration of the write-lock. Our "scheduled tasks" (peer discovery, synchronization, mempool cleanup etc.) generally work such that they *ask* "is this action needed" and then only performs the action if this boolean evaluates to true. Resolving the boolean value should be done only with a read-lock on the global state.

dan-da: I created an issue for each of these.

src/models/state/wallet/wallet_state.rs
dan-da: I think it is ok for now.  the variable is only used on 3 subsequent lines, so no real chance of confusion.  might revisit later.

Issue #122, opened 2024-03-26
Make long running CPU operations async friendly


dan-da:
We should not let CPU intensive functions block the tokio executor for long as this harms concurrency, which can potentially stall other tasks: p2p connections, rpc calls, mining, etc. [1]

The task here will be to identify non-async functions in neptune-core that perform lengthy operations, and then wrap them with tokio's spawn_blocking() or possibly block_in_place().

"lengthy" is of course a vague term.  As a starting point I would consider 10 ms lengthy, 100 ms very lengthy, and 1+ seconds as falling over dead. [2]

Known candidates:

 * mining loop, see #123
 * generating transaction proof.   see: GlobalStateLock::create_transaction_from_data() 
 * executing triton-vm scripts [3]

Add others above as we identify them.

note: At this time, I'm really most interested in dealing with the very worst offenders, such as proving, which I'm told may take minute(s).   I think others are low priority and can be dealt with as we notice them through testing.  So I'm not proposing with this issue that we spend time now with profiling and optimization.

[1] as an alternative, or in addition, we might consider creating an OS thread for the rpc server, mining, and possibly each p2p connection, which are presently all tokio tasks in tokio's threadpool.  async/await uses cooperative multi-tasking, while OS threads use preemptive multi-tasking.

[2]  places where triton-vm scripts are run:
```
$ find src -type f | xargs grep "\.run(" | grep -v peer_loop | grep -v currency.rs | grep -v time_lock.rs | grep -v integrity.rs | grep -v lib.rs | grep -v "//"
src/models/consensus/mod.rs:                            .run(claim.input.clone().into(), raw_witness.clone().into())
src/models/consensus/mod.rs:                            .run(claim.input.clone().into(), raw_witness.clone().into())
src/models/consensus/mod.rs:                TimeLock.run(&public_input.individual_tokens, non_determinism)
src/models/consensus/mod.rs:                NativeCurrency.run(&public_input.individual_tokens, non_determinism)
src/models/blockchain/transaction/mod.rs:            match lock_script.program.run(
src/models/blockchain/transaction/mod.rs:                .run(public_input.into(), NonDeterminism::new(secret_input))
```

[3] I'm not sure there is any clear definition of what constitutes "blocking".  This [quote](https://www.reddit.com/r/rust/comments/ebpzqx/comment/fb7brnh/?utm_source=share&utm_medium=web2x&context=3) seems insightful:

[...]

PR #133, opened 2024-04-07
make long-running CPU tasks run on tokio's blocking threadpool.

dan-da:
Partially addresses #122.

This PR makes the following potentially CPU intensive operations execute on tokio's blocking threadpool, so they do not block the tokio executor:

* `create_transaction_from_data()` which calls the prover
* the inner mining loop (still single-threaded)
* calls to triton_vm::program::Program::run() in `Transaction`

I am making this a PR rather than just pushing the commits for two reasons:

1. bring greater team awareness to the need to run any lengthy CPU ops on tokio's threadpool (or separate OS thread) and examples/methodology for doing so.
2. solicit input/suggestions as to other potentially lengthy/blocking computations.  (please comment in #122)
3. `spawn_blocking()` expects a non-async fn, so I changed the (inner) mining loop so it doesn't make any async calls, which means it can't check the global state `syncing` flag.  Instead the main loop notifies the (outer) mining loop when syncing starts and stops, which in turn halts or starts the (inner) mining loop which is running in tokio's threadpool.  This should speed the inner mining loop a little, as it no longer needs to acquire a lock after computing each hash.  Anyway, I thought @Sword-Smith may wish to review this change.

note: #130 has macros for logging fn call duration.  Once that is merged, we can easily log duration of any suspected targets to determine which are worth moving to the blocking threadpool.

Commits:

```
    perf: add spawn_blocking around script execution
    
    Wraps spawn_blocking around triton::program::Program::run() calls
    within Transaction::validate_primitive_witness().
    
    This enables tokio to put potentially lengthy blocking code on its own
    thread.
    
    There are additional ::run() calls in consensus/mod.rs however
    they are within sync code, and spawn_blocking() can only be called
    from within an async fn. Making those async is a big refactor, so
    it is possible future work.
```

```
    perf: wrap mining loop with spawn_blocking()
    
    The mining loop is very CPU intensive and thus very async-unfriendly.
    
    We wrap the loop with spawn-blocking so it can execute on tokio's
    blocking threadpool and thus avoid blocking async tasks on the
    same thread.
```

```
    fix: send syncing start/stop messages to miner
    
    The mining loop is no longer async as it is wrapped by
    spawn_blocking().

    Thus it can no longer await async futures.  So we cannot check
    the GlobalState syncing flag, which requires an async lock
    acquisition.
    
    Instead, the main task sends StartSyncing and StopSyncing
    messages to the mining task which cancels the mining thread
    when it receives StartSyncing.
```

```
    perf: wrap tx prover with spawn_blocking
    
    The prover may take a very long time (minutes) to run and should not
    block the tokio executor.
    
    As such, we wrap `create_transaction_from_data()` with
    `spawn_blocking()` in order to execute this task in tokio's blocking
    threadpool.
```

### DIFF
--- a/src/main_loop.rs
+++ b/src/main_loop.rs
@@ -407,6 +407,7 @@ impl MainLoopHandler {
                         if !stay_in_sync_mode {
                             info!("Exiting sync mode");
                             global_state_mut.net.syncing = false;
+                            self.main_to_miner_tx.send(MainToMiner::StopSyncing)?;
                         }
                     }
 
@@ -463,6 +464,7 @@ impl MainLoopHandler {
                     socket_addr, claimed_max_height, claimed_max_pow_family
                 );
                     global_state_mut.net.syncing = true;
+                    self.main_to_miner_tx.send(MainToMiner::StartSyncing)?;
                 }
             }
             PeerThreadToMain::RemovePeerMaxBlockHeight(socket_addr) => {
--- a/src/mine_loop.rs
+++ b/src/mine_loop.rs
@@ -104,12 +104,46 @@ fn make_block_template(
 
 /// Attempt to mine a valid block for the network
 async fn mine_block(
+    block_header: BlockHeader,
+    block_body: BlockBody,
+    sender: oneshot::Sender<NewBlockFound>,
+    coinbase_utxo_info: ExpectedUtxo,
+    difficulty: U32s<5>,
+    unrestricted_mining: bool,
+) {
+    // We wrap mining loop with spawn_blocking() because it is a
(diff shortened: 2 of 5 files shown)


Sword-Smith:
Did you see my [comment](https://github.com/Neptune-Crypto/neptune-core/issues/122#issuecomment-2042352812) in the associated issue? I would like to make sure we agree on the fundamentals before proceeding with the specifics.

Sword-Smith:
Glad we agree on the fundamentals. On that note, LGTM.

Sword-Smith (approved):

Issue #138, opened 2024-04-19
Display CPU temperature in Dashboard


aszepieniec:
The dashboard has a field indicating the CPU temperature, but this field is currently unpopulated. It turns out there is a crate called [systemstat](https://lib.rs/crates/systemstat) that can return this number.

aszepieniec:
(Note that this task does not have a very high priority; I just made the issue to save the link somewhere.)

elielnfinic:
Can I work on this task?

dan-da:
@aszepieniec note that the dashboard may be running on a different machine from neptune-core.   If the goal is to display CPU temp of the machine hosting the node, then it seems this should be impl'd as a neptune-core rpc api.   Do you agree, or you have something else in mind?

@elielnfinic sure.  we'd welcome a PR.   You can take a look at neptune-core/src/rpc_server.rs for adding a new rpc call, perhaps `get_cpu_temps`.      Dashboard overview source is in src/bin/dashboard_src/overview_screen.rs.

That said, I'd suggest this functionality is outside the mission/scope of a cryptocurrency node, and there are plenty of dedicated tools for monitoring local or remote machines using all kinds of metrics so this might not be a long-lived feature, but I suppose it doesn't hurt to include for launch if it makes the dashboard a little more useful.

aszepieniec:
>  If the goal is to display CPU temp of the machine hosting the node, then it seems this should be impl'd as a neptune-core rpc api.

I certainly agree that this feature should be implemented with an RPC call.

The statement about `neptune-dashboard` potentially running on a different machine from `neptune-core` is perhaps more true in the future than in the present, or true with a little engineering work but false without. Right now, there is no authentication between `neptune-core` and `neptune-dashboard` and to send coins you do not need to enter a password or anything like that. So anybody who connects the dashboard successfully can spend coins. Since this is a security hazard, `neptune-core` only admits connections coming from localhost. So in practice, you a) either run `neptune-dashboard` on the same machine as `neptune-core` or b) set up a SSH tunnel or something similar, which is the part that requires a little engineering..

elielnfinic:
Alright, thanks @dan-da and @aszepieniec for the information.
Let me go through the files/structure of the project and build something. If I have any questions, I will let you know.
About the RPC call, I agree that this might be useful when `CPU temperature` is  requested from another machine.
I will explore the SSH possibility too.

Thanks.

PR #139, opened 2024-04-22
Feature : Get CPU temperature over the RPC 

elielnfinic:
Addresses : #138.

**Description**:

This pull request introduces the implementation of the get_cpu_temps method, which retrieves the CPU temperature for use in the dashboard. This feature leverages the systemstat library.

The `get_cpu_temps` method is located in` /src/rpc_server.rs` to facilitate its use via the RPC server.

It's important to note that the systemstat library does not support CPU temperature retrieval on Silicon Macs, as Apple has disabled this feature. Consequently, the associated test is designed to expect a None result when running on macOS.

As in `/src/bin/dashboard_src/overview_screen.rs`, `cpu_temperature` is a Option<f64>, I converted the f32 we get from `systemstat` into f64.

### DIFF
--- a/src/rpc_server.rs
+++ b/src/rpc_server.rs
@@ -752,7 +752,7 @@ impl RPC for NeptuneRPCServer {
             .await
     }
 
-    async fn cpu_temp(self, _context: Context) -> Option<f64> {
+    async fn cpu_temp(self, _context: tarpc::context::Context) -> Option<f64> {
         let current_system = System::new();
         match current_system.cpu_temp() {
             Ok(temp) => Some(temp.into()),
--- a/src/rpc_server.rs
+++ b/src/rpc_server.rs
@@ -1111,7 +1111,7 @@ mod rpc_server_tests {
     #[traced_test]
     #[tokio::test]
     async fn test_can_get_server_temperature() {
-        let (rpc_server, state_lock) =
+        let (rpc_server, _state_lock) =
             test_rpc_server(Network::Alpha, WalletSecret::new_random(), 2).await;
         let current_server_temperature = rpc_server.get_cpu_temps(context::current()).await;
         // Silicon Macs do not provide CPU temperature using systemstat
--- a/src/rpc_server.rs
+++ b/src/rpc_server.rs
@@ -749,7 +749,7 @@ impl RPC for NeptuneRPCServer {
             .await
     }
 
-    async fn cpu_temp(self, _context: tarpc::context::Context) -> Option<f64> {
+    async fn cpu_temp(self, _context: Context) -> Option<f64> {
         let current_system = System::new();
         match current_system.cpu_temp() {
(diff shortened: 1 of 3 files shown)


Sword-Smith:
Thanks for adding a test. PRs that introduce new tests make me happy :)

Your PR introduces a compiler warning. Can you fix that? I personally run `make all` to check that there are no bugs/compiler warnings but if you don't have a powerful machine, maybe you just want to run `cargo clippy --all-targets` and `cargo fmt --all --check` instead.

```
warning: unused variable: `state_lock`
    --> src/rpc_server.rs:1114:26
     |
1114 |         let (rpc_server, state_lock) =
     |                          ^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_state_lock`
     |
     = note: `#[warn(unused_variables)]` on by default

```

elielnfinic:
Thanks @Sword-Smith for your review. 
I will fix that quickly.

src/rpc_server.rs
```diff
@@ -196,6 +197,9 @@ pub trait RPC {
 
     /// Gracious shutdown.
     async fn shutdown() -> bool;
+
+    /// Get CPU temperature.
+    async fn get_cpu_temps() -> Option<f64>;
```
dan-da, line 202: I suggested naming it get_cpu_temps (plural) because I was expecting there would be a Vec of temps, one for each CPU core.  I see now that systemstat only returns a single value which surprises me.   But in light of that, perhaps its best just to name it `cpu_temp`.  I think the "get_" is superfluous.

elielnfinic: You are right @dan-da .

src/rpc_server.rs
```diff
@@ -744,6 +748,14 @@ impl RPC for NeptuneRPCServer {
             .get_all_own_coins_with_possible_timelocks()
             .await
     }
+
+    async fn cpu_temp(self, _context: tarpc::context::Context) -> Option<f64> {
```
dan-da, line 752: nit: no need for fully qualified path on Context.

```suggestion
    async fn cpu_temp(self, _context: Context) -> Option<f64> {
```

src/rpc_server.rs
```diff
@@ -1123,6 +1135,21 @@ mod rpc_server_tests {
         Ok(())
     }
 
+    #[allow(clippy::shadow_unrelated)]
```
dan-da, line 1138: nit: is this really needed?   I don't see any `let` shadowing.

dan-da (approved):
a couple minor nits but otherwise lgtm.

src/rpc_server.rs
elielnfinic: Sure, we can remove it.

Sword-Smith:
Thank you for your contribution, @elielnfinic. The PR was merged manually through CLI by me. You have of course been credited :)

Issue #125, opened 2024-04-03
opt: Use constant-time index arithmetic in MMRs


aszepieniec:
In particular, look at `get_peaks_with_heights_async` which currently uses the (out-dated) node-index and tree-traversal logic.

zkclay:
Hey all, could I take a look at this for my 2nd PR?

aszepieniec:
Absolutely!

Flagging @Sword-Smith because he knows more about efficient MMR arithmetic than I do.

Sword-Smith:
I'm not a fan of the call to `get_peaks_with_heights_async`, as is reads from disk needlessly. It's only used in the `get_peaks` method on `ArchivalMmr`, so let's get rid of `get_peaks_with_heights_async` completely.

I think it can be replaced by a call to `node_index_to_leaf_index` in `twenty-first` followed by a call to `get_peak_heights_and_peak_node_indices`, also in `twenty-first`. Then you have the node-indices of the MMR-peaks. The actual peak values can then be fetched by a `get_many` method call on the `digests` field on `ArchivalMmr`.

This reduces IO significantly, and it gives us index arithmetic that runs in $log(N)$ time, $N$ being the number of leaves in the MMR. It does not quite give us $O(1)$ running time, but I don't really think the difference will be measurable, so let's just focus on minimizing IO for now.

Sword-Smith:
I added a benchmark such that you can easily track your progress. It's already on `master` and can be run with `cargo bench get_peaks`.

Issue #127, opened 2024-04-03
Optimize IO in archival Mutator Sets


aszepieniec:
Depends on #126.

Sword-Smith:
@zkclay If you want to have a look at this, maybe have a look at this [blogpost](https://neptune.cash/blog/mutator-sets/). If you understand the mechanics of Merkle Mountain Ranges (MMR) you are 80 % towards understanding mutator sets. A mutator set contains two lists: 
1. an append-only commitment list which represents the outgoing UTXOs, and
2. an unbounded Bloom filter which represents spent UTXOs.

Both of these lists are represented by MMRs, so the logic toupdate a mutator set is 80 % MMR logic and 20 % other stuff.

zkclay:
Alright, I'll take a look at this! Thanks for the resources!

zkclay:
Hey @Sword-Smith  

I spent some time reading up on the Merkle Mountain Ranges and the Mutator Set docs that you wrote and taking a look at the code. The blogs posts are dense and I had to reread it a few times, but now I have a much better understanding of the two concepts now :) Thanks for the work you and the team has put into them!

However, I’m still a bit unsure as to the optimization strategy we want to pursue here. In taking a look at the mutator set code, it seems like the MMR operations are well encapsulated in the Archival MMR abstraction, and the bulk of the code for the mutator set is concerned with managing the sliding window bloom filter.  Are the optimizations we want to do here? I’m still a little hazy on the specifics of the SWBF logic and all the chunks we are manipulating, but I can look deeper into this, if it is the intended focus. 

I also noticed that there seems to be some functionality duplicated across the Mutator Set and Mutator Set Accumulator, is this what we want to look into? Or are you thinking of something else entirely?

Sword-Smith:
My motivation for this issue and #126 was to optimize the time that the node spends when updating its state with a new block. I looked into the archival-MMR and the archival-Mutator set code and it didn't look optimal to me. I wrote the original implementation of all of the MMR code and the mutator set code, but multiple (beneficial) rewrites have occurred since then: all access to persistent storage has been made `async`, which benefits concurrency, and we got rid of two traits: one that tied `MmrAccumulator` and `ArchivalMmr` together, and one that tied `MutatorSetAccumulator` and `ArchivalMutatorSet` together. We got rid of the two traits because we didn't want the accumulator versions to have to be implemented through `async` functions. Although these rewrites are net beneficial, I have the sense that some performance was lost in it.

[...]

PR #141, opened 2024-04-23
Optimize get_peaks_and_heights for archival MMR.

zkclay:
This PR is a response to Issue #125 

I chose to keep the `get_peaks_and_heights_async` and place the new functionality inside it rather than in `get_peaks` because the async function is used 5 or 6 times within the test suite.  I ran the benchmarks before and after the change and obtained the following results:

Master:

before_change

My PR:

after_change

In multiple runs I noticed the same small but consistent improvement on my machine, its possible that in setups with different tradeoffs between disk I/O and CPU speed, the difference would be more pronounced. 

EDIT: 

I definitely see even more gains after removing the `get_peaks_and_heights_async` function completely

deeper_cut

### DIFF
--- a/src/util_types/mutator_set/archival_mmr.rs
+++ b/src/util_types/mutator_set/archival_mmr.rs
@@ -16,7 +16,7 @@ use itertools::Itertools;
 
 use twenty_first::util_types::mmr::{
     mmr_accumulator::MmrAccumulator, mmr_membership_proof::MmrMembershipProof, mmr_trait::Mmr,
-    shared_advanced, shared_basic,
+    shared_advanced,
 };
 
 /// A Merkle Mountain Range is a datastructure for storing a list of hashes.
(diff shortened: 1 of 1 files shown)


dan-da:
Thanks @zkclay for the PR.  I like the comparative simplicity of this approach.

lgtm, but i will defer to @Sword-Smith on this one.

Sword-Smith:
Let's get rid of `get_peaks_and_heights_async` all calls to it in test can be replaced by a call to `get_peaks`, or be deleted entirely.

zkclay:
@Sword-Smith I have gotten rid of the function completely and added the updated benchmark above. I think removing it completely was a good ideas, aside from the speedup, it's satisfying to delete even more code 😄

Sword-Smith (approved):

Sword-Smith (approved):
Very nice :) Glad to see a 20 % speedup. LGTM.

Edit: We're intending to do the same for the ArchivalMutatorSet btw. In case you're looking for more stuff to do :)

zkclay:
Thanks, @Sword-Smith1 I will get to work on it! I assume that's the work covered by Issue #127 ?

Sword-Smith:
> Thanks, @Sword-Smith1 I will get to work on it! I assume that's the work covered by Issue #127 ?

It is :)

Issue #145, opened 2024-05-07
Can the mempool handle reorganizations?


Sword-Smith:
I don't think it can currently. One simple solution would be to simply clear the mempool in the case of a reorganization.

dan-da:
Clearing would lose transactions pending/unconfirmed tx.  I think those need to be kept.

With bitcoin, I believe it adds any extra/orphaned tx from the reorg into the mempool.   But iiuc we can't really do that because we no longer have the user-level tx, everything was aggregated into block-tx.    So any such orphans would just be lost.

I believe (need to verify) that bitcoin also must remove any tx from mempool that have already been included in one of the new blocks.   This is where we may have a problem, and probably why you suggested clearing the mempool.

My question would be:  does there exist any possibility to detect if a user-level tx has been accumulated into a block-level tx?   If so, then it would seem we can detect and remove these.    If not, then I don't see any solution other than clearing mempool unfortunately.

aszepieniec:
> does there exist any possibility to detect if a user-level tx has been accumulated into a block-level tx?

Yes: the user-level transactions are no longer valid. More specifically: the removal records induce updates of the mutator set which is designed to catch would-be double-spending attempts. Removal records consist of a list of indices and membership proofs in the SWBF MMR, and the latter field has no bearing on equality or duplicates -- so you are looking for a duplicate of index sets.

PR #153, opened 2024-05-11
Thv/#145 reorganization of mempool

Sword-Smith:
Simple solution to the inability of the mempool to handle reorganizations: Clear the mempool if that happens.

Context: all transactions need to contain proofs relating to the mutator set of the latest block. 3rd parties can update this information in the common case (new ti is child of previous tip), but this proof updating is complicated for reorganizations. So for now, we just clear the mempool in the case of a reorganization.

### DIFF
--- a/developer_docs/reorganizations.md
+++ b/developer_docs/reorganizations.md
@@ -1,0 +1,51 @@
+# How different parts of the state handle reorganizations
+Neptune is a blockchain which features recursive STARK proofs as part of its
+consensus mechanism. This implies that participants can synchronize trustlessly
+by simply downloading the latest block and verifying this. Unlike most other
+blockchains, it is not necessary to download all historical blocks to get a
+cryptographically verified view of the state of the blockchain.
+
+It is possible, though, to run an archival node that downloads all historical
+blocks. This archival node comes with additional functionality such as being
+able to reconstruct transaction's membership proofs, provide some historical
+transaction statistics, and allow other archival nodes to synchronize.
+
+This document provides an overview of how different parts of the client's state
+handle reorganizations.
+
+## State overview
+The client's state consists of the following parts:
+- wallet
+- light state
+- archival state (optional)
+- mempool
+
+The wallet handles transactions that the client holds the spending keys for.
+The light state contains the latest blocks which verifies the validity of the
+entire history of the blockchain. The archival state is optional and allows,
+among other things, the client to re-synchronize wallets that are no longer
+up-to-date. The mempool keeps track of transactions that are not yet included
+in blocks, thus allowing miners to confirm transactions by picking some from
+the mempool to include in the next block.
+
+### Wallet
+The wallet can handle reorganizations that are up to `n` blocks deep, where `n`
+can be controlled with the CLI argument `number_of_mps_per_utxo`.
+Reorganizations that are deeper than this will make the membership proofs of
+the transactions temporarily invalid until they can be recovered either through
+the client's own archival state (if it exists), or through a peer's archival
+state. This recovery process happens automatically.
+
+### Light State
+The light state only contains the latest block and thus can handle arbitrarily
+deep reorganizations.
+
+### Archival State
+The archival state can handle arbitrarily deep reorganizations.
+
+### Mempool
+The mempool can *currently* not handle reorganizations. If a reorganization
+occurs, all transactions in the mempool will be deleted, and the initiator of a
+transaction will have to publish the transaction again. The transactions that
+were included in blocks that are abandoned through this reorganization are not
+added to the mempool again, they also have to be published again.
--- a/developer_docs/reorganizations.md
+++ b/developer_docs/reorganizations.md
@@ -1,4 +1,4 @@
-# How different parts of the state handle reorganizations
+# How neptune-core handles reorganizations
 Neptune is a blockchain which features recursive STARK proofs as part of its
 consensus mechanism. This implies that participants can synchronize trustlessly
 by simply downloading the latest block and verifying this. Unlike most other
@@ -21,7 +21,7 @@ The client's state consists of the following parts:
 - mempool
 
 The wallet handles transactions that the client holds the spending keys for.
-The light state contains the latest blocks which verifies the validity of the
+The light state contains the latest block which verifies the validity of the
 entire history of the blockchain. The archival state is optional and allows,
 among other things, the client to re-synchronize wallets that are no longer
 up-to-date. The mempool keeps track of transactions that are not yet included
--- a/src/models/state/mempool.rs
+++ b/src/models/state/mempool.rs
@@ -73,11 +73,15 @@ pub struct Mempool {
     /// both descending and ascending order.
     #[get_size(ignore)] // This is relatively small compared to `LookupTable`
     queue: DoublePriorityQueue<Digest, FeeDensity>,
+
+    /// Records the digest of the block that the transactions were synced to.
+    /// Used to discover reorganizations.
+    tip_digest: Digest,
 }
 
 impl Mempool {
     /// instantiate a new, empty `Mempool`
-    pub fn new(max_total_size: ByteSize) -> Self {
+    pub fn new(max_total_size: ByteSize, tip_digest: Digest) -> Self {
         let table = Default::default();
         let queue = Default::default();
         let max_total_size = max_total_size.0.try_into().unwrap();
@@ -85,7 +89,13 @@ impl Mempool {
             max_total_size,
             tx_dictionary: table,
             queue,
+            tip_digest,
         }
+    }
+
+    /// Update the block digest to which all transactions are synced.
+    fn set_tip_digest_sync_label(&mut self, tip_digest: Digest) {
+        self.tip_digest = tip_digest;
     }
 
     /// check if transaction exists in mempool
@@ -181,6 +191,12 @@ impl Mempool {
         }
 
         None
+    }
+
+    /// Delete all transactions from the mempool.
+    pub fn clear(&mut self) {
+        self.queue.clear();
+        self.tx_dictionary.clear();
     }
 
     /// Return the number of transactions currently stored in the Mempool.
@@ -304,6 +320,15 @@ impl Mempool {
         previous_mutator_set_accumulator: MutatorSetAccumulator,
         block: &Block,
     ) {
+        // If we discover a reorganization, we currently just clear the mempool,
+        // as we don't have the ability to roll transaction removal record integrity
+        // proofs back to previous blocks. It would be nice if we could handle a
+        // reorganization that's at least a few blocks deep though.
+        let previous_block_digest = block.header().prev_block_digest;
+        if self.tip_digest != previous_block_digest {
+            self.clear();
+        }
+
         // The general strategy is to check whether the SWBF index set of a given
         // transaction in the mempool is disjoint (*i.e.*, not contained by) the
         // SWBF indices coming from the block transaction. If they are not disjoint,
@@ -356,6 +381,10 @@ impl Mempool {
         // transactions in the mempool. So we should shrink it to max size after
         // applying the block.
         self.shrink_to_max_size();
+
+        // Update the sync-label to keep track of reorganizations
+        let current_block_digest = block.hash();
+        self.set_tip_digest_sync_label(current_block_digest);
     }
 
     /// Shrink the memory pool to the value of its `max_size` field.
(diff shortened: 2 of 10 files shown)


Sword-Smith:
The failing test in CI is the one addressed in #150 and has nothing to do with this PR.

src/models/state/mempool.rs
```diff
@@ -283,13 +315,22 @@ impl Mempool {
...
-    /// transactions that were not removed in the previous step.
+    /// of a newly received block. Also update all mutator set data for mempool
+    /// transactions that were not removed.
     pub async fn update_with_block(
         &mut self,
         previous_mutator_set_accumulator: MutatorSetAccumulator,
         block: &Block,
     ) {
+        // If we discover a reorganization, we currently just clear the mempool,
+        // as we don't have the ability to roll transaction removal record integrity
```
dan-da, line 326: Is is it expected/planned that we will eventually be able to rollback transaction removal record integrity proofs?  ie, is it "just" an engineering issue or a theoretical one?

I think that forgetting user's transactions during re-org could be a pretty serious drawback for neptune, so I'm hoping this can be changed by launch, or at least that we'll be able to tell users it is "coming soon".

src/util_types/mutator_set/removal_record.rs
```diff
@@ -65,6 +66,38 @@ impl AbsoluteIndexSet {
...
+    /// active part of the Bloom filter.
+    ///
+    /// Returns an error if a removal index is a future value, i.e. one that's
+    /// not yet covered by the active window.
+    #[allow(clippy::type_complexity)]
+    pub fn split_by_activity(
+        &self,
+        mutator_set: &MutatorSetAccumulator,
+    ) -> Result<(HashMap<u64, Vec<u128>>, Vec<u128>), MutatorSetError> {
+        let mut inactive = indices_to_hash_map(&self.0);
```
dan-da, line 81: style: consider re-writing the for loop using iterator combinators to avoid the mutable variables.

note: @jan-ferdinand is quite good at this, if there's something tricky.

dan-da (approved):
The code lgtm, with a few nits/suggestions.

I am concerned about clearing user's tx from mempool during re-org, and hoping that is a temporary situation.   Either way, I'd like to see a note about it in the readme/docs, as I would put this in the realm of surprising behavior.

Also, I wonder if the wallet needs some extra logic to become aware that a transaction has been removed from the mempool and thus the user's balance and tx history may need updating?   I suppose that if the wallet is always obtaining this info by querying the archival_state + mempool, then it should automatically be updated, but I just want to pose the q to ensure this is what happens and that you've considered/tested wallet/dashboard state.

Finally, I'm not up-to-speed on mutator-set stuff yet, so you may want to loop in @aszepieniec  to review this pr also.

dan-da (changes requested):
CI is reporting a compile error with latest changes.  (doctest, I believe)

```
test src/models/state/mempool.rs - models::state::mempool::Mempool::get_sorted_iter (line 414) ... FAILED
test src/locks/tokio/atomic_rw.rs - locks::tokio::atomic_rw::AtomicRw<T>::lock_mut (line 261) ... ok

failures:

---- src/models/state/mempool.rs - models::state::mempool::Mempool::get_sorted_iter (line 414) stdout ----
error[E0061]: this function takes 2 arguments but 1 argument was supplied
Error:   --> src/models/state/mempool.rs:419:15
   |
8  | let mempool = Mempool::new(ByteSize::gb(1));
   |               ^^^^^^^^^^^^----------------- an argument of type `neptune_core::prelude::triton_vm::prelude::Digest` is missing
   |
note: associated function defined here
  --> /home/runner/work/neptune-core/neptune-core/src/models/state/mempool.rs:84:12
   |
84 |     pub fn new(max_total_size: ByteSize, tip_digest: Digest) -> Self {
   |            ^^^
help: provide the argument
   |
8  | let mempool = Mempool::new(ByteSize::gb(1), /* neptune_core::prelude::triton_vm::prelude::Digest */);
   |                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

src/util_types/mutator_set/removal_record.rs
jan-ferdinand: Something along the following lines could work:

```rust
    pub fn split_by_activity(
        &self,
        mutator_set: &MutatorSetAccumulator,
    ) -> Result<(HashMap<u64, Vec<u128>>, Vec<u128>), MutatorSetError> {
        let (aw_chunk_index_min, aw_chunk_index_max) = mutator_set.active_window_chunk_interval();

        let (inactive, active): (HashMap<_, _>, HashMap<_, _>) = indices_to_hash_map(&self.0)
            .into_iter()
            .partition(|&(chunk_index, _)| chunk_index < aw_chunk_index_min);
        let active = active.into_values().flatten().collect_vec();

        if let Some((&chunk_index, _)) = inactive.iter().find(|(&i, _)| i > aw_chunk_index_max) {
            return Err(MutatorSetError::AbsoluteRemovalIndexIsFutureIndex {
                current_max_chunk_index: aw_chunk_index_max,
                saw_chunk_index: chunk_index,
            });
        }

        Ok((inactive, active))
    }
```

By the way, I agree with clippy: the return type is quite complex.

src/models/state/mempool.rs
Sword-Smith: > Is is it expected/planned that we will eventually be able to rollback transaction removal record integrity proofs? ie, is it "just" an engineering issue or a theoretical one?

This limitation can certainly be solved through "relatively" simple engineering. The easiest way to solve it would be for each node to store multiple copies of the same transaction, with a copy for the last $n$ blocks. So if a reorganization happens, we can take a copy of the transaction at the latest block that both branches agree on and start from there.

This, however, does not solve the problem of adding to the mempool transactions that were mined on the abandoned fork and now would need to be re-added to the mempool.

Consider a scenario where tip changes from `3a` to `3b`
```text
                                      +-- block 2a <-- block 3a     
                                      |                          
 Genesis <-- block 1 <--+                            
                                      |                           
                                     +-- block 2b <-- block 3b
```

The transactions in the mempool at the time of `3a` can be updated to be valid for `3b` provided the client knows a version of the transaction that was valid at block `1`. The transactions that are no longer mined since block `2a` and `3a` were abandoned would also have to be stored by the client in a version valid at block `1`. So it's certainly possible to preserve the mempool after a reorganization that's only a few blocks deep, but you pay a quite heavy price in terms of RAM consumption for this.

src/util_types/mutator_set/removal_record.rs
Sword-Smith: I removed the mutability and adapted most of Ferdinand's suggestion. I'm keeping the return type though, as it in my opinion is not overly complicated (I think clippy recently made the rules for "complex" return types stricter). We don't yet have benchmarks for the mutator sets, so I can't tell which one is faster. If I had to take a guess they are equally fast though. Also: Since this all relates to values that live in RAM, this function is very unlikely to be any kind of bottleneck.

jan-ferdinand: > I think clippy recently made the rules for "complex" return types stricter

The only hits for `type_complexity` in the [clippy changelog](https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md) are from rust v1.59 from 2022-02-24 and rust v1.31 from 2018-12-06, both of which fix false positive, neither of which touches the default threshold for that lint. I conclude that no, clippy has not gotten more sensitive on this.

Note that a `type` alias is often sufficient. Grepping through the codepase, `HashMap<u64, Vec<u128>>` seems to be a thing. Giving that thing a name could be beneficial not just to calm clippy, but to signify what that thing is.

Sword-Smith:
I added a document about the program's tolerance to reorganizations. Please review.

Sword-Smith:
Apparently my `make all` does not do the same checks as the CI. I'll fix that and update the PR.

Sword-Smith:
Fixed the discrepancy between `make all` and CI by adding `cargo t --doc` to make's `test` recipe.

src/models/state/mempool.rs
dan-da: > This, however, does not solve the problem of adding to the mempool transactions that were mined on the abandoned fork and now would need to be re-added to the mempool

Good point.  

> So it's certainly possible to preserve the mempool after a reorganization that's only a few blocks deep, but you pay a quite heavy price in terms of RAM consumption for this

My thought would be that whenever we remove utxo(s) from the mempool because they were confirmed in a block, we could write them to disk, indexed by block hash.  We could perhaps keep such mempool-blocks to a depth of 100 (or some configurable amount).   We would only need to consult the on-disk store during a re-org and would not be using any additional RAM.   And only re-orgs deeper than 100 would lose tx, which hopefully never happen anyway.

developer_docs/reorganizations.md
```diff
@@ -0,0 +1,51 @@
...
+The wallet handles transactions that the client holds the spending keys for.
+The light state contains the latest blocks which verifies the validity of the
+entire history of the blockchain. The archival state is optional and allows,
+among other things, the client to re-synchronize wallets that are no longer
+up-to-date. The mempool keeps track of transactions that are not yet included
+in blocks, thus allowing miners to confirm transactions by picking some from
+the mempool to include in the next block.
+
+### Wallet
+The wallet can handle reorganizations that are up to `n` blocks deep, where `n`
```
dan-da, line 32: What happens if my wallet has a (monitored) utxo that gets dropped from the mempool during re-org?    iiuc, it seems the wallet would be out-of-sync with the mempool, but not the wallet's fault.

developer_docs/reorganizations.md
```diff
@@ -0,0 +1,51 @@
+# How different parts of the state handle reorganizations
```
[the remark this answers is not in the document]
Sword-Smith: Picked the 1st suggestion.

developer_docs/reorganizations.md
```diff
@@ -0,0 +1,51 @@
...
+
+## State overview
+The client's state consists of the following parts:
+- wallet
+- light state
+- archival state (optional)
+- mempool
+
+The wallet handles transactions that the client holds the spending keys for.
+The light state contains the latest blocks which verifies the validity of the
```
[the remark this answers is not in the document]
Sword-Smith: accepted.

developer_docs/reorganizations.md
```diff
@@ -0,0 +1,51 @@
...
+entire history of the blockchain. The archival state is optional and allows,
+among other things, the client to re-synchronize wallets that are no longer
```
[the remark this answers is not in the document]
Sword-Smith: monitored UTXOs have to be kept up to date. That's what I'm referring to here.

developer_docs/reorganizations.md
Sword-Smith: The wallet currently *only* deals with and displays transactions that have been mined. It completely ignores anything that's in the mempool.

Sword-Smith: Does that address your question? If not, please elaborate :)

src/models/state/mempool.rs
Sword-Smith: Currently the mempool *only* lives in memory and is never persisted on disk. I'm pretty happy with that quality but I guess it's not set in stone. Feel free to open an issue about the limited reorganization capabilities of the mempool.

Sword-Smith:
This closes #145.

developer_docs/reorganizations.md
dan-da: > It completely ignores anything that's in the mempool.

Ok, so that implies that spends are not reflected in the wallet prior to confirmation either.

Eg, when I send a payment P for amount A, and P is not yet confirmed in a block:
1. A is not reflected in the wallet balance.
2. P is not shown in the wallet's transaction history.
3. The wallet would let me attempt to re-spend P's inputs because it is unaware that those utxo are already in the mempool.   But any such attempt would fail, because spend() actually does check the mempool.

Is the above accurate?

If so, then I think this behavior deserves a mention in the docs somewhere also, until such time as the wallet becomes mempool-aware.

src/models/state/mempool.rs
dan-da: yeah, so I would look at it as the mempool remains unchanged, with everything in mem, but we add a new thing, which might be called mempool-archive, that would be stored on disk, only for re-org usage.

ok, I will open an issue for it to remind us.

PR #159, opened 2024-06-07
refactor create_transaction,  avoid holding lock across prove()

### LINKED ISSUE #122: Make long running CPU operations async friendly

dan-da:
Closes #122.   Obsoletes #136 

This PR builds on #136 which modifies GlobalState::create_transaction() to take `&self` instead of `&mut self` and makes the state update (for expected_utxo) responsibility of the caller.  The primary motivation is to avoid holding a write lock across the call to prove() which may take a very long time.

This PR focuses on improving the flexibility and ergonomics of the create_transaction() API.

In particular, it improves on #136 in the following ways:
 * handles the general case of `notify wallet of any incoming utxos` instead of the sub-case of `notify wallet of incoming change utxo` -- as discussed with @aszepieniec.
 * enables the caller to individually specify that each output utxo use either on-chain or off-chain notification, and to easily default to off-chain for utxos that our wallet can claim and on-chain for all other utxos.
 * create_transaction() now returns a tuple (Transaction, TransactionDetails) where the latter contains all the tx components, including ExpectedUtxos, which caller must use to inform the wallet.
 * adds rpc endpoint `send_to_many()` which exposes multi-output functionality at the rpc level.    The `send()` endpoint now just wraps `send_to_many()`.
 * adds a test case for `send_to_many()` RPC
 * adds a mock MainLoop channel endpoint for `send_to_many()` to transmit to, which avoids an error case that was previously not logged and now triggered by the test case.
 * adds `UtxoNotifyMethod` and `ChangeNotifyMethod` enums for selecting `OnChainPubKey` and `OffChain` notifications:   aka `PublicAnnouncement` and `ExpectedUtxo`.  It also has a variant `OnChainSymetricKey` in preparation for when we add symmetric-key notifications as alternative to OffChain.
 * Simplifies creation of `UtxoReceiver` struct, which represents each output passed to `create_transaction()`.

These changes are a bit experimental.  I think they are a net overall improvement to the API, but am open to criticisms/suggestions.

I should point out that the ability to select onchain/offchain notification per output utxo is available at the level of the rust API, ie `create_transaction()`, but is *not* available in the RPC API, which just uses the defaults.  So this functionality is not presently exposed to end-users.

One further change worth mentioning:  this eliminates the pause/unpause miner steps inside send/send_to_many rpc calls.  As I see it, a new block could arrive at any moment from a peer while we are creating/proving, so there is nothing special about mining in that regard.   Maybe there's an argument to be made that we can have CPU contention and the Tx will take longer to prove(), however I don't think that is valid right now while mining is single-threaded.   I am open to putting this back if it should prove necessary, but right now I don't see a need for the added complexity.

For purposes of reviewing, I would recommend to start with rpc_server.rs to see how usage has changed, then models/state/mod.rs and models/state/utxo_receiver.rs for most impl changes.  Other files are mostly unit test adaptations.

- [x] TODO:  I just realized I also should add send_to_many command to neptune-cli.  That can happen in a follow-up commit.  (done in 592ff16)

ALSO:  I noticed that our wallet is always just using derivation index 0.  I wrote a couple wallet functions that pretend this is not the case, ie `is_wallet_utxo()` and `get_known_spending_keys()`.    This seems an area that needs attention, ie the wallet should keep track (state) of the last-used key index and increment each time a new wallet address is requested.

### DIFF
--- a/src/models/state/mod.rs
+++ b/src/models/state/mod.rs
@@ -567,8 +567,8 @@ impl GlobalState {
     /// creates a Transaction.
     ///
     /// UtxoReceiver::auto() should normally be used.  This will generate
-    /// OffChain notifications for UTXOs destined for our wallet and OnChain
-    /// notifications for all other UTXOs.
+    /// OffChain notifications for UTXOs destined for our wallet and
+    /// OnChainPubKey notifications for all other UTXOs.
     ///
     /// Likewise ChangeNotifyMethod::default() should be used, which corresponds
     /// to ChangeNotifyMethod::OffChain.  This controls notification behavior
@@ -652,7 +652,7 @@ impl GlobalState {
         let mut public_announcements = receiver_data
             .iter()
             .filter_map(|rd| match &rd.utxo_notify_method {
-                UtxoNotifyMethod::OnChain(pa) => Some(pa.clone()),
+                UtxoNotifyMethod::OnChainPubKey(pa) => Some(pa.clone()),
                 _ => None,
             })
             .collect_vec();
@@ -675,7 +675,8 @@ impl GlobalState {
 
             match change_notify_method {
                 ChangeNotifyMethod::OffChain => expected_utxos.push(change_expected_utxo),
-                ChangeNotifyMethod::OnChain => {
+                ChangeNotifyMethod::OnChainSymmetricKey => unimplemented!(),
+                ChangeNotifyMethod::OnChainPubKey => {
                     public_announcements.push(change_public_announcement)
                 }
             }
@@ -1418,7 +1419,7 @@ mod global_state_tests {
         let public_announcement = recipient_address
             .generate_public_announcement(&output_utxo, sender_randomness)
             .unwrap();
-        let receiver_data = vec![UtxoReceiver::onchain(
+        let receiver_data = vec![UtxoReceiver::onchain_pubkey(
             output_utxo.clone(),
             sender_randomness,
             receiver_privacy_digest,
@@ -1501,7 +1502,7 @@ mod global_state_tests {
                 .generate_public_announcement(&utxo, other_sender_randomness)
                 .unwrap();
             output_utxos.push(utxo.clone());
-            other_receiver_data.push(UtxoReceiver::onchain(
+            other_receiver_data.push(UtxoReceiver::onchain_pubkey(
                 utxo,
                 other_sender_randomness,
                 other_receiver_digest,
--- a/src/models/state/mod.rs
+++ b/src/models/state/mod.rs
@@ -54,12 +54,13 @@ pub mod shared;
 mod utxo_receiver;
 pub mod wallet;
 
+pub use utxo_receiver::ChangeNotifyMethod;
 pub use utxo_receiver::UtxoNotifyMethod;
-pub use utxo_receiver::UtxoReceiverData;
+pub use utxo_receiver::UtxoReceiver;
 
 #[derive(Debug, Clone)]
-pub struct TransactionData {
-    pub receiver_data: Vec<UtxoReceiverData>,
+pub struct TransactionDetails {
+    pub receiver_data: Vec<UtxoReceiver>,
     pub removal_records: Vec<RemovalRecord>,
     pub addition_records: Vec<AdditionRecord>,
     pub input_utxos: Vec<(Utxo, LockScript, MsMembershipProof)>,
@@ -431,7 +432,7 @@ impl GlobalState {
 
     /// Given a list of UTXOs with receiver data, generate the corresponding
     /// addition records.
-    pub fn generate_addition_records(receiver_data: &[UtxoReceiverData]) -> Vec<AdditionRecord> {
+    pub fn generate_addition_records(receiver_data: &[UtxoReceiver]) -> Vec<AdditionRecord> {
         receiver_data
             .iter()
             .map(|rd| {
@@ -448,7 +449,10 @@ impl GlobalState {
     /// in input amount and output amount goes back to us.
     /// The caller should also notify the wallet to expect the change Utxo
     /// so that we can synchronize it after it is confirmed.
-    pub async fn add_change(&self, change_amount: NeptuneCoins) -> ExpectedUtxo {
+    pub async fn add_change(
+        &self,
+        change_amount: NeptuneCoins,
+    ) -> Result<(ExpectedUtxo, PublicAnnouncement)> {
         // generate utxo
         let own_spending_key_for_change = self
             .wallet_state
@@ -471,12 +475,17 @@ impl GlobalState {
 
         let receiver_preimage = own_spending_key_for_change.privacy_preimage;
 
-        ExpectedUtxo::new(
+        let public_announcement =
+            own_receiving_address.generate_public_announcement(&utxo, sender_randomness)?;
+
+        let expected_utxo = ExpectedUtxo::new(
             utxo,
             sender_randomness,
             receiver_preimage,
             UtxoNotifier::Myself,
-        )
+        );
+
+        Ok((expected_utxo, public_announcement))
     }
 
     /// Generate a primitive witness for a transaction from various disparate witness data.
@@ -522,7 +531,7 @@ impl GlobalState {
     /// from some selection of owned UTXOs.
     async fn create_transaction_from_tx_data(
         &self,
-        tx_utxo_data: TransactionData,
+        tx_utxo_data: TransactionDetails,
     ) -> Result<Transaction> {
         let mutator_set_accumulator = self
             .chain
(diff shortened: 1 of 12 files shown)


aszepieniec:
The name `ChangeNotifyMethod` is a little confusing because:

 - Non-change UTXOs can also have recipients that need to be notified. Is there a good reason to limit the struct's scope to exclude them?
 - I don't foresee a world in which `OnChainPubKey` would be used for *change*. If it's going back to yourself, you don't need public key crypto and symmetric crypto suffices.

Maybe just `NotifyMethod` suffices? That's more of an invitation to elaborate than a suggestion.

src/models/state/mod.rs
```diff
@@ -37,25 +20,57 @@ use super::blockchain::type_scripts::TypeScript;
...
+    pub receiver_data: Vec<UtxoReceiver>,
+    pub removal_records: Vec<RemovalRecord>,
+    pub addition_records: Vec<AdditionRecord>,
+    pub input_utxos: Vec<(Utxo, LockScript, MsMembershipProof)>,
+    pub output_utxos: Vec<Utxo>,
+    pub public_announcements: Vec<PublicAnnouncement>,
+    pub expected_utxos: Vec<ExpectedUtxo>,
+    pub fee: NeptuneCoins,
+    pub timestamp: Timestamp,
+}
```
aszepieniec, line 72: There seems to be a lot of overlap with `TransactionKernel` and `transaction::primitive_witness::PrimitiveWitness`. Let's a) avoid code duplication, and b) use names specific enough to meaningfully distinguish similar models.

In my mental model:
 - A `TransactionKernel` is the unit of data that appears on the blockchain and effects a transfer of tokens. It is generally circulated with a proof in which case the thing is upgraded to a `Transaction`. The `proof` field of a `Transaction` certifies the validity of the `TransactionKernel`.
 - A `PrimitiveWitness` is all the data you need to generate a transaction proof in the base case (*i.e.*, no recursion. Where the `TransactionKernel` contains commitments, the `PrimitiveWitness` contains the preimages to these commitments.

I'm not sure what the purpose of `TransactionDetails` is so I can't comment on whether it needs all the fields it has or what a better name might be. Please elaborate.

src/models/state/wallet/wallet_state.rs
```diff
@@ -286,6 +285,29 @@ impl WalletState {
...
+
+    // TODO: These spending keys should probably be derived dynamically from some
+    // state in the wallet. And we should allow for other types than just generation
+    // addresses.
+    //
+    // Probably the wallet should keep track of index of latest derived key
+    // that has been requested by the user for purpose of receiving
+    // funds.  We could also perform a sequential scan at startup (or import)
+    // of keys that have received funds, up to some "gap".  In bitcoin/bip32
+    // this gap is defined as 20 keys in a row that have never received funds.
```
aszepieniec, line 305: Performing a sequential scan at startup or import is a fantastic idea. I think 20 is a fine default value for the gap but this value should be overridable through command line arguments.

dan-da:
yeah, I'm not yet really satisfied with `UtxoNotifyMethod` and `ChangeNotifyMethod` as they both have the same variants.

The reason they are distinct is that `UtxoNotifyMethod` is used by the caller to actually provide `PublicAnnouncement` to `create_transaction()`.   And that is done because `create_transaction()` deals with Utxo and PublicAnnouncement, not GenerationAddress.   That keeps `create_transaction()` agnostic about the type of address used, which is good, I think.  (and preserves prior behavior)

`ChangeNotifyMethod` is used only for telling create_transaction() how to notify for the change output.  In this case, create_transaction() generates the PublicAnnouncement itself, using a SpendingKey.  So that kind of breaks the above model of being address-type agnostic... but then that code pre-exists this PR.   This is only necessary because create_transaction() creates the change utxo if sum(inputs) > sum(outputs).   Perhaps a cleaner/better solution would be to return an error in this case, which forces the caller to create the correct change utxo itself.   With that modification, there is nothing special about the change utxo, it is an output created by the caller like any other.  I am leaning toward this solution.

> The name `ChangeNotifyMethod` is a little confusing because:

yeah, we can get rid of that enum entirely with above proposal.

> * Non-change UTXOs can also have recipients that need to be notified. Is there a good reason to limit the struct's scope to exclude them?

Those utxo are taken care of by the `UtxoNotifyMethod` enum.  With above proposal, it would cover all utxo, including change.

> * I don't foresee a world in which `OnChainPubKey` would be used for _change_. If it's going back to yourself, you don't need public key crypto and symmetric crypto suffices.

Fair point.  Though it also applies to any output UTXO that can be claimed/spent by "my" wallet.    So I think that create_transaction() should just throw an error in that case.

Does the above sound reasonable to you?

src/models/state/mod.rs
dan-da: Sure.

`TransactionDetails` started as a private struct for passing data between some of the create_transaction() worker methods, which previously just used tuples. In particular `generate_tx_data_for_transaction()` returns a `TransactionDetails` to `create_transaction()` which then passes it to `create_transaction_from_tx_data()`.

I decided to make it public and return it from `CreateTransaction` because it contains some extra data that is not available in the `Transaction` itself. 

But yes you are right there is a lot of duplication, and that is not clean, so it is good you mentioned about it.  The fields not in TransactionKernel are:

```
    pub receiver_data: Vec<UtxoReceiver>,
    pub input_utxos: Vec<(Utxo, LockScript, MsMembershipProof)>,
    pub output_utxos: Vec<Utxo>,
    pub expected_utxos: Vec<ExpectedUtxo>,
```

The only required field for present usage is `unexpected_utxos`, which is necessary for the caller to inform the wallet of these.  So at simplest we could have `create_transaction()` return (Transaction, Vec&lt;ExpectedUtxo>) instead and perhaps that is best for now.

I think `TransactionDetails` could remain, only as an implementation detail private to the module.

dan-da: question:  would it be a bad idea to add an expected_utxos field to `Transaction` itself?   Then create_transaction() could just return Transaction, as before, and that would be much cleaner.

Previously I was thinking I shouldn't touch Transaction as it goes on the wire, but you are saying that really it is TransactionKernel on the wire, so maybe it's ok?

and/or am open to any suggestion that makes the create_transaction() api cleaner.

dan-da:
closing.  obsoleted by #162.

PR #160, opened 2024-06-11
Add MdBook for Documentation

aszepieniec:
- Add mdBook for documentation
 - Migrate existing documentations there
 - Add documentation for consensus logic wrt transactions


Sword-Smith:
> `RemovalRecordsIntegrity :: (transaction_kernel_mast_hash : Digest) ⟶ (inputs_salted_utxos_hash : Digest)`

Is the `RemovalRecordsIntegrity` program supposed to output something? I don't see that being the case currently. But maybe that's just in the pipeline?

aszepieniec:
> > `RemovalRecordsIntegrity :: (transaction_kernel_mast_hash : Digest) ⟶ (inputs_salted_utxos_hash : Digest)`
> 
> Is the `RemovalRecordsIntegrity` program supposed to output something? I don't see that being the case currently. But maybe that's just in the pipeline?

Yes, although an equally valid design would have been to take `input_salted_utxos_hash` as an argument and authenticate it. We still need to update the code.

docs/src/consensus/transaction.md
```diff
@@ -0,0 +1,176 @@
...
+     - collect the type script hash
+   - filter out duplicates
+   - output the unique type script hashes
+ - `TypeScript :: (transaction_kernel_mast_hash : Digest) ⟶ ∅` Authenticates the correct evolution of all UTXOs of a given type. The concrete program value depends on the token types involved in the transaction. For Neptune's native currency, Neptune Coins, the type script asserts that *a)* all output amounts are positive, and *b)* the sum of all input amounts is greater than or equal to the fee plus the sum of all output amounts.
+
+All subprograms can be proven individually given access to the transaction's witness.
+
+Note that none of the subprograms require that each removal record lists one SWBF index that does not yet live in the mutator set SWBF. This absence is required for the transaction to be confirmable, but not for it to be valid. If the transaction has an input whose index set is already entirely contained by the mutator set SWBF, then this transaction can never be confirmed. Even if there is a reorganization that results in the absence criterion being satisfied, the transaction commits to the mutator set hash and this commitment cannot be undone.
```
Sword-Smith, line 91: Doesn't `RemovalRecordsIntegrity` require that at least one index has not been seen before?

docs/src/consensus/transaction.md
```diff
@@ -0,0 +1,176 @@
...
+   - divine the salt for input UTXOs
+   - authenticate the salted input UTXOs against the given hash digest
+   - divine all output UTXOs
+   - divine the salt for output UTXOs
+   - authenticate the salted output UTXOs against the given hash digest
+   - for each input or output UTXO:
+     - collect the type script hash
+   - filter out duplicates
+   - output the unique type script hashes
+ - `TypeScript :: (transaction_kernel_mast_hash : Digest) ⟶ ∅` Authenticates the correct evolution of all UTXOs of a given type. The concrete program value depends on the token types involved in the transaction. For Neptune's native currency, Neptune Coins, the type script asserts that *a)* all output amounts are positive, and *b)* the sum of all input amounts is greater than or equal to the fee plus the sum of all output amounts.
```
Sword-Smith, line 87: Do we need a `TypeScript` execution proof for *each* digest returned by `CollectTypeScripts`? That doesn't seem clearly specified.

docs/src/consensus/transaction.md
aszepieniec: Let me clarify: program `RemovalRecordsIntegrity` requires that every removal record in the transaction adds at least one new index to the set of indices announced by that transaction, not to the set of indices contained by the SWBF.

docs/src/consensus/transaction.md
aszepieniec: Yes. Thanks; will clarify.

docs/src/consensus/transaction.md
```diff
@@ -0,0 +1,178 @@
...
+
+A transaction is valid if another transaction that is identical except for fixing an older mutator set hash or timestamp, was valid. Specifically, the program The program `TransactionDataUpdate :: (transaction_kernel_mast_hash : Digest) ⟶ ∅` verifies the update of transaction data as follows:
+ - divine `tx : TransactionKernel`
+ - verify `tx` with some divined proof
+ - create a new `TransactionKernel` object `kernel`
+ - set all fields of `kernel` to the matching field of `tx` except:
+   - set `kernel.timestamp` such that `kernel.timestamp >= tx.timestamp`
+   - set `kernel.mutator_set_hash =/= tx.mutator_set_hash` only if the following instructions execute gracefully without crashing
+     - divine the mutator set AOCL MMR accumulator `kernel_aocl`
+     - authenticate `kernel_aocl` against the mutator set MAST hash `kernel.mutator_set_hash` using a divined authentication path
```
Sword-Smith, line 173: Do we also need to deal with the removal records here?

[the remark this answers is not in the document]
aszepieniec: No, we are authenticating just the AOCL MMR, and the root for this MAST path is the MS MAST hash. The confusion arises from the fact that this mutator set hash lives on the transaction kernel, which lends name but has a wholly different MAST hash.

Sword-Smith: The removal records contain authentication paths into the SWBF MMR. Those authentication paths need to be updated whenever the tip is changed.

PR #165, opened 2024-07-17
docs: danda's first read edits

dan-da:
I read through the mdbook doc for the first time and made some edits that bring it more up to date with current codebase.

A few explanatory comments here:

re tarpc:  HTTP/JSON implies that a client could connect from any language or a web browser, but that is not correct.   We do not use http as transport.  Rather we use tarpc's serde_transport to transmit JSON.  Clients must also use rust+tarpc (unless they reverse engineer tarpc::serde_transport)

re threads/tasks/g in neptune-core overview:  It is incorrect nomenclature to call these threads.  They have never been threads.  Rather they are tokio spawned *tasks* running on tokio's threadpool.  They may run on the same or different operating system thread from the parent task, at tokio's discretion.  Unfortunately there are still some comments and variables in the code that refer to tasks as threads, which may perpetuate the misconception.  When I have time I will make a commit to correct this.

re canonical ordering of locks:  That concept existed in order to acquire multiple locks in a certain order to avoid deadlock, but now there is only a single RwLock, that can only ever be write-acquired by one party at a time, so no need for ordering.

re deadlocks:  now that we have a single lock over globalstate it's pretty hard to deadlock.  you almost have to do it intentionally.  updated docs with some simple rules.


aszepieniec (approved):
Looks good to me. Note that branch [`asz/transaction-consensus`](https://github.com/Neptune-Crypto/neptune-core/tree/asz/transaction-consensus) has more documentation in case you are interested.

docs/src/neptune-core/overview.md
```diff
@@ -67,25 +66,38 @@ Different tasks are scheduled in the main loop every N seconds. These currently
...
+
+  1. avoid deadlocking yourself. If a function has read-acquired the global lock then it must be released before write-acquiring. Likewise never attempt to write-acquire the lock twice.
+
+  2. avoid deadlocking others. Always be certain that the global lock will be released in timely fashion. In other words if you have some kind of long running task with an event loop that needs to acquire the global lock, ensure that it gets acquired+released inside the loop rather than outside.
+
 - Atomic writing to databases: The archival mutator set is spread across multiple databases due to how the underlying data structures are defined. If one of the databases are updated but the other is not, this will leave the archival mutator set in an invalid state. We could fix this by allowing an archival mutator set to be stored in only one database. We should also add logic to rebuild the archival mutator set state from the `block_index` database and the blocks stored on disk since it can be derived from the blocks. This functionality could be contained in a separate binary, just like we have a binary for the CLI interface in the form of the RPC client.
```
Sword-Smith, line 101: We manage the entire mutator set through *one* database now, so the 1st sentence is now incorrect. However, this warning/problem is true for the `ArchivalState`, so maybe we can just replace "archival mutator set" with "archival state" here, and it still holds?

Please note, I am not advocating or suggesting any change to the code. I'm happy with how the state looks, this is just about the documentation.

docs/src/neptune-core/overview.md
```diff
@@ -67,25 +66,38 @@ Different tasks are scheduled in the main loop every N seconds. These currently
...
 
 ## Challenges
-- Deadlocks. Solution: always acquire locks in the same order. Note though that locks from `std::sync` may not be held over an `await`. The linter should tell you if you do this. When a function requires more than one lock, **the only correct ordering in which to acquire these locks is the order in which the fields are defined in `GlobalState`. Any deviation from this is a bug.**
+
+- Deadlocks. We only have a single RwLock over the GlobalState. This is encapsulated in struct `GlobalStateLock`. This makes deadlocks pretty easy to avoid, following some simple rules:
+
+  1. avoid deadlocking yourself. If a function has read-acquired the global lock then it must be released before write-acquiring. Likewise never attempt to write-acquire the lock twice.
+
+  2. avoid deadlocking others. Always be certain that the global lock will be released in timely fashion. In other words if you have some kind of long running task with an event loop that needs to acquire the global lock, ensure that it gets acquired+released inside the loop rather than outside.
```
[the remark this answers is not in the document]
dan-da: well it could just be temporarily blocking if the lock eventually gets released.  But I was trying to warn against a task with a "forever loop" that acquires the lock before the loop, and thus never releases it, and effectively deadlocks the program should any other task ever try to acquire it.

also, I like the symmetry of 1. yourself, and 2. others.

So I'm inclined to leave as-is, unless you have a suggestion how to make it clearer?

docs/src/neptune-core/overview.md
dan-da: ok, I changed it.

Sword-Smith: Let me add a few considerations:

Atomic writing to databases: neptune-core presently writes to the following databases: wallet_db, block_index_db, archival_mutator_set, peer_state. If one of the databases are updated but the other is not, this can leave data in an invalid state. We could fix this by storing all state in a single transactional database *but that might make the code base less modular.*

note: We should also add logic to rebuild the archival state from the block_index_db and the blocks stored on disk since it can be derived from the blocks. This functionality could be contained in a separate binary, just like we have a binary for the CLI interface in the form of the RPC client, *or this could be a check that was performed at startup.*

dan-da: Ok, done. I shortened the last sentence of 2nd para to:

This functionality could be contained in a separate binary or a check could be performed at startup.

PR #170, opened 2024-07-24
Review Changes Resulting from Dropping `leaf_index` from `MmrMembershipProof`

aszepieniec:
This is an artificial PR designed to focus reviewer attention on specific changes. Merger or non-merger is irrelevant; and branch `asz/transaction-consensus` will continue to be the most up-to-date topic branch.

This PR captures a sequence of commits following a dependency update. The new dependency on `tasm-lib` master points to `triton-vm` v42-alpha6, which in turn points to the most up-to-date version of `twenty-first`, which drops the field `leaf_index` from `MmrMembershipProof`. This last elision is what required a lot of attention to a) make compile and b) fix consequently failing tests.

A lot of the work to anticipate this change was already done, but you really only know the impact of a change once things start breaking down.

Here is what I want the reviewer's attention on: a lot of the failing tests deal with mutator sets somehow. The tests pass now, but I'm not sure I fixed them in the right way. Divining their intention was difficult, the mechanics of batch applying records with batches of to-be-preserved membership proofs is notoriously difficult. I'm afraid I *a)* neutered tests in otherwise benign ways; or worse *b)* neutered tests while breaking the (up till now) (probably) correct mutator set logic.


src/util_types/mutator_set/archival_mutator_set.rs
```diff
@@ -88,8 +88,8 @@ where
         self.accumulator().await.hash()
     }
 
-    /// Apply a list of removal records while keeping a list of mutator set membership proofs
-    /// up-to-date
+    /// Apply a list of removal records while keeping a list of mutator set
+    /// membership proofs up-to-date.
     pub async fn batch_remove(
```
Sword-Smith, line 93: I don't like how this function uses the mutator set accumulator as a library. Seems sloppy but the changes look correct to me. Though, it's, due to multiple rewrites, a rather confusing implementation.

src/util_types/mutator_set/ms_membership_proof.rs
```diff
@@ -99,24 +96,55 @@ impl MsMembershipProof {
...
+            &mutator_set.aocl.peaks(),
         );
 
         // if window does not slide, we are done
         if !MutatorSetAccumulator::window_slides(new_item_index) {
-            return Ok(indices_for_updated_mps);
+            return Ok(indices_for_mps_updated_from_append);
         }
 
+        assert!(membership_proofs
```
Sword-Smith, line 134: Again?

src/util_types/mutator_set/ms_membership_proof.rs
```diff
@@ -201,38 +232,45 @@ impl MsMembershipProof {
...
-        let mut mmr_membership_proof_index_to_membership_proof_index: Vec<usize> = vec![];
+        let mut mmr_mp_index_to_ms_mp_index = vec![];
+        let mut mmr_membership_indices = vec![];
         for (i, mp) in membership_proofs.iter_mut().enumerate() {
             if mps_for_batch_append.contains(&i) {
-                for (_, (mmr_mp, _chnk)) in mp.target_chunks.iter_mut() {
-                    mmr_membership_proofs_for_append.push(mmr_mp);
-                    mmr_membership_proof_index_to_membership_proof_index.push(i);
+                for (chunk_index, (mmr_mp, _chunk)) in mp.target_chunks.iter_mut() {
+                    if *chunk_index != old_window_start_batch_index {
```
Sword-Smith, line 244: This condition seems new. I think it's covered by line 203: `if chunk_index < old_window_start_batch_index`

src/util_types/mutator_set/mutator_set_accumulator.rs
```diff
@@ -674,10 +689,13 @@ mod ms_accumulator_tests {
                     archival_after_remove.add(&addition_record).await;
                     archival_before_remove.add(&addition_record).await;
 
-                    let updated_mp_indices = update_result.unwrap();
                     println!("{}: Inserted", i);
                     for j in 0..items.len() {
-                        if updated_mp_indices.contains(&j) {
+                        if indices_of_updated_mps.contains(&j) {
+                            assert_ne!(
+                                previous_mps[j], membership_proofs_batch[j],
```
Sword-Smith, line 696: nice addition

src/util_types/mutator_set/removal_record.rs
```diff
@@ -254,20 +255,25 @@ impl RemovalRecord {
         let mut mmr_membership_proofs_for_append: Vec<
             &mut mmr::mmr_membership_proof::MmrMembershipProof<Hash>,
         > = vec![];
+        let mut leaf_indices = vec![];
         for (i, rr) in removal_records.iter_mut().enumerate() {
             if rrs_for_batch_append.contains(&i) {
-                for (_, (mmr_mp, _chnk)) in rr.target_chunks.iter_mut() {
-                    mmr_membership_proofs_for_append.push(mmr_mp);
+                for (chunk_index, (mmr_mp, _chnk)) in rr.target_chunks.iter_mut() {
+                    if *chunk_index != old_window_start_batch_index {
```
Sword-Smith, line 262: Again: Not sure about the need for this check. Covered by line 227: `if chunk_index < old_window_start_batch_index`?

Sword-Smith (changes requested):
There are two sanity checks implemented as asserts that I disagree with. And two conditional checks that I think can be removed. Otherwise it looks correct.

Cargo.toml
```diff
@@ -123,5 +123,4 @@ name = "consensus"
 harness = false
 
 [patch.crates-io]
-# 8dfd41266dec89c26498fb48a8ad9e5bc07954c9 is tip of branch: master as of 2024-07-17
-tasm-lib = { git = "https://github.com/TritonVM/tasm-lib.git", rev = "8dfd41266dec89c26498fb48a8ad9e5bc07954c9" }
+tasm-lib = { git = "https://github.com/TritonVM/tasm-lib.git", rev = "121eb10" }
```
dan-da, line 126: nit: it's nice to have a comment about what the revision represents.   eg previously the comment was:

# 8dfd41266dec89c26498fb48a8ad9e5bc07954c9 is tip of branch: master as of 2024-07-17

src/util_types/mutator_set/ms_membership_proof.rs
```diff
@@ -99,24 +96,55 @@ impl MsMembershipProof {
             "Function must be called with same number of membership proofs and items. Got {} items and {} membership proofs", own_items.len(), membership_proofs.len()
         );
 
-        let new_item_index = mutator_set.aocl.count_leaves();
+        assert!(membership_proofs
```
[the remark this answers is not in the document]
aszepieniec: Some of the sanity checks were inserted as a means to track down bugs. In regards to this one, let me see if I can distill it down to a test case.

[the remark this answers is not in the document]
aszepieniec: Actually, I'm not sure how to distill this one into a test case because it occurs at the start of the function body. It seems to me like it should be part of a function interface contract, in which case it is okay to assert it where it is now. I don't know what a test case that could cover it would look like.

All of that said, this sanity check was definitely inserted with the objective of tracking down a bug. I inserted essentially the same block of code in more places in the same function, but later removed those.

src/util_types/mutator_set/ms_membership_proof.rs
aszepieniec: Thanks for spotting. I'm removing this one.

src/util_types/mutator_set/ms_membership_proof.rs
aszepieniec: I think it's not covered because if I drop this condition then tests fail. That said, it certainly looks as though it should be covered by this line. So I am left wondering why it's not and what we are overlooking and whether a soundness or completeness error is lurking there.

src/util_types/mutator_set/removal_record.rs
aszepieniec: Same story: if I remove the condition then tests fail. But I'm puzzled as to why the line you point to doesn't catch it.

PR #174, opened 2024-07-31
rewrite(removal_records_integrity): Don't check for double-spend

Sword-Smith:
With this change, the removal records integrity program no longer checks that the Bloom-filter indices are unset for the input being spent. This change is made to reduce the computational cost of the transaction initiator. This change must be done by the miner.

This commit reduces the computation cost of this program by more than 90 %. It divides the table height by between 11 and 12.


docs/src/consensus/transaction.md
```diff
@@ -62,16 +62,15 @@ The motivation for splitting transaction validity into subclaims is that the ind
...
-     - read the removal record chunks dictionary from memory
-     - for each entry in this dictionary, verify that the chunk belongs to the SWBF MMR from the mutator set accumulator (authentication paths are either read from memory or divined in -- to be decided)
-     - for all indices in the index set, verify that if it is in the inactive part of the SWBF, then it lives in some dictionary entry (chunk)
-     - verify that its AOCL leaf index is unique
-   - hash the list of removal records and authenticate it against the given transaction kernel MAST hash
+     - verify that the calculated removal record index set matches the claimed index set
+   - hash the list of removal record sets and authenticate it against the given transaction kernel MAST hash
    - output the hash of the salted input UTXOs.
+
+   Checks ensuring that each AOCL index is unique and that the published authentication paths are valid are delegated to the miner and do, for performance reasons, not belong here. Checks that the removal record has not already been applied (i.e. no double-spend) is also delegated to the miner.
```
aszepieniec, line 73:
```suggestion
   Checks ensuring that each AOCL index is unique and that the published authentication paths are valid, are delegated to the miner and do, for performance reasons, not belong here. Checks that the removal record has not already been applied (i.e. no double-spend) is also delegated to the miner.
```

src/models/blockchain/transaction/validity/removal_records_integrity.rs
```diff
@@ -380,13 +403,12 @@ impl ConsensusProgram for RemovalRecordsIntegrity {
...
         // iterate over all input UTXOs
-        let mut all_aocl_indices: Vec<u64> = Vec::new();
         let mut input_index: usize = 0;
         while input_index < input_utxos.len() {
             let utxo: &Utxo = &input_utxos[input_index];
             let utxo_hash = Hash::hash(utxo);
-            let msmp: &MsMembershipProof = &rriw.membership_proofs[input_index];
-            let removal_record: &RemovalRecord = &rriw.removal_records[input_index];
+            let msmp = &rriw.partial_membership_proofs[input_index];
+            let claimed_absolute_indices = &rriw.removal_records[input_index].absolute_indices;
```
aszepieniec, line 411: I included the explicit type indications in order to make the rust code compiler-friendly. Since we are writing everything by hand, it is less important. But there is a reason why we might want to keep the rust code compiler-friendly: if we want to verify, as an extra precaution to boost our confidence in the soundness of the consensus programs, then we can test the handwritten against the compiled tasm code.

aszepieniec (approved):
LGTM

Sword-Smith:
> LGTM

I added a new structure `RemovalRecordsIntegrityWitnessMemory` for handling the initial memory layout. And then I moved everything that saved clock cycles from being initialized in memory to the secret-in stream. This saves a few clock cycles in the program and, more importantly, makes the memory layout simpler and easier to reason about.

PR #177, opened 2024-08-26
feat: Add snippet for storing RRI-claim in dyn memory

Sword-Smith:
Here's the pattern I would go with. I think it can be generalized somewhat to handle more claims as well, but I'm not sure that's actually a good idea. Let me know what you think.

### DIFF
--- a/src/models/blockchain/transaction/validity/tasm.rs
+++ b/src/models/blockchain/transaction/validity/tasm.rs
@@ -2,5 +2,6 @@ pub(crate) mod coinbase_amount;
 pub mod compute_indices;
 mod hash_removal_record_indices;
 mod hash_utxo;
+pub mod proof_collection;
 pub mod removal_records_integrity;
 pub mod transaction_kernel_mast_hash;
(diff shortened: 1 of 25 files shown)


aszepieniec (commented):
This PR factors code out in a very specific and non-generic way, with next to no potential for code reuse. If we are going to write a snippet like this for every claim to be verified in the `SingleProof` program, we would end up with a lot of duplicated code which comes with increased the cost of maintenance.

The selling point for this PR, as strong as I can make it, is to make the code more navigable and easy to read. It achieves this by moving the location where existing code lives and moving comments into variable and function names. It does not reduce the scope of logical pathways.

There is a case to be made for moving code and comments in order to benefit readability. However, that transformation needs to take place *after* refactorings that minimize the scope of logical pathways, or else it will be in conflict with them.

Sword-Smith:
Two more claim-generations added to this PR.

There are two benefits of this approach:
1. The tasm code for populating memory with the claims is tested in isolation. So if this breaks, you don't have to go through the entire `SingleProof` program to find the problem. The tests comparing a `BFieldCodec` encoded claim stored to memory is compared with the memory after the program execution on Triton VM.
2. Having a seperate snippet for each claim is a more flexible approach if and when the transaction sub-claim architecture changes.

PR #178, opened 2024-08-26
devops: git-ignore Triton VM proofs for test-runs

Sword-Smith:


Sword-Smith:
You're right. Let's keep LFS-committing them.

Issue #183, opened 2024-09-10
perf: split lengthy mutating methods into prepare() and finalize().


dan-da:
There are some methods that take &mut self and then perform lengthy operations.  This means that they are holding the write-lock and blocking all other tasks from processing until they complete.

The two I am presently looking at are:
  a. GlobalState::resync_membership_proofs_from_stored_blocks(&mut self)
  b. WalletState::update_wallet_state_with_new_block(&mut self)

(a) is run every 59 seconds from a timer.
(b) is run every time a new block is received, approx every 10 minutes.

I was faced with this situation when writing a new method called claim_utxo_for_block().   I solved it by splitting the method into two:
  1. prepare_claim_utxo_in_block(&self) -> updates
  2. finalize_claim_utxo_in_block(&mut self, updates)

(1) takes &self and performs the lengthy operations to gather all the necessary updates.  It does this with only a read-lock so other tasks run currently with no problem.

(2) takes &mut self and the updates returned from (1).  It quickly writes the updates and returns. so it holds the write-lock for the minimum possible time.

I am creating this issue as a reminder to myself to attempt this approach with (a) and (b), and scan for any others that could possibly benefit as well.

PR #186, opened 2024-09-17
split resync membership proofs

dan-da:
addresses #183.

The comment for GlobalStateLock::resync_membership_proofs() explains:
```
    /// This fn performs potentially lengthy operations.  The work has been
    /// split into two: an immutable fn that prepares a batch of updates and a
    /// mutable fn that finalizes (writes) the updates as quickly as possible.
    /// In this way, we do not hold the write-lock any longer than necessary.
    ///
    /// Splitting the logic this way requires buffering updates in RAM.  We batch
    /// the updates to avoid running out of mem.  The batch size is initially
    /// set to 1000.  This is just a wild-ass-guess.  It seems high enough that
    /// most wallets will only need a single batch, but low enough that we
    /// shouldn't run out of mem if processing a huge wallet.
```

### DIFF
--- a/src/models/state/mod.rs
+++ b/src/models/state/mod.rs
@@ -7,6 +7,7 @@ pub mod shared;
 pub mod wallet;
 
 use std::cmp::max;
+use std::collections::BTreeMap;
 use std::ops::Deref;
 use std::ops::DerefMut;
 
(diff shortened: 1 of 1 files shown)


dan-da:
closing in favor of #187.

Issue #189, opened 2024-09-25
create_transaction(), send(), should consider Tx in mempool.


dan-da:
Presently create_transaction() will re-use the same inputs when it is called multiple times in the same block.

eg, consider this loop:

```
    let mut gsm = global_state_lock
        .lock_guard_mut()
        .await;

    for _ in 0..5 {
        let tx = gsm
            .create_transaction(
                &mut tx_outputs,
                change_key,
                UtxoNotifyMethod::OnChain,
                NeptuneCoins::zero(),
                Timestamp::now(),
            )
            .await?;

        gsm.mempool
            .insert(tx)?;
    }
```

What happens is that 5 tx get created that all spend the same input.

(but Mempool::insert() de-dups them into just one tx)

This is because create_transaction() calls `WalletState::allocate_sufficient_input_funds_from_lock()` which fetches unspent utxos that are valid as of tip block.  It does not consider utxos in the mempool at all.

This means that:
 1. create_transaction() will re-use input utxos.
 2. it is not possible to chain unconfirmed tx.   eg alice sends 5 to bob, and 3 change to herself.  From the change alice sends 2 to carol and 1 change to herself.

Fixing this will require that create_transaction() inspects the utxos in the mempool.

dan-da:
hmm, digging into the code, I don't immediately see how to achieve unconfirmed tx chaining.   Tx validation requires that the PrimitiveWitness have one mutator-set memership proof per input.   But input utxos in the mempool are represented as RemovalRecord, with no such membership proof.

I am not certain if it is possible with Neptune's design or not.  a question for @aszepieniec.

Even if unconfirmed tx chaining is not presently achievable, create_transaction() can still avoid adding any input that is already in the mempool.    That would prevent the input re-use.

aszepieniec:
About chaining unconfirmed transactions: it is possible in principle, but requires a substantial amount of engineering. What would need to be done is 1) generalize inputs; and 2) write a new SingleProof path.

1. Generalize inputs. Currently, inputs can only be mutator set removal records. Eventually they should instances of an enum that that can be a) a mutator set removal record (for lockable UTXOs); b) a MMR item (for lock-free UTXOs); or c) unconfirmed transaction output.
[...]

PR #197, opened 2024-10-03
unconfirmed tx chaining.  spend unconfirmed utxos.

dan-da:
Addresses #189

Extends #184.  ( [diff](https://github.com/dan-da/neptune-core/compare/danda/184_unconfirmed_balance_pr...dan-da:neptune-core:danda/189_spend_unconfirmed_utxo) )

This PR implements chaining of unconfirmed transactions.

Or at least it appears to.  I've been told this is not feasible without triton-vm programs. 

So maybe it only appears to work, but is not fully correct.  That's for reviewers to decide, and tell me how the approach is wrong.  Or maybe it is only cut-through that requires tricky triton-vm stuff, and this PR doesn't attempt cut-through.

Primarily for this reason, I am making this a draft PR.   Another reason is that this PR does not (yet?) attempt to handle (or test) the case where unconfirmed tx B depends on A, and B has a higher fee, so it gets confirmed in a block and A does not.  I'm not certain what the code will do... undefined behavior.

The approach taken here is a natural extension of the changes in #184.  Namely the wallet-state keeps track of all wallet-affecting utxos in the mempool. It considers them for balance calculations, for selecting inputs, when scanning for spent-inputs, and when updating wallet state with a new block.

The PR includes a test case, spend_unconfirmed_tx() in wallet_state.rs.  The test case starts with a single input utxo then sends Tx A worth 5, then tx B worth 10 in the same block (B depends on A) and verifies balances, then mines another block and verifies balances.  In detail:

1. mines a block to our wallet, creating confirmed utxo A with 100 tokens.
2. sends 5 coins to an external address spending A and creating unconfirmed utxo B with 95 tokens.
3. sends 10 coins to an external address spending B and creating unconfirmed utxo C with 85 tokens.
4. verifies that confirmed balance is 100 (A) and unconfirmed balance is 85 (C).
5. mines a block to our wallet, confirming B and C, and creating utxo D with 100 tokens, for total confirmed balance of 185 (C+D)
6. confirms that both the unconfirmed and confirmed balance is 185.

steps 5 and 6 were the trickiest as it required some changes to WalletState::update_wallet_state_with_new_block() which is a complex method.

There may be better way(s) to represent the unconfirmed utxos in WalletStatus for more efficient balance calcs.

But for now, I don't want to put further effort into improving unless/until the concept/direction is approved by reviewers.

My thought is that this approach may not be the best/ideal way to achieve this functionality, but if not fatally incorrect then it may be worth having as a stop-gap for now, until a better approach can be implemented.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -700,9 +700,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-channel"
-version = "0.5.12"
+version = "0.5.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ab3db02a9c5b5121e1e42fbdb1aeb65f5e02624cc58c43f2884c6ccac0b82f95"
+checksum = "33480d6946193aa8033910124896ca395333cae7e2d1113d1fef6c3272217df2"
 dependencies = [
  "crossbeam-utils",
 ]
--- a/src/bin/dashboard_src/history_screen.rs
+++ b/src/bin/dashboard_src/history_screen.rs
@@ -168,7 +168,7 @@ impl HistoryScreen {
                                 None => NeptuneCoins::zero(),
                             };
                         }
-                        else { balance = balance + *amount; }
+                        else { balance += *amount; }
                         history_builder.push((*block_height, *timestamp, *amount, balance));
                     }
                     *balance_updates.lock().unwrap() = history_builder;
--- a/src/bin/dashboard_src/overview_screen.rs
+++ b/src/bin/dashboard_src/overview_screen.rs
@@ -36,6 +36,7 @@ use super::screen::Screen;
 #[derive(Debug, Clone)]
 pub struct OverviewData {
     available_balance: Option<NeptuneCoins>,
+    available_unconfirmed_balance: Option<NeptuneCoins>,
     timelocked_balance: Option<NeptuneCoins>,
     confirmations: Option<BlockHeight>,
     synchronization_percentage: Option<f64>,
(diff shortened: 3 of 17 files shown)


dan-da:
closing this as it doesn't seem likely to be merged anytime soon, and also I'm told it is infeasible to chain tx, although I don't clearly understand why.

Perhaps @aszepieniec can clarify?

aszepieniec:
Short answer: where would the proof come from?

Long answer:

Every transaction must be supported by a `SingleProof`. There are three ways to get a `SingleProof`:
 1. Start from a `PrimitiveWitness`, pass by `ProofCollection`, and then produce a `SingleProof` based on having a valid `ProofCollection`.
 2. Update a transaction with a valid `SingleProof`.
 3. Merge two transactions with valid `SingleProof`s.

Chaining transactions means that the second operand has inputs that are outputs in the first. You cannot produce a `SingleProof` (or even a `ProofCollection`) for that second transaction because you cannot prove membership of the input UTXOs in the mutator set because the mutator set has not been updated yet.

To address this deficiency, we would need a different type of transaction and supporting proof. For instance, `ChainTransaction`, where the difference is that the inputs are either
 - UTXOs that live in the mutator set, as with regular `Transaction`s
 - UTXOs that are generated by some other `ChainTransaction`
 - (for completeness sake let's include) Lock-free UTXOs (but these are irrelevant to transaction chaining).

With that data structure we could write the following proof pipelines:
 - From `PrimitiveWitness` to `ChainProofCollection` to `ChainProof`
 - Producing a `ChainProof` from updating a `ChainTransaction`
 - Producing a `ChainProof` from mergins two `ChainTransactions` (and cutting the generated-and-consumed UTXOs)
 - Producing a `SingleProof` from a `ChainTransaction` by showing that all input UTXOs live in the mutator set (or lock-free MMR).

dan-da:
thanks @aszepieniec.  That's very clear.

I will just note for anyone looking at this in the future that this PR was created before the fully fleshed out proof system was merged.   Thus tx would validate at that time which will not validate today.

PR #198, opened 2024-10-04
ignore unconfirmed spent input

### LINKED ISSUE #189: create_transaction(), send(), should consider Tx in mempool.

dan-da:
Closes #189.

Extends #184.

Alternative to #197.  (we should choose one or the other)

This PR provides a fix for the issue reported in #189.  Unlike #197, it does not implement unconfirmed tx chaining.   

from commit msg:
```
    input selection now ignores spent inputs from unconfirmed tx in the
    mempool.
    
    Also fixes an input selection bug where the balance check considers
    available funds (not timelocked) but the input utxo selection does not.
    
    Adds a test to verify that same input can no longer be spent twice.
```

### DIFF
--- a/src/models/state/mempool.rs
+++ b/src/models/state/mempool.rs
@@ -13,6 +13,7 @@ use std::collections::HashMap;
 use std::collections::HashSet;
 use std::iter::Rev;
 
+use anyhow::Result;
 use bytesize::ByteSize;
 use get_size::GetSize;
 use num_traits::Zero;
@@ -59,7 +60,13 @@ pub const TRANSACTION_NOTIFICATION_AGE_LIMIT_IN_SECS: u64 = 60 * 60 * 24;
 
 type LookupItem<'a> = (Digest, &'a Transaction);
 
-#[derive(Debug, Clone, PartialEq, Eq, GetSize)]
+#[derive(Debug, Clone)]
+pub enum MempoolEvent {
+    AddTx(Transaction),
+    RemoveTx(Transaction),
+}
+
+#[derive(Debug, GetSize)]
 pub struct Mempool {
     max_total_size: usize,
 
@@ -75,6 +82,13 @@ pub struct Mempool {
     /// Records the digest of the block that the transactions were synced to.
     /// Used to discover reorganizations.
     tip_digest: Digest,
+
+    /// a mpmc channel for interested parties to listen to mempool events
+    #[get_size(ignore)] // does not impl GetSize
+    event_channel: (
+        tokio::sync::broadcast::Sender<MempoolEvent>,
+        tokio::sync::broadcast::Receiver<MempoolEvent>,
+    ),
 }
 
 impl Mempool {
@@ -88,6 +102,7 @@ impl Mempool {
             tx_dictionary: table,
             queue,
             tip_digest,
+            event_channel: tokio::sync::broadcast::channel(100),
         }
     }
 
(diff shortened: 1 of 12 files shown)


src/models/state/mempool.rs
```diff
@@ -470,7 +529,7 @@ mod tests {
     use super::*;
 
     #[tokio::test]
-    pub async fn insert_then_get_then_remove_then_get() {
+    pub async fn insert_then_get_then_remove_then_get() -> Result<()> {
```
Sword-Smith, line 532: If the test does not return anything, then we get a nice stack trace whenever it fails. If the test returns some error, then the stack trace is not as useful, is my experience. For this reason, I prefer using `unwrap` or `expect` in tests, not the `?` operator.

src/models/state/mempool.rs
```diff
@@ -59,7 +60,24 @@ pub const TRANSACTION_NOTIFICATION_AGE_LIMIT_IN_SECS: u64 = 60 * 60 * 24;
...
+    /// a transaction was removed from the mempool
+    RemoveTx(Transaction),
+
+    /// the mutator-set of a transaction was updated in the mempool.
+    ///
+    /// (Digest of Tx before update, Tx after mutator-set updated)
+    UpdateTxMutatorSet(Digest, Transaction),
+}
+
+#[derive(Debug, GetSize)]
```
Sword-Smith, line 80: Two comments:
1. Our files are growing pretty big. Can we have a separate file for this datatype?
2. The MempoolEvent should only contain a `TransactionKernel`. It doesn't need the transaction proof for anything, as the mempool already has that (where it's used for sharing transactions with peers and merging transactions when mining). The `MempoolEvent`s received by the wallet only need the `TransactionKernel`.

src/models/state/mempool.rs
dan-da: backtraces can be automatically displayed for anyhow errors via:

cargo.toml
```
anyhow = { version = "1.0", features = ["backtrace"] }
```

.cargo/config.toml
```
[env]
RUST_BACKTRACE="1"
```

With these enabled, when a test returns an anyhow::Error a backtrace is displayed.   example:

```
---- rpc_server::rpc_server_tests::send_to_many_test stdout ----
Error: Do we get a stack trace?

Stack backtrace:
   0: anyhow::error::<impl anyhow::Error>::msg
             at /home/danda/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.81/src/backtrace.rs:27:14
   1: anyhow::__private::format_err
             at /home/danda/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.81/src/lib.rs:688:13
   2: neptune_core::rpc_server::rpc_server_tests::send_to_many_test::{{closure}}
             at ./src/rpc_server.rs:1399:9
<truncated>
```

I will add these changes to the PR. 

I choose to avoid unwrap/expect whenever possible in code I author, but I agree it is nice to have backtraces for errors in tests (and elsewhere).  Fortunately, its possible to do both.

Sword-Smith:
I just realized we have a great usecase for the `MempoolEvent::UpdateTxMutatorSet` variant: If we've performed the `update` of a transaction in our mempool, we should broadcast this transaction (as a transaction notification) to all peers since they might not have performed the `update` of the transaction proof.

Very cool! Just reveals that this is great design :)

Sword-Smith:
I'm merging this now :)

Sword-Smith (approved):

PR #214, opened 2024-10-25
RemovalRecordIntegrity: Disallow coinbase when tx has inputs

Sword-Smith:
Assert that a coinbase transaction has no inputs. This, combined with a potential future softfork requiring each block to have at least one input in its transaction, forces the miner to prove at least one execution of a `Merge` which removes any incentive to mine empty blocks. Mining empty blocks could otherwise be beneficial since it allows the miner to complete the proofs faster and get to the guessing part as soon as possible, to the detriment of the network.


aszepieniec (approved):
LGTM

PR #216, opened 2024-10-31
Set up tools and write up process for binary artifact distribution

jan-ferdinand:
The suggested pipeline is [documented in the mdbook](https://github.com/Neptune-Crypto/neptune-core/blob/dist/docs/src/contributing/releasing.md). The release process uses various tools to automate what can be automated in a meaningful way. Certain steps are best done manually. The document describes every step of the checklist with the amount of detail I think appropriate. Since I'm a bit in the weeds with this stuff, you might think a different level of detail is better suited – do let me know.

### Questions for this PR

- Are `neptune-cash-tui` and `neptune-cash-cli` good names for the respective binaries?

### Potential follow-up todos

- Add a section “Installation” to `README.md`
- Add a section “Install” to the website
- Host the mdbook “Documentation” (if it isn't already)


dan-da:
> Are neptune-cash-tui and neptune-cash-cli good names for the respective binaries?

thoughts:
1. fwiw, I asked in a meeting if binary names would change with the "neptune-cash" rebranding and was told no, they stay the same.
2. I prefer shorter names that are easier to type.  two hyphens often means an extra tab completion.  So I'd vote to keep them single-hyphen, ie neptune-cli.
3. I'm torn about neptune-tui vs neptune-dashboard.  dashboard is more descriptive.  Also longer to type.  neptune-tui is shorter to type and is nicely symmetric with neptune-cli.  "tui" is more generic and the acronym may not be understood by some/many outside rust community.  Still documentation can cover that.   Weighing all these, I prefer neptune-tui I think.

Cargo.toml
```diff
@@ -129,6 +143,23 @@ build-override.opt-level = 3
...
+[profile.dist]
+inherits = "release"
+lto = "thin"
+
+[[bin]]
+name = "neptune-core"
+path = "src/main.rs"
+
+[[bin]]
+name = "neptune-cash-tui"
```
dan-da, line 156: I think we should be consistent with the naming.

neptune-core, one hyphen.  neptune-cash-tui, two hyphens.

I vote for neptune-<core|tui|cli>

or failing that:

neptune-cash-<core|tui|cli>

dan-da (approved):
This is a quality writeup of the release process and should enable any team member to do it if necessary.   thumbs up, thx.

jan-ferdinand:
> I asked in a meeting if binary names would change with the "neptune-cash" rebranding and was told no, they stay the same.

I seem to have forgotten that you had already raised the point. Thanks. :+1:

> "tui" […] may not be understood by some/many outside rust community

The acronym “TUI” [has nothing to do with rust per-se](https://en.wikipedia.org/wiki/Text-based_user_interface), but your point certainly stands: it's quite technical.

I wonder: do we expect that the dashboard is the first tui our users ever come across? I think a certain affinity to the command line is needed to comfortably launch & navigate the dashboard. It seems reasonable that the acronym tui means something, then. Summing up, I'm in slight favor of `tui` over `dashboard` for the name of the executable.

Having read your arguments, I also think `neptune-[…]` is the better naming choice than `neptune-cash-[…]`. I can see an argument that brand name consistency is more important still. In any case, I'll wait until @aszepieniec & @Sword-Smith have weighed in before making any naming changes.

Thanks for your review!

docs/src/contributing/releasing.md
```diff
@@ -0,0 +1,156 @@
...
+cd /path/to/neptune-core
+```
+
+### Bump Version
+
+Bump the version in `Cargo.toml` [as appropriate](https://doc.rust-lang.org/cargo/reference/semver.html).
+
+### Confirm Version Bump as Semantic
+
+> ℹ️ Because binaries cannot be used as a dependency, this step is only relevant if Neptune Core has library targets.
```
aszepieniec, line 48: I think library targets is something we should support, but probably not urgently.

I'm thinking of external developers of external applications who want to do something trivial like parsing one or two types of messages or verifying a proof. They should be able to do so with minimal hassle, and I think having an importable library helps in that regard. Otherwise they must *a)* run a neptune node or *b)* copy code, is that right? However, I'm not sure what the next step would be towards making this codebase more library-like, so please comment if you know more than I do.

docs/src/contributing/releasing.md
```diff
@@ -0,0 +1,156 @@
...
+In the previous step, `cargo-release` automatically created one or multiple git tags.
+Edit them until you are happy (which probably means no editing), then push the tags to GitHub.
+
+```sh
+cargo release --execute push
+```
+
+> 🎉 Congrats on the new release!
+
+Oh… distributing binaries?
```
aszepieniec, line 152:
```suggestion
### Distributing Binaries

Oh… distributing binaries?
```

aszepieniec (approved):
Thanks for the {set,write}-up!

I'm surprised how well the rust ecosystem supports distribution of binaries. Looking ahead into the (rather further than near) future, we should consider integrating Triton VM in to this ecosystem support to enable verifiable builds.

Something that I am confused about -- and I'm not sure `releasing.md` is the right place to answer this question -- how about permissions, publishing rights, and authenticity? Specifically:
 - who has the right to publish a release (forget about internal protocol, the question is about git/github/ssh keys)?
 - how does a user verify that an installer or source code package is the real thing and not a malicious forgery?

jan-ferdinand:
> * who has the right to publish a release (forget about internal protocol, the question is about git/github/ssh keys)?

According to the [GitHub documentation](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases), “only people with write permissions to a repository can manage releases.”


> * how does a user verify that an installer or source code package is the real thing and not a malicious forgery?

The keyword on GitHub is [artifact attestation](https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds). Artifact attestation [is supported by `dist`](https://opensource.axo.dev/cargo-dist/book/supplychain-security/attestations/github.html) but I haven't enabled it yet. Verifying the integrity of these artifacts seems to require the GitHub CLI, which does not strike me as ideal. It might be prefferable to have a flawed-but-existing integrity verification process over having no such process at all.

Cargo.toml
```diff
@@ -1,10 +1,24 @@
 [package]
-authors = ["Triton Software"]
-edition = "2021"
 name = "neptune-core"
+description = "A layer-1 cryptocurrency featuring scalable privacy."
```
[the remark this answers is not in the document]
jan-ferdinand: Happy to make that change. Note that I just copied this description from [neptune.cash](https://neptune.cash/); you might want to consider changing it there, too.

docs/src/contributing/releasing.md
jan-ferdinand: A cargo package supports [up to one library target](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#library) among its potentially many binary targets. It should be possible to split the binary `neptune-core` into a library and surround it with a binary wrapper without having to publish multiple crates. I don't know the architecture well enough to be able to suggest a natural fault line; alsmost certainly, anything handling command line arguments belongs into the binary, not the library.

jan-ferdinand:
Interestingly, the GitHub merge interface has not created a merge commit, but rather has squashed the individual commits and produced one commit from that. I'm not sure this was your intention. As a side-effect, the agreed-upon maximum line width in the commit's body is violated.

PR #226, opened 2024-11-05
Asz/kill proof queue gracefully

Sword-Smith:
A template for how to kill runs of triton-vm's `prove`. Uses a spawned process and communicates via JSON/bincode. If a more lightweight solution exists, then that's probably preferable.

### DIFF
--- a/src/bin/triton-vm.rs
+++ b/src/bin/triton-vm.rs
@@ -1,0 +1,54 @@
+use std::io::BufRead;
+use std::io::Write;
+use std::io::{self};
+use std::time;
+use std::time::Duration;
+
+use neptune_core::models::proof_abstractions::tasm::consensus_program_prover_job::ConsensusProgramProverJobResult;
+use tasm_lib::triton_vm::prelude::Program;
+use tasm_lib::triton_vm::proof::Claim;
+use tasm_lib::triton_vm::prove;
+use tasm_lib::triton_vm::stark::Stark;
+use tasm_lib::triton_vm::vm::NonDeterminism;
+use tracing::info;
+
+fn main() {
+    let stdin = io::stdin();
+    let mut iterator = stdin.lock().lines();
+    let claim: Claim = serde_json::from_str(&iterator.next().unwrap().unwrap()).unwrap();
+    let program: Program = serde_json::from_str(&iterator.next().unwrap().unwrap()).unwrap();
+    let non_determinism: NonDeterminism =
+        serde_json::from_str(&iterator.next().unwrap().unwrap()).unwrap();
+    let default_stark: Stark = Stark::default();
+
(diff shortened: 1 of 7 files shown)


Sword-Smith:
There's some serialization of the proof that fails, for some reason. I think that's a benign type error, though. I think the central quality is that this allows neptune-core to stop an execution of triton-vm's `prove` function while it is running.

dan-da:
thx for the effort that went this.   my notes/thoughts:

1. I have a prototype of the job-queue that implements fifo ordering of same-priority jobs, and also removes deps on async-priority-channel crate.  I'd like to integrate that into neptune-core before anything else.  should be today.
2. tokio::runtime::Runtime::shutdown_timeout() can be used to force tokio to shutdown even if there are still spawn_blocking() tasks running.  We can implement that with a few lines in main() and I think it is a preferable option.  I can submit a PR for that today/tmw.
3. The cancellation token stuff in this PR is interesting.  I will have a look at it in more detail.  thx.
4. Running triton-vm tasks in a separate process is an interesting idea that seems to have some possible advantages and maybe introduce some new considerations.  On the plus side if the operation runs out of RAM, probably the triton-vm process would be killed first and neptune-core would likely keep running.  But new consideration: how does it handle the error?  Another consideration:  what if someone replaces the binary with some malware?  Are we introducing a new security risk?    Also does this help or hurt our build times?    I would guess the latter.
5. if we were to go forward with prover-out-of-process, I don't think the changes to JobQueue are necessary (with possible exception of the cancellation token piece).  Rather, I think the command execution and serialization should all be an implementation detail of ConsensusProgramProverJob, and JobQueue doesn't need to know or care, so it stays simple.  Anyway, that's how I would initially approach it.
6. imo the ideal solution would be that triton-vm itself exposes the prove (and run?) operations as cancellable tasks.  This could mean async, or it could mean that the methods accept a channel over which a cancellation token could be sent and the internal loop(s) watch for the cancellation msg.

dan-da:
update:

1. JobQueue now supports fifo ordering of same-priority jobs.  that's done.
2. tokio::runtime::Runtime::shutdown_timeout() is being used in main with 10 sec timeout.  done.
3. cancellation token and drop are unnecessary in JobQueue and don't fix the issue.
4. I may yet impl the ConsensusProgramProverJob::run() to exec the triton-vm process in this PR.  Still debating though, as that introduces a whole bunch of new failure modes.
5. see (4)
6. @Sword-Smith any chance that triton-vm could offer a cancellation token/channel for cancelling prove? 

keeping this open until (4) and (6) are resolved.

Sword-Smith:
Superseeded by @dan-da's  work on this. Please close this if you agree, @dan-da.

PR #227, opened 2024-11-06
ArchivalMutatorSet: Fix get_mutator_set_update_to_tip

### LINKED ISSUE #225: Bug in `get_mutator_set_update_to_tip`
This function is supposed to return a `MutatorSetUpdate` to go from the indicated `MutatorSetAccumulator` (`A`) to the current one in tip, but the `RemovalRecord` list that is returned is not correct, as it refers to each interleaving block's removal records, and not to the removal records as they look in `A` which is what we need.

We either need to roll these removal records back somehow, or we need to fetch them from an archival mutator set.

Sword-Smith:
Fix a problem in get_mutator_set_update_to_tip where the wrong removal records were returned.

The function attempts to find a `MutatorSetUpdate` that transforms the mutator set from a state B to a later state A where B and A can be separated by an arbitrary number of blocks. The problem was the the removal records were collecting from all blocks between B and A. But the MutatorSetUpdate data structure needs the removal records valid for block B.

To fix this, we make ephemeral changes to the archival mutator set to assemble the required information. And then roll those ephemeral changes back once the information is gathered.

This closes #225.

A solution where a "snapshot" of the archival mutator set was returned -- an archival mutator set that could read from the databases but only write to cache, not databases -- would be a more elegant and robust solution, I think. But that seems like a rather big rewrite. 

Since a write-lock is held over the `ArchivalMutatorSet` throughout this operation, I believe that the operation is safe and should not lead to data corruption.

Prior to this commit, the test `models::state::archival_state::archival_state_tests::update_mutator_set_rollback_many_blocks_multiple_inputs_outputs_test` was flaky. With this commit, that is no longer the case.

### DIFF
--- a/src/models/state/archival_state.rs
+++ b/src/models/state/archival_state.rs
@@ -13,6 +13,7 @@ use tracing::warn;
 use twenty_first::math::digest::Digest;
 
 use super::shared::new_block_file_is_needed;
+use super::StorageVecBase;
 use crate::config_models::data_directory::DataDirectory;
 use crate::config_models::network::Network;
 use crate::database::create_db_if_missing;
@@ -30,6 +31,7 @@ use crate::models::database::BlockRecord;
 use crate::models::database::FileRecord;
 use crate::models::database::LastFileRecord;
 use crate::prelude::twenty_first;
+use crate::util_types::mutator_set::addition_record::AdditionRecord;
 use crate::util_types::mutator_set::mutator_set_accumulator::MutatorSetAccumulator;
 use crate::util_types::mutator_set::removal_record::RemovalRecord;
 use crate::util_types::mutator_set::rusty_archival_mutator_set::RustyArchivalMutatorSet;
@@ -725,13 +727,12 @@ impl ArchivalState {
     /// max search depth, as it loads all the blocks in the search path into
     /// memory. A max search depth of 0 means that only the tip is checked.
     pub(crate) async fn get_mutator_set_update_to_tip(
-        &self,
+        &mut self,
         mutator_set: &MutatorSetAccumulator,
         max_search_depth: usize,
     ) -> Option<MutatorSetUpdate> {
         let mut search_depth = 0;
-        let mut addition_records = vec![];
-        let mut removal_records = vec![];
+        let mut block_mutations = vec![];
 
         let mut haystack = self.get_tip().await;
         let mut parent = self.get_tip_parent().await;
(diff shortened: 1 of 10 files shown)


src/models/state/archival_state.rs
```diff
@@ -718,13 +720,12 @@ impl ArchivalState {
     /// max search depth, as it loads all the blocks in the search path into
     /// memory. A max search depth of 0 means that only the tip is checked.
     pub(crate) async fn get_mutator_set_update_to_tip(
```
dan-da, line 722: 1. perhaps consider changing the name of this fn.   It seems odd to have a fn prefixed with `get_` that mutates self.

2. just a note regarding the change to &mut self.  any time we perform a lot of db read operations inside a write-lock, it is potentially bad for concurrency as all readers must wait including rpc calls.  Hopefully we can eventually refactor such methods into separate read and write methods where the read method only acquires read lock for gathering updates and the write-lock is held as briefly as possible to write out the updates.  The tricky part though is atomicity over the entire read+write.  (not proposing a change here at this time)

dan-da (changes requested):
lgtm overall. 

I requested one change to remove clear_cache() from StorageVecBase trait, where I don't think it belongs.  I didn't review the actual mutator-set mods in great detail as I'm not clear on how all that works anyway, so I trust your judgement on that.

I worry any time we do a lot of work inside an &mut self method, for concurrency reasons.  at least we have the log-slow-write-lock feature now we can check for any serious problems.

src/database/storage/storage_vec/traits.rs
```diff
@@ -62,6 +62,9 @@ pub trait StorageVecBase<T: Send> {
     ///       never an intermediate state.
     async fn set_many(&mut self, key_vals: impl IntoIterator<Item = (Index, T)> + Send);
 
+    /// Clear non-persisted values
```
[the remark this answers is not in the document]
Sword-Smith: Avoiding this method on the trait, means I have to add this `impl` block in `archival_mmr`, which I'm OK with.

```
impl ArchivalMmr<DbtVec<Digest>> {
    /// Delete ephemeral (cache) values, without persisting them.
    ///
    /// Can be used to roll-back ephemeral values to a persisted state.
    pub(crate) async fn delete_cache(&mut self) {
        self.digests.delete_cache().await;
    }
}
```

src/models/state/archival_state.rs
Sword-Smith: It takes a `&mut self` but does not actually mutate it -- as in it's the same mutator set before and after the call. It does persist any cached values but from a mathematical perspective, it's the same mutator set.

PR #228, opened 2024-11-06
job_queue: Return joinhandles to caller for graceful shutdown

Sword-Smith:
Competing implementation to shut down `JobQueue`. This PR would replace functionality introduced in 4a36ec54d6607dbcdce98ec040c737edf86c8442. 

I personally prefer this PR to 4a36ec54d6607dbcdce98ec040c737edf86c8442 since this PR only uses Rust's ownership concept to ensure that only one code path can shut down the job queue. Also: It follows the existing pattern of accumulating `JoinHandle`s in a list and looping over that list in `main_loop`'s graceful shutdown.

### DIFF
--- a/src/job_queue/queue.rs
+++ b/src/job_queue/queue.rs
@@ -1,8 +1,10 @@
 use std::collections::VecDeque;
 use std::sync::Arc;
 use std::sync::Mutex;
+
 use tokio::sync::mpsc;
 use tokio::sync::oneshot;
+use tokio::task::JoinHandle;
 
 use super::traits::Job;
 use super::traits::JobResult;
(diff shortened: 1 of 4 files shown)


dan-da:
Thx @Sword-Smith  for the time and effort spent looking into this issue and coming up with this PR.

You are already aware, but I will summarize current situation here for the record.

----

Unfortunately this doesn't actually fix the problem.    No fix exists yet because the problem isn't entirely understood.  It might be a tokio issue, or it might even be due to CPU starvation caused by proving, which delays the mining task from shutting down for longer than it otherwise would.

The issue can be reproduced by starting a single neptune-core with --mine, wait unti it starts generating a single proof, then hit ctrl-c.

I reproduced the problem with the branch in this PR as well as the previous "fix" in 4a36ec54d6607dbcdce98ec040c737edf86c8442 on master.

The code in this branch and previous PR makes the assumption that the sub-tasks in JobQueue need to be attached by storing the JoinHandle, and then the joinhandle can be used to abort the tasks.    However, the problem is not that the tasks aren't stopping.   Rather it is that they stop too soon during the tokio runtime shutdown, before the JobQueue itself has been dropped.  This occurs whether or not JobQueue owns the JoinHandles.

I've made a hack/fix in 5b721a31e6934e450dddceaa15022b7a282cd231.   I also started a [discussion](https://github.com/tokio-rs/tokio/discussions/6961) on tokio github with a [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d888dff82b5940950e1ae2b862592249).

Lastly, I want to keep JobQueue as much of a black-box as possible, so I prefer not to expose internal impl details such as JoinHandles. 

closing this.

PR #231, opened 2024-11-11
Improve mempool tx update handling

Sword-Smith:
When we receive a new block, we might have to generate a Triton VM proof for the `update` program which can take minutes. This task was previously done during `set_new_tip`, specifically  in `Mempool::update_with_block_and_predecessor`. With this commit, we instead return the update jobs that need to be run and then spawn a task in which this expensive computation is performed, after the write-lock on `GlobalState` has been released.

There's one warning left in there intentionally, as I wasn't quite sure how to handle the returned `Vec<MempoolEvent>`. Maybe you, apart from the general architecture, @dan-da, can also see if we're using `Vec<MempoolEvent>` correctly here?

### DIFF
--- a/src/models/state/mempool.rs
+++ b/src/models/state/mempool.rs
@@ -42,17 +42,14 @@ use priority_queue::DoublePriorityQueue;
 use serde::Deserialize;
 use serde::Serialize;
 use tasm_lib::triton_vm::proof::Proof;
-use tokio::task::JoinHandle;
 use tracing::debug;
 use tracing::error;
-use tracing::info;
 use tracing::warn;
 use twenty_first::math::digest::Digest;
 
 use super::transaction_kernel_id::TransactionKernelId;
 use super::tx_proving_capability::TxProvingCapability;
-use crate::job_queue::triton_vm::TritonVmJobPriority;
-use crate::job_queue::triton_vm::TritonVmJobQueue;
+use crate::main_loop::proof_upgrader::UpdateMutatorSetDataJob;
 use crate::models::blockchain::block::Block;
 use crate::models::blockchain::transaction::transaction_kernel::TransactionKernel;
 use crate::models::blockchain::transaction::validity::proof_collection::ProofCollection;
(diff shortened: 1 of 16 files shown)


src/models/state/mempool.rs
```diff
@@ -303,12 +314,28 @@ impl Mempool {
...
+                    .iter()
+                    .any(|x| !matches!(&x.1.proof, TransactionProof::SingleProof(_))),
+            }
+        }
+
         let mut events = vec![];
 
-        match transaction.proof {
+        match new_tx.proof {
             TransactionProof::Invalid => panic!("cannot insert invalid transaction into mempool"),
```
dan-da, line 339: same nit as above.  If Mempool only accepts a type that doesn't include Invalid variant, then caller has to deal with the issue somehow, and its not mempool's problem.  Probably it should be dealt with when its first determined the proof is invalid.

src/models/state/mod.rs
```diff
@@ -1472,19 +1460,21 @@ impl GlobalState {
...
                     &new_block,
                     &tip_parent,
-                    vm_job_queue,
-                    TritonVmJobPriority::Highest,
+                    myself.net.tx_proving_capability,
                     myself.cli().compose,
                 )
                 .await;
 
+            // TODO: Inform wallet about mempool_events
```
dan-da, line 1476: If the idea is that we should not inform wallet until the proof and mutatorsetupdate has finished (which seems logical), then it seems that these methods should also return Ok((mempool_events, update_jobs)) and the caller can inform the wallet after the update jobs have completed successfully.

Of course that is gross as it separates the logic/location of where "inform wallet" actions take place into different modules.

A bigger problem with it is that the user will not see the spend reflected in their wallet balance or history until the jobs finish.  Which can be very disconcerting.  That is the case now of course, but we may have an opportunity to improve it, and should.

perhaps GlobalState::update_with_block_and_predecessor can inform the wallet of the mempool events as before, but the wallet could somehow detect / inform that these tx are incomplete/partial.

Failing something like that, we need to somehow inform the user that the send succeeded, but is expected to take minutes before it will be reflected in wallet balance/history.  I suppose that could just be client UI messaging.

something to think about.

dan-da (changes requested):
I looked the changes over but did not try building/running this.

I see there is a CI clippy error (maybe a compiler error?) about mempool_events variable not being used.  So that needs to be fixed.

I left a comment about possibilities for dealing with the mempool events.  The biggest issue I see is getting immediate feedback to the user that the send succeeded, or at least setting expecations so user is not left wondering if the send happened or not while the jobs are running.

architecturally this seems fine for now.  It would be possible, but very messy, to return the ConsensusProgramProverJob JobHandle all the way up the stack from prove_consensus_program().  I think the way you do it with the UpdateMutatorSetDataJob is cleaner, especially as there is additional work that needs doing around the prove(s).

At some point it may become worthwhile to refactor the jobs in main_loop into JobQueue jobs.   So they can be easily cancelled, consistent logging, consistent interface, etc.  Though we may not want everything serialized, so that's a consideration.   And of course any such job has to be very careful not to hold the write-lock for long.

src/models/state/mempool.rs
```diff
@@ -303,12 +314,28 @@ impl Mempool {
...
-        transaction: Transaction,
+        new_tx: Transaction,
         origin: TransactionOrigin,
     ) -> Vec<MempoolEvent> {
+        fn new_tx_has_higher_proof_quality(
+            new_tx: &Transaction,
+            conflicts: &[(TransactionKernelId, &Transaction)],
+        ) -> bool {
+            match &new_tx.proof {
+                TransactionProof::Invalid => panic!("Invalid proofs don't belong in mempool"),
```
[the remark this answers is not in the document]
Sword-Smith: Seems like we can entirely get rid of the `Invalid` variant of `TransactionProof`. Thanks for pointing in that direction.

Sword-Smith:
I deleted the `TransactionProof::Invalid` variant. I'll ruminate a bit on what to do with the mempool events. Other than that, I think this PR is basically ready to be merged. This branch is the first one where I managed to mine a block with non-zero inputs in the transaction kernel, so it'll be helpful for all to get this on `master` ASAP.

PR #243, opened 2024-11-15
Offchain notifications

Sword-Smith:
Implementation is very close to that of #185. But the derivation of the mutator set membership proofs is significantly simpler here.

Had to replace a `stream_many` with a `get_many` in `ArchivalMutatorSet` because of a rustc type limitation. I prefer this approach though, as the sequence in question is only 45 elements long.

### DIFF
--- a/scripts/linux/testnet-premine-receiver-no-mine-from-genesis-faked-time.sh
+++ b/scripts/linux/testnet-premine-receiver-no-mine-from-genesis-faked-time.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+#
+# Run one instance without mining, with a faked time such that the premine UTXO
+# can be spent immediately.
+
+set -e # Exit on first error.
+
+export RUST_LOG=debug;
+
+# Build before proceeding
+cargo build
+
+NC_LOCAL_STATE_DIR="$HOME/.local/share/neptune-integration-test-from-genesis"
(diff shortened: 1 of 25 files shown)


dan-da:
I see the CI runs did not complete, due to some tests taking too long.  Maybe uncached proofs.

dan-da:
I did some manual testing and was able to successfully claim a utxo and see it reflected in my balance.   I tried both claiming scenarios:
  1. claim before tx is mined in a block
  2. claim after tx mined in a block.

here's the commands I used, adapted from procedure in #185.

```
./scripts/linux/run-multiple-instances-from-genesis-log-locks.sh

echo './target/debug/neptune-cli --server-addr 127.0.0.1:19790 $1 $2 $3 $4 $5 $6 $7 $8' > neptune-cli-alice.sh

echo './target/debug/neptune-cli --server-addr 127.0.0.1:19791 $1 $2 $3 $4 $5 $6 $7 $8' > neptune-cli-alice.sh

./neptune-cli-alice.sh own-receiving-address > bob.address

./neptune-cli-alice.sh send `cat bob.address` 5 0.1 bob --on-chain off-chain
<file>

./neptune-cli-bob.sh claim-utxo file <file>
```

I think the PR (or a followup) should include a brief description of the feature and basic usage example/instructions as per above in the readme or a separate docs file.

Issue #244, opened 2024-11-16
dashboard_overview_data() is often quite slow.


dan-da:
I have instrumented dashboard_overview_data() to log duration of most of the functions it calls.

Below are warnings from a recent invocation that lasted 4.3 seconds.

setup:
 * node is composing at time of RPC call.
 * block-height is 0.
 * wallet and mempool are empty.

```
2024-11-16T02:42:45.616949478Z  WARN ThreadId(02) RPC{rpc.trace_id=00 rpc.deadline=2024-11-16T02:42:55.604790658Z otel.kind="server" otel.name="RPC.d
ashboard_overview_data"}: neptune_core: executed dashboard_overview_data()::cpu_temp_inner() in 0.007878622 secs.  exceeds slow fn threshold of 0.001 secs.  location: /home/danda/neptune-core/src/rpc_server.rs:826:13

2024-11-16T02:42:47.129279367Z  WARN ThreadId(02) RPC{rpc.trace_id=00 rpc.deadline=2024-11-16T02:42:55.604790658Z otel.kind="server" otel.name="RPC.dashboard_overview_data"}: neptune_core: executed dashboard_overview_data()::unconfirmed_balance() in 1.512249878 secs.  exceeds slow fn threshold of 0.001 secs.  location: /home/danda/neptune-core/src/rpc_server.rs:830:13

2024-11-16T02:42:48.548855423Z  WARN ThreadId(02) RPC{rpc.trace_id=00 rpc.deadline=2024-11-16T02:42:55.604790658Z otel.kind="server" otel.name="RPC.dashboard_overview_data"}: neptune_core: executed dashboard_overview_data()::synced_unspent_available_amount() in 1.419075826 secs.  exceeds slow fn threshold of 0.001 secs.  location: /home/danda/neptune-core/src/rpc_server.rs:850:13

2024-11-16T02:42:49.96716567Z  WARN ThreadId(02) RPC{rpc.trace_id=00 rpc.deadline=2024-11-16T02:42:55.604790658Z otel.kind="server" otel.name="RPC.dashboard_overview_data"}: neptune_core: executed dashboard_overview_data()::synced_unspent_timelocked_amount() in 1.418132393 secs.  exceeds slow fn threshold of 0.001 secs.  location: /home/danda/neptune-core/src/rpc_server.rs:854:13

[...]

PR #245, opened 2024-11-16
ConsensusProgram: cache all used program hashes

Sword-Smith:
The program hashes stay constant for all the `ConsensusProgram` implemenations that we are using, as the implementing structures don't take any arguments. So for each of those implementations we only need to calculate the program hash once. This commit introduces caching to that effect.

Opening this as a PR since it contains a bit of repeated code and @dan-da might have a better approach, or opinions on style.

This addresses #244 and improves the situation considerably, I think. But I don't think it closes it, as I still see some warnings of exceeded fn thresholds in the log when I ran the client with this commit.

### DIFF
--- a/src/models/blockchain/block/validity/block_program.rs
+++ b/src/models/blockchain/block/validity/block_program.rs
@@ -42,6 +42,12 @@ impl BlockProgram {
 }
 
 impl ConsensusProgram for BlockProgram {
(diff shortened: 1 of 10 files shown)


Sword-Smith:
Replaced by 1b81397d7f103e362aca93ff45383edbd7c0eee4

PR #254, opened 2024-11-21
Block: Add test that double-spending blocks are rejected

Sword-Smith:
@aszepieniec Let me know if these small rewrites conflict with your current work.

TODO (for future commits, not for this PR): Block's `is_valid` function should return an error code indicating why the block was rejected. That would make negative tests of block validity much more robust. I'll add an issue to that effect.

Includes a rewrite of helper function for constructing a new block that makes it easier to create a new block from a BlockPrimitiveWitness. Specifically, a new function `block_template_from_primitive_witness` is added which is used in this test as well as in the call graph for creating a new block from a transaction and a predecessor.

Also adds an "unsafe" version of MutatorSetUpdate's "apply" function that allows the caller to calculate a new MS accumulator which ignores double spends. This was needed to allow the BlockPrimitiveWitness constructor to complete the construction of a block with a double- spending transaction.

### DIFF
--- a/src/models/blockchain/block/mod.rs
+++ b/src/models/blockchain/block/mod.rs
@@ -169,26 +169,26 @@ impl Eq for Block {}
 
 impl Block {
     fn template_header(
-        predecessor: &Block,
+        predecessor_header: &BlockHeader,
+        predecessor_digest: Digest,
         timestamp: Timestamp,
         nonce: Digest,
         target_block_interval: Option<Timestamp>,
     ) -> BlockHeader {
         let difficulty = difficulty_control(
             timestamp,
-            predecessor.header().timestamp,
-            predecessor.header().difficulty,
+            predecessor_header.timestamp,
+            predecessor_header.difficulty,
             target_block_interval,
-            predecessor.header().height,
+            predecessor_header.height,
         );
 
         let new_cumulative_proof_of_work: ProofOfWork =
-            predecessor.kernel.header.cumulative_proof_of_work
-                + predecessor.kernel.header.difficulty;
+            predecessor_header.cumulative_proof_of_work + predecessor_header.difficulty;
         BlockHeader {
             version: BLOCK_HEADER_VERSION,
-            height: predecessor.kernel.header.height.next(),
-            prev_block_digest: predecessor.hash(),
+            height: predecessor_header.height.next(),
+            prev_block_digest: predecessor_digest,
             timestamp,
             nonce,
             cumulative_proof_of_work: new_cumulative_proof_of_work,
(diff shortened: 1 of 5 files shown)


src/models/blockchain/block/mutator_set_update.rs
```diff
@@ -50,6 +57,15 @@ impl MutatorSetUpdate {
...
+    ) -> Result<()> {
+        self.apply_to_accumulator_and_records_inner(ms_accumulator, removal_records, true)
+    }
+
+    fn apply_to_accumulator_and_records_inner(
+        &self,
+        ms_accumulator: &mut MutatorSetAccumulator,
+        removal_records: &mut [&mut RemovalRecord],
+        check_for_double_spend: bool,
     ) -> Result<()> {
```
aszepieniec, line 69: This function interface strikes me as an unsafe potential pitfall.

Suggestion for improvement: The inner function should return a bool that indicates whether removing the items was successful. In the unsafe wrapper, you ignore this boolean. In the safe public wrapper you bail if it is true and return `Ok(())` otherwise.

src/models/blockchain/block/validity/block_program.rs
```diff
@@ -179,11 +179,15 @@ pub(crate) mod test {
     use itertools::Itertools;
     use tasm_lib::triton_vm::vm::PublicInput;
     use tracing_test::traced_test;
+    use triton_vm::prelude::Digest;
 
     use super::*;
-    use crate::job_queue::triton_vm::TritonVmJobQueue;
+    use crate::job_queue::triton_vm::{TritonVmJobPriority, TritonVmJobQueue};
```
aszepieniec, line 185: "item"

src/models/blockchain/block/validity/block_program.rs
```diff
@@ -179,11 +179,15 @@ pub(crate) mod test {
...
     use crate::models::blockchain::block::validity::block_primitive_witness::test::deterministic_block_primitive_witness;
+    use crate::models::blockchain::block::{Block, BlockPrimitiveWitness, TritonVmProofJobOptions};
```
aszepieniec, line 187: granularity

src/models/blockchain/block/mod.rs
```diff
@@ -220,32 +219,17 @@ impl Block {
         Block::new(header, body, appendix, proof)
     }
 
-    async fn make_block_template_with_valid_proof(
-        predecessor: &Block,
-        transaction: Transaction,
-        block_timestamp: Timestamp,
+    pub(crate) async fn block_template_from_primitive_witness(
```
aszepieniec, line 222:
```suggestion
    pub(crate) async fn block_template_from_block_primitive_witness(
```

aszepieniec (approved):
To answer your question: no, I do not think it is in conflict with my branch. Please go ahead and merge. But before you do, please consider my suggestions.

src/models/blockchain/block/validity/block_program.rs
Sword-Smith: Thanks for noticing. Apparently I haven't set up the right formatting rules on my desktop at home. Will fix.

Sword-Smith:
Merged in 4746a778eee2a2dd17f236d62f89aba00d195dad

PR #261, opened 2024-11-23
test: List all consensus-critical snippets

### LINKED ISSUE #252: Write a list of snippets used for consensus
Compose a list of all Triton VM assembler snippets used in the consensus programs. This list is the basis of the internal security review we'll do on our Triton VM programs.

jan-ferdinand:
1. Introduce helper function `ConsensusProgram::library_and_code()` to easily access a consensus program's imported snippets.
2. Use associated functions over methods for trait `ConsensusProgram` and its implementors.

I am not sure whether the second point is the right way to go. Several consensus programs are not implemented yet, and their respective structs are not unit structs. If those consensus programs require information from those fields, associated functions are a no-go.

Fix #252.


Sword-Smith:
I would prefer to keep the flexibility that methods give us. It would, for example, allow us to mock proof validation inside a program by adding a boolean to the ConsensusProgram struct whether to do STARK proof verification or not.

Sword-Smith (changes requested):
Let's keep the trait functions as methods, and not switch to associated functions. Also: LGTM as soon as that's done.

Sword-Smith (approved):

PR #267, opened 2024-11-26
fix(mine_loop): Avoid using different blocks in block composition

### LINKED ISSUE #266: Race condition in block witness composition
Node crashed, apparently when its miner found a block simultaneously with receiving a block of the same height from a peer. For some reason, this lead to an inconsistent `BlockPrimitiveWitness` being made. Potentially, it would suffice to just replace an `assert` in `BlockPrimitiveWitness` with a `anyhow` error.

[Offending `assert`](https://github.com/Neptune-Crypto/neptune-core/blob/master/src/models/blockchain/block/validity/block_primitive_witness.rs#L91) :
```rust
       let predecessor_msa_digest = self.predecessor_block
            .mutator_set_accumulator_after()
            .hash();
            let tx_msa_digest = self.transaction.kernel.mutator_set_hash;
            assert_eq!(
                predecessor_msa_digest,
                tx_msa_digest,
                "Mutator set of transaction must agree with mutator set after previous block.\
                \nPredecessor block had {predecessor_msa_digest};\ntransaction had {tx_msa_digest}\n\n"
            );
```

I'm pretty sure the problem is that the functions `make_coinbase_transaction` reads the latest block from light state whereas `Block::compose(` uses the `latest_block` value given as argument to `compose_block` in `mine_loop.rs`. The block witness generation process must agree on which block to use. And it should probably always use the block that is currently considered canonical.

Sword-Smith:
Prior to this fix, a race condition could lead to two different blocks being used in the witness composition for a new block. This happened since one block was input to the function, and inside the function to construct the coinbase transaction, a block was read from global state. With this fix, the block used as input argument is always used.

This closes #266.

### DIFF
--- a/src/mine_loop.rs
+++ b/src/mine_loop.rs
@@ -292,6 +292,7 @@ fn guess_nonce_iteration(
 }
 
 pub(crate) async fn make_coinbase_transaction(
+    latest_block: &Block,
     global_state_lock: &GlobalStateLock,
     guesser_block_subsidy_fraction: f64,
     timestamp: Timestamp,
@@ -304,6 +305,11 @@ pub(crate) async fn make_coinbase_transaction(
     //     we actually win the mining lottery.
     //  3. also this way we do not have to modify global/wallet state.
 
+    // It's important to use the input `latest_block` here instead of
+    // reading it from state, since that could, because of a race condition
+    // lead to an inconsistent witness higher up in the call graph. This is
+    // done to avoid holding a read-lock throughout this function.
+
     let coinbase_recipient_spending_key = global_state_lock
         .lock_guard()
         .await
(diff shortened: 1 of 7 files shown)


dan-da:
Just adding some context/observation.   As I understand it the situation is something like:

1. miner starts generating coinbase Tx for block 55
2. receives block 56 from peer.
3. finishes proof for 55.
4. hit assert, 55 mutator set != 56 mutator set.  fixed by this PR.
5. finishes coinbase Tx for 55.

and it would be bad if:

6. starts mining on 55?

Instead of mining on 55, the miner needs to be made aware of 56, abort work on 55, and start work on 56.

This is in fact happening in the select!() inside mine_loop::mine().  When a MainToMiner::NewBlock message is received, the composer_task is aborted.  

note: I believe the composer_task.abort() to be redundant, as select!() drops any other uncompleted branch anyway.  But it doesn't seem to be hurting anything.

So that's all fine.

This PR keeps the compose task working on 55 in a happy consistent state operating on 55 data instead of switching to 56 data, which was causing the assert.    In fact this task will soon be aborted anyway, but we don't want it asserting in the meantime.

It seems the panic was happening after globalstate had been modified, but before mining loop had received the message from main.

dan-da (approved):
lgtm!   good diagnosis/fix.

PR #268, opened 2024-11-27
Bootstrapper Mode

aszepieniec:
We are observing that people would like to participate in the network but can't because the two nodes listed in the README.md refuse connections. They do this because their max number of peers has already been reached.

This PR adds functionality for running a node in "bootstrapper mode", wherein it will drop the longest-lived peer connection whenever the maximum is reached. This mode is activated through a CLI argument. Whenever an incoming connection is established, if this mode is set, and if the max number of peers has been reached, the peer loop will send a  `DisconnectFromLongestLivedPeer` message to the main loop. The main loop iterates over all peers, filters out the peers given as CLI arguments, selects the one with the earliest connection established date, and instructs that peer's peer loop to close the connection.

### DIFF
--- a/src/connect_to_peers.rs
+++ b/src/connect_to_peers.rs
@@ -134,10 +134,20 @@ async fn check_if_connection_is_allowed(
         return ConnectionStatus::Refused(ConnectionRefusedReason::IncompatibleVersion);
     }
 
+    // If this connection touches the maximum number of peer connections, say
+    // so with special OK code.
+    if cli_arguments.max_num_peers as usize == global_state.net.peer_map.len() + 1 {
+        info!("ConnectionStatus::Accepted, but max # connections is now reached");
+        return ConnectionStatus::AcceptedMaxReached;
+    }
+
     info!("ConnectionStatus::Accepted");
     ConnectionStatus::Accepted
 }
 
+/// Respond to an incoming connection initiation.
+///
+/// Catch and process errors (if any) gracefully.
 pub(crate) async fn answer_peer_wrapper<S>(
     stream: S,
     state_lock: GlobalStateLock,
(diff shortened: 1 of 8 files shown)


src/connect_to_peers.rs
```diff
@@ -134,10 +134,20 @@ async fn check_if_connection_is_allowed(
         return ConnectionStatus::Refused(ConnectionRefusedReason::IncompatibleVersion);
     }
 
+    // If this connection touches the maximum number of peer connections, say
+    // so with special OK code.
+    if cli_arguments.max_num_peers as usize == global_state.net.peer_map.len() + 1 {
```
Sword-Smith, line 139: Shouldn't you also check that the node is running as a bootstrapper node here? It seems to me that all nodes will display this "disconnect" behavior.

src/main_loop.rs
```diff
@@ -811,13 +811,40 @@ impl MainLoopHandler {
 
                 self.main_to_miner_tx.send(MainToMiner::NewBlockProposal);
             }
+            PeerTaskToMain::DisconnectFromLongestLivedPeer => {
+                let global_state = self.global_state_lock.lock_guard().await;
+
+                // get all peers
+                let all_peers = global_state.net.peer_map.iter();
+
+                // filter out CLI peers
```
Sword-Smith, line 820: Excellent that you're allowing CLI arguments to preserve peers that will never be disconnected.

Sword-Smith (changes requested):
I think you're missing a check on the `bootstrap` value in a branch. I think you're not using the `bootstrap` field at all. Otherwise, the logic looks excellent.

Sword-Smith (approved):

PR #273, opened 2024-11-29
Add transaction history to recovery of `MonitoredUtxo`

Sword-Smith:
Previously, monitored UTXOs were recovered without information about in which block they were included. This fixes that.

Note that this changes the structure of the pivotal `BlockIndex` database, so all nodes who want this upgrade have to delete these directories: `<datadir>/databases/` and `<datadir>/blocks/`. On a Linux machine, that's usually `~/.local/share/neptune/beta/blocks/` and `~/.local/share/neptune/beta/databases/`, assuming they're on betanet.

I tested this against a composing node with a non-zero balance, and it works for both announced UTXOs and for coinbase-mined UTXOs.

### DIFF
--- a/src/models/blockchain/block/block_height.rs
+++ b/src/models/blockchain/block/block_height.rs
@@ -5,6 +5,7 @@ use std::ops::Sub;
 
 use arbitrary::Arbitrary;
 use get_size::GetSize;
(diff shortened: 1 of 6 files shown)


Sword-Smith:
Merged through #274

PR #279, opened 2024-12-06
perf: serialize only 'seed' field of GenerationSpendingKey

dan-da:
We serde(skip) fields that can be computed from the seed and then manually impl Deserialize to populate those fields.  This makes the serialized representation smaller including bech32m.  SpendingKey are not transmitted/stored much, so this is only a minor improvement.

I was hoping to do the same thing for GenerationReceivingAddress in order to reduce its (huge) bech32m representation, however that does not appear to be possible because all of its fields are derived from the original (secret) seed and none are derived from the public key.    It seems we could possibly make recipient_id derive from the pubkey, and then the spending-key could also derive it from pubkey so they match.  So that would be a minor reduction in size.  Perhaps with some study we could figure out a more compact representation...


aszepieniec (commented):
LGTM

PR #280, opened 2024-12-09
feat: Time-Locked Coinbase

aszepieniec:
In Neptune Cash, the coinbase is divided into two parts. The first is liquid immediately; the second is locked for 3 years. This PR applies the latter lock.

Note that the guesser reward is already split in two parts, one liquid immediately and the other locked for 3 years. This commit modifies the `NativeCurrency` type script so as to require that if the transaction has a coinbase, then half of it is locked -- however this amount is distributed across time-locked outputs and guesser fee.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3373,14 +3373,13 @@ dependencies = [
 [[package]]
 name = "tasm-lib"
 version = "0.43.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e2086093477a23c66dc1cfcbb3b3cd6eb9cbc96ab0b6d0b9f796ea14942760e"
+source = "git+https://github.com/TritonVM/tasm-lib.git?rev=bc50c3320b7ca7527a0066e2b28455f4b4e18003#bc50c3320b7ca7527a0066e2b28455f4b4e18003"
 dependencies = [
  "anyhow",
  "arbitrary",
  "const_format",
  "hex",
- "itertools 0.13.0",
+ "itertools 0.10.5",
  "ndarray",
  "num",
  "num-traits",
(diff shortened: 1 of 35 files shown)


dan-da:
Can you describe (or link to a description) the purpose (game theory? security? economics? marketing?) of time-locking 50% for 3 years, and why those specific pararameters are chosen.   thx.

aszepieniec:
The dynamic we are hoping to engineer is to align the incentives of miners with the long-term success of the cryptocurrency. Miners have an interest in the price now as well as in the price three years from now, which is affected by the quality of the protocol at that time. In particular:

 - It turns miners into cheerleaders for Neptune Cash in a way that alternative blockchains don't an analogue for.
 - In a proof-of-work blockchain, miners have some deciding power over proposed protocol changes, and in this context we want their incentives aligned with the long-term success of the blockchain.
 - We want to incentivize miners to support continued protocol development and maintenance.

As for the particular parameters:

 - Three years coincides with the halving period. This makes it easier to draw circulating supply graphs, for those that are interested in that sort of thing.
 - Half coincides with a right shift by one bit, which is the simplest possible non-trivial arithmetic for dividing amounts. Bear in mind that whatever alternative you might propose would need to be implemented robustly in Triton assembler, because otherwise you lose one of the main selling points of Neptune, which is the cheap verification of consensus.

We did consider alternative parameters. We had the following considerations:

 - Short term and early term price play should not affect architectural decisions.
 - Complex policies are a hurdle for spreading understanding. So simplicity and similarity to existing templates are good.
 - Bitcoin launched into a world where no-one knew about the potential of blockchain money; by contrast, the environment Neptune Cash is launching into is one where everyone knows about the potential of blockchain money. As a result, it is reasonable to expect faster adoption rates and as a consequence of that we can afford a faster approach to the steady state where transaction fees dominate over block subsidies without risking a lopsided distribution of money. This line of thought motivates having a 3-year rather than 4-year halving period.

aszepieniec:
This branch has been extended (asz/negative-fees) and then merged into master.

PR #296, opened 2024-12-19
test(SingleProof): verify set `merge_bit` in PC->SP path disallowed

Sword-Smith:
Verify that the merge_bit must be set to false in the `ProofCollection` -> `SingleProof` branch in `SingleProof`.

Required generalization of a test-case generator which I think will come in handy later.

This PR was opened because of the change to the test-case generators which are @aszepieniec's  brain child. He should have a say here.

### DIFF
--- a/src/models/blockchain/transaction/primitive_witness.rs
+++ b/src/models/blockchain/transaction/primitive_witness.rs
@@ -567,6 +567,7 @@ impl PrimitiveWitness {
         num_inputs: Option<usize>,
         num_outputs: usize,
         num_public_announcements: usize,
+        merge_bit: bool,
     ) -> BoxedStrategy<Self> {
         // Primitive witnesses may not simultaneously have inputs and set a
         // coinbase. In combination with a rule in `Block::is_valid` that
@@ -685,6 +686,7 @@ impl PrimitiveWitness {
                         fee,
                         maybe_coinbase,
                         timestamp,
+                        merge_bit,
                     )
                 },
             )
@@ -704,6 +706,7 @@ impl PrimitiveWitness {
         let output_utxos = output_utxos.to_vec();
         let public_announcements = public_announcements.to_vec();
 
+        let merge_bit = false;
         arb::<Timestamp>()
             .prop_flat_map(move |now| {
                 Self::arbitrary_primitive_witness_with_timestamp_and(
@@ -714,6 +717,7 @@ impl PrimitiveWitness {
                     fee,
                     coinbase,
                     now,
+                    merge_bit,
                 )
             })
             .boxed()
@@ -727,6 +731,7 @@ impl PrimitiveWitness {
         fee: NeptuneCoins,
         coinbase: Option<NeptuneCoins>,
         timestamp: Timestamp,
+        merge_bit: bool,
     ) -> BoxedStrategy<PrimitiveWitness> {
         let num_inputs = input_utxos.len();
         let num_outputs = output_utxos.len();
@@ -796,6 +801,7 @@ impl PrimitiveWitness {
                                 timestamp,
                                 inputs_salt,
                                 outputs_salt,
+                                merge_bit,
                             )
                         })
                         .boxed()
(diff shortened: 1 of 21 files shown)


src/models/blockchain/transaction/primitive_witness.rs
```diff
@@ -567,6 +567,7 @@ impl PrimitiveWitness {
         num_inputs: Option<usize>,
         num_outputs: usize,
         num_public_announcements: usize,
+        merge_bit: bool,
```
aszepieniec, line 570: Unless I'm mistaken, this function isn't called with `merge_bit` set to `true`. If that is correct, don't complicate the function signature.

And if this function is called with `merge_bit` set to `false` 99% of the time, then the 1% does not justify complicating the function signature.

It is opaque to me why you want the this extra argument, but if you really want to keep it then I would suggest to proceed as follows:
 - Rename `arbitrary_with_size_numbers` with its current signature to `arbitrary_with_size_numbers_and_merge_bit`.
 - Wrap that function into a new one `arbitrary_with_size_numbers` which fixes that argument.

src/models/blockchain/transaction/validity/single_proof.rs
```diff
@@ -794,49 +794,98 @@ mod test {
...
-            let txk_mast_hash_as_input_as_public_input =
-                PublicInput::new(txk_mast_hash.reversed().values().encode());
+            let good_witness = SingleProofWitness::from_collection(good_proof_collection);
+            positive_prop(good_witness);
 
-            let nondeterminism = single_proof_witness.nondeterminism();
+            // Setting the `merge_bit` must make program crash, as this bit may
+            // only be set to true through the execution of the merge branch.
+            let bad_primitive_witness =
+                PrimitiveWitness::arbitrary_with_size_numbers(Some(6), 6, 7, true)
```
aszepieniec, line 833: As an alternative to passing everything through the function interface, consider sampling an arbitrary object and then modifying it. The `merge_bit` is rather free of internal consistency dependencies, so I don't think it will be a problem.

aszepieniec (changes requested):
The problem with arbitrary/test-strategy for complex structures like `PrimitiveWitness` is that the many knobs interfere with each other and cannot always be independently turned. Using one framework to generate arbitrary objects for *negative* tests quickly devolves into a spaghetti mess of hacks upon workarounds.

It relates to the exploding complexity of pairs. Because you have $n$ fields, and $n(n-1)/2$ relations that come with constraints. The second you fix something in one place, you break it in several others.

To try to tame this mess I propose:
 - *Keep simple function signatures.* Note that this commandment is different from "Keep function signatures simple." because you are allowed to add new functions with complex signatures as long as you don't expose them to the user much/by-default/in-many-places.
 - *Whenever possible, generate then modify.* This obviously doesn't work when some level of internal consistency is required. But when it does work it is preferable to adding more logic to the already complex arbitrary generators.

In this particular one would suggest sacrificing a principle that I otherwise endorse:
 - *(For negative tests) it's okay to duplicate code.* In particular, I have spent an order of magnitude more agonizing work shoe-horning *prematurely deduplicated code* into something that works, than fixing the same bug again and again across duplications.

I'm eager to learn how experts deal with the exploding complexity of test case generators for *negative* tests.

As for this PR: please refactor into something that follows the guidelines I propose, or else propose and argue different principles.

src/models/blockchain/transaction/primitive_witness.rs
Sword-Smith: > Wrap that function into a new one arbitrary_with_size_numbers which fixes that argument.

I'm positively inclined towards this approach.

src/models/blockchain/transaction/validity/single_proof.rs
Sword-Smith: > As an alternative to passing everything through the function interface, consider sampling an arbitrary object and then modifying it. The `merge_bit` is rather free of internal consistency dependencies, so I don't think it will be a problem.

This was my 1st attempt. But `PrimitiveWitness` witness data in the form of authentication paths which will be invalid if I flip the `merge_bit` after the witness has been produced.

Sword-Smith:
Handled off-web in df6e2ec1925e1f258a5e37e92d975b389b0c4b46

PR #306, opened 2024-12-28
refactor: add 'arbitrary-impl' feature-flag

### LINKED ISSUE #199: Hide `Arbitrary` implementations/functions behind a feature flag
Some arbitrary implementations, specifically in `time_lock` are used in the benchmarking crate. So hiding them behind the test flag makes them unimportable. We tried fixing this with cc3a92de5d18d620ff45f3f8f747bfb4babe9533 but that makes the code available in all build profiles so it defeats the purpose, we think.

dan-da:
Closes #199.

This moves all 'impl Arbitrary' and related functions into module 'neptune_arbitrary', which is built
if either:
  1) --features arbitrary-impl
  2) --tests
  3) --test &lt;x>
  4) --benches
  5) --bench &lt;x>

It is necessary to include for tests because some unit tests require this functionality.

This modularization cleanly separates code that is only needed for arbitrary/testing from that which is necessary for debug/release builds.

I verified that both of these give no warnings or errors:

```
$ cargo clippy --all-features
$ cargo +nightly clippy --all-features --all-targets
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,9 @@ track-lock-location = []
 # not generally useful by itself.
 track-lock-time = []
 
+# implements arbitrary::Arbitrary for various types
+arbitrary-impls = []
+
 [dependencies]
 aead = { version = "0.5", features = ["std"] }
 aes-gcm = "0.10"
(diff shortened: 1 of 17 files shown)


Sword-Smith:
Thank you. Merge-conflicts solved, and merged into master.

Issue #288, opened 2024-12-12
Trigger Untriggered Error IDs in `NativeCurrency` with Dedicated Negative Proptests


aszepieniec:
As of dc8f90be597ad5e2daa8297df483bd8fa9543e9f on branch `asz/negative-fees`, the type script `NativeCurrency` has some asserts with error codes that are not triggered yet by any negative proptests.

There is a lot to unpack in this statement, so let me elaborate.

 - UTXOs hold state, and this state evolves as the UTXOs are consumed and produced by transactions. Type scripts determine which state transitions are allowed and which are not.
 - `NativeCurrency` is a type script which regulates amounts of Neptune Cash's native currency. This is where the no-inflation constraint lives.
 - Type scripts are proved in zero-knowledge so just because the VM knows at execution time what the amounts are, does not mean that the node on the network who observes the transaction will have any clue what the amounts are.
 - Type scripts are written in tasm, which is the assembly language native to [TritonVM](https://triton-vm.org/).
 - One of the instructions available in tasm is `assert`,  and we use this to catch malicious attempts to violate the constraints induced by type scripts.
 - There are plenty of `assert`s in `NativeCurrency`, checking various things that could go wrong. Each assert is labeled with an Error ID. The Error IDs live at the top of the file `src/models/blockchain/type_scripts/native_currency.rs`.
 - We have quite a few test cases in the same file. Most of them are proptests, meaning property tests using the proptest and test-strategy framework. Proptests typically use an *arbitrary* implementation of some sort, and in this case it is an *arbitrary* witness for the prover as it proves (or attempts to) prove the correct execution with graceful halt of the `NativeCurrency` program.
 - We want to have negative tests too. In this case we use the proptest/strategy framework to generate a witness, and then run the `NativeCurrency` based on that witness, and then verify that this malicious strategy triggers some assert. The presence of error IDs allows us to verify that some *specific* assert is triggered.
 - Triggering targeted error IDs forces you to think carefully about the logic embedded in the program. It is not a water-proof audit, but it is pretty strong.
 - Up until now we were trying to minimize code duplication and total code path length of the arbitrary implementations, but it is challenging to generate *arbitrary* witnesses for every error ID using very generic *arbitrary* code.
 - Right now we trigger the majority of error IDs for `NativeCurrency` through dedicated negative proprests, but not all of them.
 - Since it is desirable to trigger *every* error ID with a dedicated negative proptest, it is okay to add specific *arbitrary* implementations to accomplish this.

Issue #307, opened 2024-12-29
Node Hangs on Dashboard Overview


aszepieniec:
I observed that my node started hanging as I was on the Overview screen of the Dashboard.

Last few lines of log:

```
2024-12-29T12:25:22.212611221Z  INFO ThreadId(03) neptune_cash::connect_to_peers: Established incoming TCP connection with [::ffff:92.107.196.15]:47866
2024-12-29T12:25:22.213073721Z  WARN ThreadId(02) neptune_cash::connect_to_peers: Incoming connection refused: MaxPeerNumberExceeded
2024-12-29T12:25:22.213162101Z ERROR ThreadId(02) neptune_cash::main_loop: Got error: Refusing incoming connection. Reason: MaxPeerNumberExceeded
2024-12-29T12:25:46.922136256Z  INFO ThreadId(02) neptune_cash::connect_to_peers: Established incoming TCP connection with [::ffff:107.181.240.234]:40254
2024-12-29T12:25:46.922655525Z  WARN ThreadId(03) neptune_cash::connect_to_peers: Incoming connection refused: MaxPeerNumberExceeded
2024-12-29T12:25:46.922746315Z ERROR ThreadId(03) neptune_cash::main_loop: Got error: Refusing incoming connection. Reason: MaxPeerNumberExceeded
2024-12-29T12:25:50.262841244Z DEBUG ThreadId(01) neptune_cash::main_loop: Timer: Membership proof resync job
2024-12-29T12:25:50.263691083Z DEBUG ThreadId(01) neptune_cash::models::state: Resyncing monitored UTXO number 0, with hash 12387073484739316692,04579393209447820736,18358532265651539143,03022598689779152850,06205812345775899986
2024-12-29T12:25:50.331753721Z  WARN ThreadId(01) neptune_cash::models::state: Could not recover MSMP as transaction appears to be on an abandoned chain
2024-12-29T12:25:50.33273842Z  WARN ThreadId(01) neptune_cash: executed run()::select::mp_resync_timer in 0.069901036 secs.  exceeds slow fn threshold of 0.001 secs.  location: /home/alan/repos/neptune-core/src/main_loop.rs:1468:21
2024-12-29T12:26:10.504135958Z  INFO ThreadId(02) RPC{rpc.trace_id=00 rpc.deadline=2024-12-29T12:26:20.490661799Z otel.kind="server"}: tarpc::server: ReceiveRequest
[...]

PR #308, opened 2024-12-29
fix: Spawn task for block proof verification (#288)

aszepieniec:
- Wrap call to `triton_vm::verify` in `spawn_blocking`.
 - Make function `async` and percolate `async` markers.
 - Sandwich relevant call between debug messages.

All observations point to this change eliminating issue #282. However, we do not have a mechanical explanation as to what causes the apparent fix.

### DIFF
--- a/src/mine_loop.rs
+++ b/src/mine_loop.rs
@@ -698,7 +698,7 @@ pub(crate) async fn mine(
                     continue;
                 }
 
-                if !new_block_found.block.is_valid(&latest_block, Timestamp::now()) {
+                if !new_block_found.block.is_valid(&latest_block, Timestamp::now()).await {
                     // Block could be invalid if for instance the proof and proof-of-work
                     // took less time than the minimum block time.
                     error!("Found block with valid proof-of-work but block is invalid.");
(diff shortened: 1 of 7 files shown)


aszepieniec:
Based on this thought from the discussion in #282 --

> Here is a thought though: we cannot rule out that the peer thread/task crashed, without affecting other running threads/tasks. You expect that the main_loop() would still be running, and Ctrl-C should still respond etc. Then how would we observe that the peer thread/task crashed? One possible outcome could be never terminating connections. Maybe https://github.com/Neptune-Crypto/neptune-core/issues/307 is linked after all.

-- I tend to think that this change limits the crash to the thread/task as opposed to the whole application, but does not fix it. While I do think it is good practice to wrap long-running computations like `triton_vm::verify` inside of `spawn_blocking` calls, I think it is a good idea to delay proceeding with this PR until issue #282 is comprehensively resolved.

aszepieniec:
With #315 I think issue #282 *is* comprehensively resolved and so we can proceed with this PR. Any objections, @dan-da, @Sword-Smith?

Issue #305, opened 2024-12-27
Node Crashes when Proving with too Little RAM


aszepieniec:
A gentleman on Telegram reports that his node crashes from time to time. His machine was configured to mine (composing+guessing) but he mentioned that he didn't succeed to mine any blocks yet. This is a stack trace from the last crash before we (think we) resolved the issue:

```
thread 'tokio-runtime-worker' panicked at /root/neptune-core/src/mine_loop.rs:573:26:
unexpected error while composing: external proving process failed

Caused by:
    proving process did not return any exit code

Stack backtrace:
   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: neptune_cash::mine_loop::mine::{{closure}}
   3: <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll
   4: tokio::runtime::task::harness::Harness<T,S>::poll
note: Some details are omitted, run with RUST_BACKTRACE=full for a verbose backtrace.
Aborted (core dumped)
```

My reading was that the out-of-domain proving process is consuming too much memory and is killed by the operating system, resulting in no return code. Running with the flag `TVM_LDE_TRACE="no_cache"` seems to have fixed the issue. He even mentioned that he mined some blocks after this.

I say "resolved" but I really mean "found a workaround to avoid crashing the node". Properly resolving this issue would be:
 - [ ] (In `triton-vm`) automatically switch to memory-efficient code path when memory is insufficient ([Issue 331](https://github.com/TritonVM/triton-vm/issues/331)).
 - [x] (In `neptune-core`) catch the absence of return code gracefully ~~and, if not already set, set `TVM_LDE_TRACE="no_cache"` and retry~~.
 - [x] (In `neptune-core`) inform user with informative log messages.

dan-da:
> (In neptune-core) catch the absence of return code gracefully and, if not already set, set TVM_LDE_TRACE="no_cache" and retry.

We already catch the absence of return code gracefully.  That's why there is a message about it.  However, we can never be certain of the *reason* for it.   On unix, this just means that the process exited because of a signal, eg could be killed by user, out-of-mem, etc.  On windows it may mean something else.  There isn't a standard way to detect that a process was killed due to out-of-mem or by user.

[...]

PR #310, opened 2025-01-03
feat(mine_loop): Graceful shutdown on composition failure

Sword-Smith:
A possible way of ensuring graceful shutdown on composition failure. Still prints stack traces.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,6 +77,9 @@ use crate::models::state::wallet::WalletSecret;
 use crate::models::state::GlobalStateLock;
 use crate::rpc_server::RPC;
 
+pub const SUCCESS_EXIT_CODE: i32 = 0;
+pub const COMPOSITION_FAILED_EXIT_CODE: i32 = 159;
+
 /// Magic string to ensure other program is Neptune Core
 pub const MAGIC_STRING_REQUEST: &[u8] = b"EDE8991A9C599BE908A759B6BF3279CD";
 pub const MAGIC_STRING_RESPONSE: &[u8] = b"Hello Neptune!\n";
@@ -85,7 +88,7 @@ const MINER_CHANNEL_CAPACITY: usize = 10;
 const RPC_CHANNEL_CAPACITY: usize = 1000;
 const VERSION: &str = env!("CARGO_PKG_VERSION");
 
-pub async fn initialize(cli_args: cli_args::Args) -> Result<()> {
+pub async fn initialize(cli_args: cli_args::Args) -> Result<i32> {
     info!("Starting client on {}.", cli_args.network);
 
     // Get data directory (wallet, block database), create one if none exists
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,3 +1,5 @@
+use std::process;
+
 use anyhow::Result;
 use clap::Parser;
 use neptune_cash::config_models::cli_args;
@@ -41,5 +43,9 @@ pub fn main() -> Result<()> {
 
     tokio_runtime.shutdown_timeout(tokio::time::Duration::from_secs(10));
 
-    res
+    if let Ok(exit_code) = res {
+        process::exit(exit_code)
+    } else {
+        res.map(|_| ())
+    }
 }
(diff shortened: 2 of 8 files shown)


dan-da:
I dunno if it was this PR or not, but tests are failing with this:

```
error[E0554]: `#![feature]` may not be used on the stable release channel
Error:  --> src/lib.rs:7:46
  |
7 | #![cfg_attr(feature = "track-lock-location", feature(async_fn_track_caller))]
  |                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

dan-da:
> A possible way of ensuring graceful shutdown on composition failure. Still prints stack traces.

Please descibe the approach taken so I have an idea what I'm looking at.

I see this in the commit message:

> Ensure that the `graceful_shutdown` routine is executed if the Triton VM
prover process fails -- which in our experience indicates that the
process was killed by the operating system due to a lack of RAM.

I haven't begun review yet, but this statement worries me, as the process can terminate early for any number of reasons, and it may return an exit code or no exit code.  There is no way I am aware of to know with certainty that the cause was out-of-ram.  It could have been a kill -9 or maybe something else, and it also varies by OS.

aszepieniec:
The current behavior, upon proof job failure during composition:
 - absent exit code is caught gracefully by job queue and error code is passed on
 - error is percolated up the call stack in `mine_loop.rs` until it hits a `panic!`
 - the `panic!` kills the application, prints a stack trace, and prints rather* uninformative error messages

I put the scary asterisk after rather because the value of the information that is being presented is a point of discussion. I was certainly able to pinpoint the root cause behind issue #305 but the user in question was not.

The current behavior is undesirable (IMHO) because
 - the stack trace suggests there is a problem with the code when there isn't (the problem is with the machine that's running it)
 - the error messages are tailored to developers (who know that too little RAM can cause prover jobs to be killed) not users
 - the error could happen while database-bound data is still cached in RAM and not persisted to disk yet and in this case that data is lost.

The desired behavior is:
 - shut down the application (after all, the user opted to compose by setting `--compose` and that's not possible)
 - print a helpful error message
 - exit with failing exit code
 - no stack trace.

What the current proposed changes do instead of `panic!`ing is instantiating an error code and percolating code up the call graph, in response to which the main loop initiates a graceful shutdown. In the end of `main`, the (nonzero integer) error code is used to set the exit code. To enable the percolation we had to change a bunch of function signatures.

If this strategy is determined to be adequate (read: invitation for reviewer judgment) then similar strategies should be deployed in two more locations -- there are, after all, three distinct code paths that can spawn prover jobs:
 - [x] composing/mining
 - [ ] initiating transactions
 - [ ] upgrading proofs

One matter of temperate debate was which text to display to the user and where in the code base that text should live. According to the current proposed changes, this text is split across two places (one depending on the other): `mine_loop.rs` lines 691-693, and `prover_job.rs` lines 69-71. Put together they generate this message:

```
Composition failed:
out-of-process triton-vm proving job terminated without exit code. Possibly killed by OS. You might not have enough RAM to construct this  proof.
Try adjusting the environment variables "TVM_LDE_TRACE" and "RAYON_NUM_THREADS".
```

If you feel strongly about the mode of error reporting, feel free to chime in.

aszepieniec:
> What the current proposed changes do instead of panic!ing is instantiating an error code and percolating code up the call graph, in response to which the main loop initiates a graceful shutdown. In the end of main, the (nonzero integer) error code is used to set the exit code. To enable the percolation we had to change a bunch of function signatures.

As I wrote this an alternative came to mind. Right now we are percolating the error code up to to `main` so that *it* can set the process exit code. What if we set the exit code upon catching the error in `mine_loop` and then triggering a graceful shutdown? That would save us a few function signature changes.

aszepieniec:
> > What the current proposed changes do instead of `panic!`ing is instantiating an error code and percolating code up the call graph, in response to which the main loop initiates a graceful shutdown. In the end of main, the (nonzero integer) error code is used to set the exit code. To enable the percolation we had to change a bunch of function signatures.
> 
> As I wrote this an alternative came to mind. Right now we are percolating the error code up to to `main` so that _it_ can set the process exit code. What if we set the exit code upon catching the error in `mine_loop` and then triggering a graceful shutdown? That would save us a few function signature changes.

Never mind this suggestion. The call to `process::exit`, which is what we use now, terminates the application in addition to setting the exit code. There does not seem to be an alternative to call when we just want to set the exit code without (immediately) terminating the application.

PR #315, opened 2025-01-08
fix(`mine_loop`): Segregate rayon threadpool for guessing

### LINKED ISSUE #282: Node Hangs
I observed that my node is hanging. Not sure why. Last few log messages:

```
2024-12-05T21:53:23.854059211Z  INFO ThreadId(02) neptune_cash::peer_loop: Block with height 2210 is valid. mined: 2024-12-05T21:53:17.487+00:00
2024-12-05T21:53:23.854096191Z  WARN ThreadId(02) neptune_cash::peer_loop: Received 1 blocks from peer but incoming blocks are less canonical than current tip.
2024-12-05T21:53:23.854124811Z  WARN ThreadId(02) neptune_cash: executed handle_peer_message()::PeerMessage::Block in 0.320761143 secs.  exceeds slow fn threshold of 0.001 secs.  location: /home/runner/work/neptune-core/neptune-core/src/peer_loop.rs:532:17
2024-12-05T21:53:23.854308201Z  INFO ThreadId(02) neptune_cash::peer_loop: Got new block from peer [::ffff:107.181.240.234]:41672, height 2210, mined 2024-12-05T21:53:17.487+00:00
2024-12-05T21:53:23.854733721Z  INFO ThreadId(02) neptune_cash::mine_loop: Miner task received beta block height 2210
2024-12-05T21:53:23.942761497Z  INFO ThreadId(03) neptune_cash::peer_loop: Block with height 2209 is valid. mined: 2024-12-05T21:32:56.300+00:00
2024-12-05T21:53:24.177705063Z  INFO ThreadId(02) neptune_cash::peer_loop: Block with height 2210 is valid. mined: 2024-12-05T21:53:17.487+00:00
2024-12-05T21:53:24.261931122Z  INFO ThreadId(03) neptune_cash::peer_loop: Block with height 2210 is valid. mined: 2024-12-05T21:53:17.487+00:00
2024-12-05T21:53:24.49263834Z  INFO ThreadId(02) neptune_cash::peer_loop: Block with height 2210 is valid. mined: 2024-12-05T21:53:17.487+00:00
2024-12-05T21:53:24.57640586Z  INFO ThreadId(03) neptune_cash::peer_loop: Block with height 2210 is valid. mined: 2024-12-05T21:53:17.487+00:00
2024-12-05T21:53:24.811175735Z  INFO ThreadId(02) neptune_cash::peer_loop: Block with height 2210 is valid. mined: 2024-12-05T21:53:17.487+00:00
2024-12-05T21:53:24.811267855Z  WARN ThreadId(03) neptune_cash::peer_loop: Received 1 blocks from peer but incoming blocks are less canonical than current tip.
2024-12-05T21:53:24.811291875Z  WARN ThreadId(03) neptune_cash: executed handle_peer_message()::PeerMessage::Block in 1.278402247 secs.  exceeds slow fn threshold of 0.001 secs.  location: /home/runner/work/neptune-core/neptune-core/src/peer_loop.rs:532:17
2024-12-05T21:53:24.811448405Z  WARN ThreadId(03) neptune_cash::peer_loop: Received 1 blocks from peer but incoming blocks are less canonical than current tip.
2024-12-05T21:53:24.811475485Z  WARN ThreadId(03) neptune_cash: executed handle_peer_message()::PeerMessage::Block in 1.278658637 secs.  exceeds slow fn threshold of 0.001 secs.  location: /home/runner/work/neptune-core/neptune-core/src/peer_loop.rs:532:17
2024-12-05T21:53:24.811571325Z  WARN ThreadId(03) neptune_cash::peer_loop: Received 10 blocks from peer but incoming blocks are less canonical than current tip.
2024-12-05T21:53:24.811692815Z  WARN ThreadId(03) neptune_cash: executed handle_peer_message()::PeerMessage::Block in 3.857305739 secs.  exceeds slow fn threshold of 0.001 secs.  location: /home/runner/work/neptune-core/neptune-core/src/peer_loop.rs:532:17
2024-12-05T21:53:24.811788645Z  WARN ThreadId(02) neptune_cash::peer_loop: Received 1 blocks from peer but incoming blocks are less canonical than current tip.
2024-12-05T21:53:24.811810895Z  WARN ThreadId(03) neptune_cash::peer_loop: Received 1 blocks from peer but incoming blocks are less canonical than current tip.
2024-12-05T21:53:24.811843046Z  WARN ThreadId(02) neptune_cash: executed handle_peer_message()::PeerMessage::Block in 1.278586039 secs.  exceeds slow fn threshold of 0.001 secs.  location: /home/runner/work/neptune-core/neptune-core/src/peer_loop.rs:532:17
2024-12-05T21:53:24.811851025Z  WARN ThreadId(03) neptune_cash: executed handle_peer_message()::PeerMessage::Block in 0.957542794 secs.  exceeds slow fn threshold of 0.001 secs.  location: /home/runner/work/neptune-core/neptune-core/src/peer_loop.rs:532:17
2024-12-05T21:53:24.812810554Z  WARN ThreadId(01) neptune_cash: executed run()::select::mp_resync_timer in 0.001559499 secs.  exceeds slow fn threshold of 0.001 secs.  location: /home/runner/work/neptune-core/neptune-core/src/main_loop.rs:1468:21
2024-12-05T21:53:24.994879006Z  WARN ThreadId(03) neptune_cash: executed handle_peer_message()::PeerMessage::BlockRequestByHeight in 0.182773032 secs.  exceeds slow fn threshold of 0.001 secs.  location: /home/runner/work/neptune-core/neptune-core/src/peer_loop.rs:850:17
2024-12-05T21:53:25.11751951Z  WARN ThreadId(02) neptune_cash: executed handle_peer_message()::PeerMessage::BlockRequestByHash in 0.003222898 secs.  exceeds slow fn threshold of 0.001 secs.  location: /home/runner/work/neptune-core/neptune-core/src/peer_loop.rs:826:17
2024-12-05T21:54:49.019940768Z  INFO ThreadId(01) neptune_cash::main_loop: Attempting to reconnect to peer with lost connection: 139.162.193.206:9798
2024-12-05T21:54:49.020016008Z  INFO ThreadId(01) neptune_cash::main_loop: Performing peer discovery
2024-12-05T21:54:49.020037568Z  INFO ThreadId(01) neptune_cash::main_loop: Connecting to peer [::ffff:79.157.213.217]:9798 with distance 2
2024-12-05T21:54:49.020395938Z  INFO ThreadId(03) neptune_cash::connect_to_peers: Established outgoing TCP connection with 139.162.193.206:9798
2024-12-05T21:54:49.020673748Z  INFO ThreadId(02) neptune_cash::connect_to_peers: Established incoming TCP connection with [::ffff:139.162.193.206]:48210
2024-12-05T21:54:49.020956617Z  WARN ThreadId(02) neptune_cash::connect_to_peers: Incoming connection refused: SelfConnect
2024-12-05T21:54:49.021014877Z ERROR ThreadId(02) neptune_cash::main_loop: Got error: Refusing incoming connection. Reason: SelfConnect
2024-12-05T21:54:49.021043447Z ERROR ThreadId(03) neptune_cash::connect_to_peers: An error occurred: Outgoing connection attempt refused. Reason: SelfConnect. Connection closing
2024-12-05T21:54:49.021066747Z  INFO ThreadId(03) neptune_cash::connect_to_peers: Connection closing
2024-12-05T21:55:02.601627685Z  WARN ThreadId(02) neptune_cash::connect_to_peers: Failed to establish connection: Connection timed out (os error 110)
2024-12-05T21:55:02.601689695Z  INFO ThreadId(02) neptune_cash::connect_to_peers: Connection closing
2024-12-05T21:55:04.838148482Z  INFO ThreadId(02) neptune_cash::peer_loop: Got block proposal from peer.
2024-12-05T21:55:05.152632797Z  INFO ThreadId(02) neptune_cash::peer_loop: Rewarding peer ::ffff:66.248.205.33 for NewBlockProposal
2024-12-05T21:55:05.152686637Z  WARN ThreadId(02) neptune_cash: executed handle_peer_message()::PeerMessage::BlockProposal in 0.314540425 secs.  exceeds slow fn threshold of 0.001 secs.  location: /home/runner/work/neptune-core/neptune-core/src/peer_loop.rs:1128:17
2024-12-05T21:55:05.152760117Z  INFO ThreadId(01) neptune_cash::main_loop: Received new favorable block proposal for mining operation.
2024-12-05T21:55:05.153318377Z  INFO ThreadId(03) neptune_cash::mine_loop: Miner received message about new block proposal for guessing.
2024-12-05T21:55:05.154226766Z  INFO ThreadId(572) neptune_cash::mine_loop: Mining on block with 1 outputs and difficulty 704690060. Attempting to find block with height 2211 with digest less than target: 05169561069146539959,10291194454790015845,06096537823767786088,16985605236488289992,00000000026177102695
2024-12-05T21:55:05.224629293Z  INFO ThreadId(02) neptune_cash::peer_loop: Got block proposal from peer.
2024-12-05T21:55:05.224682013Z  WARN ThreadId(02) neptune_cash::peer_loop: Rejecting new block proposal:
Insufficient fee. Proposal was 64;
current fee is: Some(NeptuneCoins(256000000000000005090463958695936))
2024-12-05T21:55:05.225672992Z  WARN ThreadId(02) neptune_cash::peer_loop: Punishing peer 51.15.139.238 for NonFavorableBlockProposal
2024-12-05T21:55:05.225688382Z  WARN ThreadId(02) neptune_cash: executed handle_peer_message()::PeerMessage::BlockProposal in 0.001061049 secs.  exceeds slow fn threshold of 0.001 secs.  location: /home/runner/work/neptune-core/neptune-core/src/peer_loop.rs:1128:17
2024-12-05T21:56:31.651265531Z  INFO ThreadId(02) neptune_cash::peer_loop: Got new block from peer 51.15.139.238:9798, height 2211, mined 2024-12-05T21:56:29.441+00:00
2024-12-05T21:56:32.264116975Z  INFO ThreadId(02) neptune_cash::peer_loop: Got new block from peer 151.115.78.81:9798, height 2211, mined 2024-12-05T21:56:29.441+00:00
```

The dashboard fails to make a connection. Ctrl-C does not work. I ran it with the following command: `neptune-core --max-num-peers 25 --bootstrap --guess --sleepy-guessing --peers 51.15.139.238:9798 --peers 139.162.193.206:9798`. After reboot it seems to be syncing okay but I am noticing this worrying error popping up a bunch of times:

```
2024-12-10T14:25:48.367618193Z ERROR ThreadId(01) neptune_cash::main_loop: Failed to send pause message to miner thread:                                                                                                             
no available capacity                                                                                                                                                                                                                
```

aszepieniec:
Fixes #282.

The issue was caused by interference between rayon parallelism in disparate tokio tasks. On the one hand, the guesser spawned the max number of rayon threads using `repeat`. On the other hand, the peer loop task spawned rayon threads in the course of `triton_vm::verify` using `par_iter`. But since the the rayon threadpool was already occupied and since tokio tasks do not know how to coordinate about rayon's thread pool, the second task ended up stalling.

This PR fixes the problem by managing the rayon parallelism for guessing in a segregated rayon threadpool. It has already been experimentally verified to solve the problem. The [documentation](https://docs.rs/rayon/latest/rayon/struct.ThreadPool.html) for `ThreadPool` explains why:

> If the current thread is part of a different thread pool, it will try to keep busy while the op completes in its target pool, similar to calling [ThreadPool::yield_now()](https://docs.rs/rayon/latest/rayon/struct.ThreadPool.html#method.yield_now) in a loop. Therefore, it may potentially schedule other tasks to run on the current thread in the meantime.

### DIFF
--- a/src/config_models/cli_args.rs
+++ b/src/config_models/cli_args.rs
@@ -113,6 +113,11 @@ pub struct Args {
     #[clap(long)]
     pub sleepy_guessing: bool,
 
+    /// Set the number of threads to use while guessing. When no value is set,
+    /// the number is set to the number of available cores.
+    #[clap(long)]
+    pub guesser_threads: Option<usize>,
+
     /// Determines the fraction of the transaction fee consumed by this node as
     /// a reward either for upgrading a `ProofCollection` to `SingleProof`, or
     /// for merging two `SingleProof`s.
--- a/src/mine_loop.rs
+++ b/src/mine_loop.rs
@@ -98,13 +98,13 @@ async fn compose_block(
 }
 
 /// Attempt to mine a valid block for the network
-#[allow(clippy::too_many_arguments)]
 async fn guess_nonce(
     block: Block,
     previous_block_header: BlockHeader,
     sender: oneshot::Sender<NewBlockFound>,
     composer_utxos: Vec<ExpectedUtxo>,
     sleepy_guessing: bool,
+    num_guesser_threads: Option<usize>,
     target_block_interval: Option<Timestamp>,
 ) {
     // We wrap mining loop with spawn_blocking() because it is a
@@ -126,6 +126,7 @@ async fn guess_nonce(
             sender,
             composer_utxos,
             sleepy_guessing,
+            num_guesser_threads,
             target_block_interval,
         )
     })
@@ -151,18 +152,21 @@ fn guess_worker(
     sender: oneshot::Sender<NewBlockFound>,
     composer_utxos: Vec<ExpectedUtxo>,
     sleepy_guessing: bool,
+    num_guesser_threads: Option<usize>,
     target_block_interval: Option<Timestamp>,
 ) {
     // This must match the rules in `[Block::has_proof_of_work]`.
     let prev_difficulty = previous_block_header.difficulty;
     let threshold = prev_difficulty.target();
     let input_block_height = block.kernel.header.height;
+    let threads_to_use = num_guesser_threads.unwrap_or_else(rayon::current_num_threads);
     info!(
-        "Mining on block with {} outputs and difficulty {}. Attempting to find block with height {} with digest less than target: {}",
+        "Guessing with {} threads on block {} with {} outputs and difficulty {}. Target: {}",
+        threads_to_use,
+        block.header().height,
         block.body().transaction_kernel.outputs.len(),
         previous_block_header.difficulty,
-        block.header().height,
-        threshold
+        threshold.to_hex()
     );
 
     // note: this article discusses rayon strategies for mining.
@@ -172,9 +176,7 @@ fn guess_worker(
     // see:  https://docs.rs/rayon/latest/rayon/fn.max_num_threads.html
     let block_header_template = block.header().to_owned();
     let (block_body_mast_hash_digest, appendix_digest) = precalculate_mast_leafs(&block);
-    let rayon_threads_available = rayon::current_num_threads();
 
-    let threads_to_use = rayon_threads_available;
     let pool = ThreadPoolBuilder::new()
         .num_threads(threads_to_use)
         .build()
@@ -656,6 +658,7 @@ pub(crate) async fn mine(
                 guesser_tx,
                 composer_utxos,
                 cli_args.sleepy_guessing,
+                cli_args.guesser_threads,
                 None, // using default TARGET_BLOCK_INTERVAL
             );
 
@@ -1333,6 +1336,7 @@ pub(crate) mod mine_loop_tests {
         );
 
         let sleepy_guessing = false;
+        let num_guesser_threads = None;
 
         guess_worker(
             block,
@@ -1340,6 +1344,7 @@ pub(crate) mod mine_loop_tests {
             worker_task_tx,
             coinbase_utxo_info,
             sleepy_guessing,
+            num_guesser_threads,
             None,
         );
 
@@ -1407,6 +1412,7 @@ pub(crate) mod mine_loop_tests {
         assert_eq!(ten_seconds_ago, initial_header_timestamp);
 
         let sleepy_guessing = false;
+        let num_guesser_threads = None;
 
         guess_worker(
             template,
@@ -1414,6 +1420,7 @@ pub(crate) mod mine_loop_tests {
             worker_task_tx,
             coinbase_utxo_info,
             sleepy_guessing,
+            num_guesser_threads,
             None,
         );
 
@@ -1492,6 +1499,7 @@ pub(crate) mod mine_loop_tests {
 
         // set initial difficulty in accordance with own hash rate
         let sleepy_guessing = false;
+        let num_guesser_threads = None;
         let num_outputs = 0;
         let hash_rate =
             estimate_own_hash_rate(Some(target_block_interval), sleepy_guessing, num_outputs).await;
@@ -1566,6 +1574,7 @@ pub(crate) mod mine_loop_tests {
                 worker_task_tx,
                 composer_utxos,
                 sleepy_guessing,
+                num_guesser_threads,
                 Some(target_block_interval),
             );
 
--- a/src/mine_loop.rs
+++ b/src/mine_loop.rs
@@ -1,6 +1,5 @@
 pub(crate) mod composer_parameters;
 
-use std::cmp;
 use std::cmp::max;
 use std::time::Duration;
 
@@ -174,12 +173,8 @@ fn guess_worker(
     let block_header_template = block.header().to_owned();
     let (block_body_mast_hash_digest, appendix_digest) = precalculate_mast_leafs(&block);
     let rayon_threads_available = rayon::current_num_threads();
-    if rayon_threads_available < 2 {
-        error!("Cannot guess when less than two threads are available to rayon");
-        return;
-    }
 
-    let threads_to_use = cmp::max(1, rayon_threads_available.saturating_sub(2));
+    let threads_to_use = rayon_threads_available;
     let pool = ThreadPoolBuilder::new()
         .num_threads(threads_to_use)
         .build()
--- a/src/mine_loop.rs
+++ b/src/mine_loop.rs
@@ -1,5 +1,6 @@
 pub(crate) mod composer_parameters;
 
+use std::cmp;
 use std::cmp::max;
 use std::time::Duration;
 
@@ -16,9 +17,9 @@ use rand::rngs::StdRng;
 use rand::Rng;
 use rand::SeedableRng;
 use rayon::iter::ParallelIterator;
+use rayon::ThreadPoolBuilder;
 use tasm_lib::prelude::Tip5;
 use tasm_lib::triton_vm::prelude::BFieldCodec;
-use tasm_lib::twenty_first::prelude::CpuParallel;
 use tasm_lib::twenty_first::prelude::MerkleTreeMaker;
 use tokio::select;
 use tokio::sync::mpsc;
@@ -49,6 +50,7 @@ use crate::models::state::wallet::transaction_output::TxOutputList;
 use crate::models::state::GlobalState;
 use crate::models::state::GlobalStateLock;
 use crate::prelude::twenty_first;
+use crate::twenty_first::util_types::merkle_tree::CpuParallel;
 use crate::COMPOSITION_FAILED_EXIT_CODE;
 
 async fn compose_block(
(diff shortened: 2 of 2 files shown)


Sword-Smith (approved):
LGTM

dan-da:
Thoughts:

1. I like this solution better than the other commit I saw that spawned a lot of tokio tasks instead of using rayon in the mining loop.  I had some concerns about that one.

2. I like the addition of the cli arg to control # of guessing threads.  

3. I don't feel like I have a good grasp of what the root cause was yet.  @aszepieniec if you feel that you do, could you perhaps reword the explanation to break it down for me, step by step?

In particular my understanding of how it has been working is that there is a single global rayon threadpool.  So any ParallelIterator workers would execute on that global rayon threadpool.  The mining-loop paralleliterator was already running in a spawn_blocking, so it should be in a dedicated tokio thread.   If there's any interaction with tokio workers it's likely due to the call to triton_vm::verify() executing paralleliterator directly inside a regular tokio task.

> If the current thread is part of a different thread pool, it will try to keep busy while the op completes in its target pool, similar to calling ThreadPool::yield_now() in a loop. Therefore, it may potentially schedule other tasks to run on the current thread in the meantime.

If I'm reading that quote correctly and in context it is talking about a situation where install() is called by a thread inside an existing rayon threadpool.  That situation doesn't apply here afaict.   Also calling yield_now() in a loop should result in a hot loop with full CPU utlization.  As I understand it, we were seeing the opposite with top reporting Idle CPU.   But again, afaik there has only been a single rayon global threadpool.

4. Let's take a high-level view of the situation with the change in this PR.  iiuc, there are now two rayon threadpools:  a) the global threadpool which is used by ParallelIterator in triton_vm::verify(), and b) a threadpool that is local to mine_loop.  For some reason (unclear to me) this arrangement seems to fix the particular hang issue we've been seeing.  But there is still the global threadpool that verify() is running in, and verify() can still be called multiple times concurrently.  Are we certain that's fine for any N of concurrency?

And back to a fundamental question:  why does having two separate rayon threadpools fix the hang issue?

5. the proof is in the pudding.  Clearly this is fixing the particular hang issue we've observed, so it's an improvement.  I just hope a) we can collectively come to a clear understanding of the root cause, and b) that with this change we haven't just placed a band-aid, only for it to popup in another scenario.

6. I forget.  Did we ever try putting the verify() calls inside spawn_blocking(), and if so what was the result?    I think we were fearful that would just hide the issue if it occurred again.  But if there is an interaction happening that stalls tokio worker threads, perhaps that would be the most correct solution... and a single global rayon threadpool would then work without hanging?

7. I'm not clear on what costs/benefits there are to having multiple rayon threadpools.  It would increase CPU context switching some, maybe not a lot.  A clear benefit seems to be that we can independently control the number of threads in each threadpool, so it mostly seems worth keeping regardless.

aszepieniec:
> I don't feel like I have a good grasp of what the root cause was yet. @aszepieniec if you feel that you do, could you perhaps reword the explanation to break it down for me, step by step?

I do, and I'm happy to elaborate but I'm afraid I would end up using many of the same words that weren't compelling the first go around. So I decided instead to write a minimum working demonstration exhibiting the behavior and the fix. Here it is: https://github.com/aszepieniec/tokio-rayon-interference

This demo reliably reproduces the same behavior on different machines.

I added an elaborate note to the repository's README.md file because I figured I might as well help out others who stumble into this bug. You should probably read that before you continue; otherwise I'm afraid my references might not make any sense.

> If I'm reading that quote correctly and in context it is talking about a situation where `install()` is called by a thread inside an existing rayon threadpool. That situation doesn't apply here afaict.
It's definitely confusing. Here is my reading, with annotations:

> If the current thread [*i.e.*, the one calling `pool.install`] is part of a different thread pool [different from `pool`], it [the current thread] will try to keep busy while the `op` [guessing] completes in its target pool [`pool`]

It's saying the current thread tries to stay busy while the `pool.install` job is running. The current thread, in turn, can yield to other tokio tasks -- such as the one which does the verification.

> As I understand it, we were seeing the opposite with top reporting Idle CPU.

The idle CPU can be explained as a consequence of using `--sleepy-guessing`, which sleeps for 100 ms in between every guess. I do not recall running `htop` when this flag was dropped.

> But there is still the global threadpool that verify() is running in, and verify() can still be called multiple times concurrently. Are we certain that's fine for any N of concurrency?

If my understanding is correct, they (the verify tasks) will be sequenced by the global thread-pool, and as a consequence they will stall until it's their turn. And I struggle to imagine a better solution. Even if you could find a way to make them share resources simultaneously, the amount of work does not change and so you would have to wait the same time but without getting the results of tasks 1 -- (N-1) early.

Mind you, `triton_vm::verify` takes around 50 ms, and only some steps invoke parallelism. Furthermore, we are not sure parallelism actually speeds things up because there is the overhead of dividing and distributing work across cores.

>  I just hope [...] that with this change we haven't just placed a band-aid, only for it to popup in another scenario.

This is a valid concern. If you were to ask me for a litmus test to distinguish band-aid-and-morphine from surgical cures you might get a checklist like this:

 - [ ] *Can you explain the root cause in simple terms?* I guess I failed in the first try but the elaborate README.md seeks to remedy that.
 - [x] *Does the hypothesized root cause explain all observations?* I believe the answer is yes.
 - [x] *Can you reliably reproduce the issue?* I believe I have.
 - [x] *Does the fix reliably make the issue go away?* Yes.

> I forget. Did we ever try putting the verify() calls inside spawn_blocking(), and if so what was the result? I think we were fearful that would just hide the issue if it occurred again. But if there is an interaction happening that stalls tokio worker threads, perhaps that would be the most correct solution... and a single global rayon threadpool would then work without hanging?

Yes, we did. At first I thought it solved the issue -- running log output, responsive dashboard, *etc*. At some point I recall noticing that the number of peer connections reported by the dashboard was very large, 80 or so. (More nodes than I think ever connected at the same time.) I didn't think it was related at the time. Looking back, I think this mechanic was at play:
 - my peer thread stalls
 - connection closed by peer, but my peer thread hasn't processed it yet
 - some time later, same node connects again but with new IP address
 - same node listed twice.

In the end, we rolled back the use of `spawn_blocking` because the possibility existed that it did not solve anything and instead made the problem *seem* to disappear -- like accumulating crashed tasks without crashing the main task.

My minimal working demo shows that indeed it does nothing to fix the issue.

> I'm not clear on what costs/benefits there are to having multiple rayon threadpools. It would increase CPU context switching some, maybe not a lot. A clear benefit seems to be that we can independently control the number of threads in each threadpool, so it mostly seems worth keeping regardless.

I'm not sure it is possible to make the issue go away without either a) making at least one of both tasks (guessing and verifying) sequential, or b) segregating rayon mempools. I *think* it is because rayon thread-pools are inherently synchronous and can only ever have one client. If true, then yes it would imply that there is a lot of context switching going on when two distinct thread-pools are taking turns.

jan-ferdinand:
Thanks for taking the time to write up the demo, both in terms of code and the readme.

You mention noise in the output due to buffering. It _might_ reduce the noise when you [lock `stdout`](https://doc.rust-lang.org/std/io/struct.Stdout.html) and [flush](https://doc.rust-lang.org/std/io/trait.Write.html#tymethod.flush) it immediately. Don't forget to drop the write guard afterward. That said, I haven't run the demo, and it might not be a problem in practice.

jan-ferdinand:
Here are my results from running the demo in all 12 configurations. Summarizing the picture, on my machine, the verifier appears to be stalled in configurations `par_guess_global` + `par_verify_spawn` and `par_guess_global` + `par_verify_direct`. No other configuration results in a stalling verifier.

image

PR #317, opened 2025-01-09
Derive within spending key  alphanet v5 compat pr

dan-da:
Supersedes #304

This modifies generation_address so that the same addresses will
be derived from a given wallet secret as happened at tag alphanet-v5 (Jan 6, 2024)

A test exists which verifies this:
    wallet_tests::generation_key_derivation::verify_derived_generation_addrs()

Significant changes since #304:

* compatibility with alphanet-v5 addresses:
    * DerivationIndex: u128 --> u64
    * replace GenerationSpendingKey::seed with secret, index
    * rename GenerationSpendingKey::from_seed() --> from_secret_and_index()
    * add GenerationSpendingKey::seed(secret, index) which uses same
        hash(secret, generation_flag, index) construction as
        WalletSecret::nth_generation_key() used to.
    * WalletSecret::master_generation_key is now DerivationIndex::MAX so
        that the 0th key can have a parent.
    * GenerationSpendingKey::derive_child() now uses wrapping_add() so that
        master key (index = DerivationIndex::MAX) can derive 0th key.

* add KeyTypeSeed to abstract over seed types

Minor changes:
* update a couple Block tests to use mainnet.
* rename SpendingKeyIter::curr      --> curr_fwd
* rename SpendingKeyIter::curr_back --> curr_rev
* updates test verify_derived_generation_keys with newly generated keys
    because serialized key format has changed.
* add some missing doc-comments for key/addr in generation_address.rs
* update GenerationSpendingKey deserialization impl
* add impl From<&GenerationSpendingKey> for GenerationReceivingAddress
* remove GenerationReceivingAddress::from_spending_key()
* update Arbitrary impls for GenerationSpendingKey seeds
* add a couple of Arbitrary impl to the feature flag.
* adapt tests to changes

-------

Of special note for review:

1. note the fields that changed in GenerationSpendingKey and the method from_seed() became from_secret_and_index().

2. consider security impiications that *every* key now embeds the wallet master secret.  This means that **if any key is compromised every key in the wallet is compromised**.  If this is not acceptable (_and it probably shouldn't be_), I believe it is a deal-breaker for backwards compatibility.  I don't believe there's any other way to have both self-deriving keys and backwards compatibility, but feel free to prove me wrong.... 

3. compare impl of GenerationSpendingKey::seed() with WalletSecret::nth_generation_spending_key() in master.

4. also see GenerationSpendingKey::derive_child()

5. note how the master generation key in WalletSecret is DerivationIndex::MAX, and that GenerationSpendingKey::derive_child() allows wrapping add, so that the master key can be parent of the 0th key.  I initially made an impl that did not do this, and instead the master-key is the 0th key, but this complicated the implementation of the nth_* methods and made them a bit ugly and non-obvious.  I feel this impl is overall more elegant.

6. the DerivationIndex is a u64 instead of u128 -- for all key types.  I think it's fine.... that's still a huge amount of possible addresses for a wallet.

7. review test: wallet_tests::generation_key_derivation::verify_derived_generation_addrs()

--------

with regards to point 2, see https://github.com/Neptune-Crypto/neptune-core/pull/317#issuecomment-2582943121

<strike>
With regards to point 2:  my conclusion after some reflection is that the weakened security risk is not worth it, and we should prefer #304 rather than this PR.    If we are not willing to accept the weakened security, we have a choice:

a. implement self-derived keys in a backwards incompatible way, eg #304.
b. prioritize backwards compatiiblity and do not use self-deriving keys at all.

I very much prefer (a) even if it means some short-term pain because we have to obtain new addresses from 3rd parties.   I think that self-deriving keys bring a lot of value for the ecosystem that will be lost if every piece of software must have access to the wallet secret directly or via RPC. 
</strike>


aszepieniec:
I can't pass judgment because I'm confused about this summary. Please clarify and/or confirm.

 - There are at least four constructions being discussed: (1) historical/current master; (2) first self-deriving impl from #304; (3) second self-deriving impl / this PR; (4) master-key-only.
 - The keys and addresses from (3) / this PR match with legacy keys from alphanet-v5.
 - The keys and addresses from (1) / current master do not match with legacy keys from alphanet-v5.
 - In proposal (1), child keys are derived by/from `WalletSecret` whereas proposals (2), (3), (4) they are derived from other keys.
 - The line `key = ...` describes the type of keys under that proposal.
 - The method `wrapping_add` modifies the zeroth argument (the thing before the dot) and returns the result of the wrapping add.
 - `master_key` denotes the secret string of random bits that lives in `WalletSecret` and is the thing that is imported through the CLI command `import-seed-phrase`.
 - What does `key::seed() = hash(wallet_secret, generation_flag, index)` do?

Based on abstract consideration of the properties (ignoring how they arise from the concrete implementations) I tend towards this position. (But I reserve the right to change my mind after understanding the matter better.)
 - Supporting legacy keys is the top priority.
 - Given that legacy keys are supported, self-derivation of keys is a feature even if it comes with the property that a compromised child compromised the entire family.
 - Self-deriving keys without that property are preferable.
 - I think it is silly and needlessly complex to limit self-derivation to distinguished keys based on the above bullet point.

dan-da:
> I can't pass judgment because I'm confused about this summary. Please clarify and/or confirm.
> 
> * There are at least four constructions being discussed: (1) historical/current master; (2) first self-deriving impl from [feat: SpendingKey self derivation, parallel iteration #304](https://github.com/Neptune-Crypto/neptune-core/pull/304); (3) second self-deriving impl / this PR; (4) master-key-only.

yes.

> * The keys and addresses from (3) / this PR match with legacy keys from alphanet-v5.

yes.

> * The keys and addresses from (1) / current master do not match with legacy keys from alphanet-v5.

no.  (they do match).

> * In proposal (1), child keys are derived by/from `WalletSecret` whereas proposals (2), (3), (4) they are derived from other keys.

In all cases keys are originally derived from WalletSecret.   What differs is the method of derivation.

(1) and (4) use identical derivation.   ie, (4) goes full circle back to (1).   The only difference is that with (4), it would be possible to export the wallet-secret via RPC, and clients could directly derive from it themselves.

(2) derives a master key from WalletSecret, and subsequent keys from the master key.

(3) includes WalletSecret and an index inside each key.

> * The line `key = ...` describes the type of keys under that proposal.

It describes the primary data/fields held by the key.  (all other fields are derivative, cached).

> * The method `wrapping_add` modifies the zeroth argument (the thing before the dot) and returns the result of the wrapping add.

It is u64::wrapping_add().  It behaves like "+" except when adding u64::MAX + 1 it returns 0 instead of a panic.

> * `master_key` denotes the secret string of random bits that lives in `WalletSecret` and is the thing that is imported through the CLI command `import-seed-phrase`.

master_key is a concept introduced in (2).    It is conceptually similar to xprv in bip32.  It is derived from hash(WalletSecret, generation_key).  The 0th key is derived from hash(master_key, 0).  The 1st key as hash(0th_key, 0) and so on.  

> * What does `key::seed() = hash(wallet_secret, generation_flag, index)` do?

In (3) the WalletSecret and Index are stored in each key, instead of a Digest.  But GenerationSpendingKey has other fields/methods that are derived from a seed (Digest).  So a new seed() method is added for generating the seed Digest.   It uses the same derivation scheme as (1), which is what enables address compatibility.

```
    fn seed(secret: XFieldElement, index: common::DerivationIndex) -> Digest {
        Hash::hash_varlen(
            &[
                secret.encode(),
                vec![GENERATION_FLAG, BFieldElement::new(index)],
            ]
            .concat(),
        )
    }
```

> * Given that legacy keys are supported, self-derivation of keys is a feature even if it comes with the property that a compromised child compromised the entire family.
> * Self-deriving keys without that property are preferable.
> * I think it is silly and needlessly complex to limit self-derivation to distinguished keys based on the above bullet point.

I'm not sure I understand the last point.

regardless, I've come to the conclusion that we should be choosing between (1) and (4).    The primary difference between them is whether we allow the wallet-secret to leave neptune-core via the rpc api or not.  If so, then client's can derive their own keys.  If not, neptune-core will have to do this work on their behalf.  That can be a lot of data transfer given the size of the keys/addresses.

If we choose (4) then there are still some useful things we can use from this PR such as the iterators.

The two things I've realized since I began working on all this are:

 1. we already have client-derivable keys, ie keys that could be derived outside neptune-core.  We just don't actually expose the wallet-secret (and a helpful type) to the clients.  It's a small change to enable that.   I say "client-derivable" and not "self-derivable" because in (1) and (4) technically the keys cannot derive themselves, they require help from a type that knows the wallet-secret.  "client-derivable" works just as well as "self-derivable" for any purpose I can think of.  obv a better name/terminology is needed.
 
 2. Impls (2) and (3) actually have worse security properties than (1) and (4) because compromising an individual SpendngKey can result in some (2) or all (3) spending keys being compromised.  Whereas in (1) and (4) no other key is compromised.  

Hopefully thats clear.   also I'm happy to jump in a chat or call to discuss further.

dan-da:
> I wonder whether you ever considered this scheme? If so, what what made you decide against it?
> 
> ```
> key_i = Hash(key_{i >> 1}, i % 1)
> master_key = key_0
> ```
> 
> It arranges all keys into a balanced binary tree such that every key is the root of some branch, and every key has two immediate children. If a key is compromised, only its branch is compromised.

no I didn't consider it. When working on (2) I was just trying to make a first-draft self-deriving impl and wasn't focused on security implications.   I figured you might have good suggestions like that once the basic concept was proven/working. 

But then the issue was raised about compatibility with legacy addresses, so that became the focus.

and this scheme would not be compatible with old addresses, so it seems a non-starter, no?

> If a key is compromised, only its branch is compromised.

the earlier/older a key is, the larger that branch can be.   So it is still strictly worse than (1) and (4) that do not compromise any other key.   (unless the wallet-secret itself is compromised of course, but that is true for all schemes).

dan-da:
Ugh, this is a tough one since we can't do xpub style address derivation.

I asked google gemini this question:

_is there any bitcoin-core rpc api that returns the wallet seed?_

> No, there is no Bitcoin Core RPC API that directly returns the wallet seed.
> 
> Security Risk: **Exposing the seed would be a major security vulnerability.** The seed is the master key to your entire wallet.
> 
> Alternative: The closest you can get is using the dumpprivkey RPC command. This command can be used to export the private keys for individual addresses within your wallet. However, this does not provide the actual seed itself.
> 

That was the answer I expected.   And that's not really a precedent I'd like to set/encourage.

-- brainstorming --

One possibility would be to require that the seed is stored encrypted on disk, ie require a password upon wallet creation.  neptune-core would then need a password each time it is started.  The wallet-seed could be passed to clients via RPC in encrypted form *only* and the client would have to ask user for their password to decrypt it.

------

regardless of deriving keys on the client, eventually we will want/need to encrypt the wallet seed on disk, perhaps optionally.   

-------

easiest: We could just error on the side of caution and forget about client-side derivation for the time being.

------

One other distinction/possibility comes to mind.

return wallet-secret from rpc API is problematic specifically because it is a network api.

However, we could just return key indexes and *assume* the client has access to the wallet-secret.

In the present code (master) a local client can in fact access the wallet-secret directly via the filesystem and then derive keys.    (and so could malware).

we would need to think about how to handle remote clients without wallet access for affected API(s).

dan-da:
closing this as there are security concerns with self-derived keys and anyway we know we can enable key derivation on the client with the current implementation, provided the client has access to the wallet-secret.    So that could be done at any time.

I may still cherry-pick the key iterators and maybe some other bits from this PR that could make sense in master.

PR #322, opened 2025-01-13
Restrict update-branch of single proof

Sword-Smith:
Prevent composers from reusing block transactions across blocks by preventing coinbase transactions from being updated and by forcing the `merge_bit` to be false in the new kernel in the update-branch.


Sword-Smith:
As discussed, I believe that preventing coinbase transactions from being updated is sufficient to prevent this attack: Anyone who wants to collect block subsidies cannot reuse old block transactions but have to build their own where they might as well merge-in non-empty transactions to collect transaction fees, if any such transaction is present in the mempool

So I'm OK leaving the merge bit rules as they were prior to this PR: That the `merge_bit` value is unchanged by the execution of the update branch.

PR #323, opened 2025-01-14
chore(deps): Upgrade dependencies

jan-ferdinand:


src/mine_loop.rs
```diff
@@ -344,7 +343,7 @@ fn guess_nonce_iteration(
         appendix_digest,
         Digest::default(),
     ];
-    let block_hash = CpuParallel::from_digests(&digests).unwrap().root();
+    let block_hash = MerkleTree::par_new(&digests).unwrap().root();
```
Sword-Smith, line 346: I would prefer to use the sequential version here.

src/models/blockchain/block/block_body.rs
```diff
@@ -143,7 +142,7 @@ impl MastHash for BlockBody {
                     digests.push(Digest::default());
                 }
 
-                twenty_first::prelude::CpuParallel::from_digests(&digests).unwrap()
+                twenty_first::prelude::MerkleTree::par_new(&digests).unwrap()
```
Sword-Smith, line 145: Sequential here also? Input has length 4.

PR #336, opened 2025-01-19
feat: True claims cache

aszepieniec:
Wrap calls to `triton_vm::verify` into a function `verify` living in `proof_abstractions`.

If the test flag is disabled, `verify` just:
 - selects the right (*i.e.*, default) `Stark` parameters
 - calls `triton_vm::verify` inside of a `task::spawn_blocking`.

If the test flag is enabled, `verify` also:
 - checks the cache of claims before calling `triton_vm::verify` and even bypasses said call if the claim lives there and instead returns true early
 - populates the cache of claims with claims that pass `triton_vm::verify`.

Furthermore, if the test flag is enabled then there is a new function `cache_true_claim`. This function inserts a claim into the cache.

By using this interface, some tests (in particular, tests pertaining to archival state, wallet state, or peer interactions) can now avoid producing proofs. Instead, they should populate the true claims cache with the correct claim and the proof field with a bogus proof (that now won't be read anyway).

Note that this optimization does not cover all tests: some tests are about verifying that proof production succeeds. It would be improper to avoid producing proofs there.

### DIFF
--- a/src/mine_loop.rs
+++ b/src/mine_loop.rs
@@ -884,6 +884,7 @@ pub(crate) mod mine_loop_tests {
     use crate::models::blockchain::type_scripts::neptune_coins::NeptuneCoins;
     use crate::models::proof_abstractions::mast_hash::MastHash;
     use crate::models::proof_abstractions::timestamp::Timestamp;
+    use crate::models::proof_abstractions::verifier::verify;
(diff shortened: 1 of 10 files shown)


Sword-Smith:
Merged in 3928e8c6cb82944d7821cbbc8247aad9414bd77d

PR #337, opened 2025-01-19
test: Add test for sync challenge phase

aszepieniec:
- Stylistic changes.
 - Use verify wrapper, which uses true claims cache to bypass call to `triton_vm::verify` for tests.
 - Move logic from `peer_loop` to `GlobalState`.
 - Add rng on `PeerLoopHandler` to derandomize randomness in challenge.
 - Add (positive) test with mocked communication between Alice and Bob whereby Bob triggers Alice to go into sync mode.

### DIFF
--- a/src/models/state/mod.rs
+++ b/src/models/state/mod.rs
@@ -346,8 +346,17 @@ impl GlobalState {
         height
     }
 
-    /// Return a boolean indicating if synchronization mode should be entered
-    pub(crate) fn should_enter_sync_mode(
+    /// Determine whether the conditions are met to enter into sync mode.
+    ///
+    /// Specifically, compute a boolean value based on
+    ///  - whether the foreign cumulative proof-of-work exceeds that of our own;
+    ///  - whether the foreign block has a bigger block height and the height
+    ///    difference exceeds the threshold set by the CLI.
+    ///
+    /// The main loop relies on this criterion to decide whether to enter sync
+    /// mode. If the main loop activates sync mode, it affects the entire
+    /// application.
+    pub(crate) fn sync_mode_criterion(
         &self,
         max_height: BlockHeight,
         cumulative_pow: ProofOfWork,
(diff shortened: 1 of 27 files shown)


src/models/state/mod.rs
```diff
@@ -1472,6 +1482,92 @@ impl GlobalState {
...
+    /// Returns (parent, child) pair of blocks.
+    pub(crate) async fn fetch_block_pair(&self, child_digest: Digest) -> Option<(Block, Block)> {
+        let child = self
+            .chain
+            .archival_state()
+            .get_block(child_digest)
+            .await
+            .expect("fetching block from archival state should work.");
+        let Some(child) = child else {
+            warn!("Got sync challenge for unknown tip");
```
Sword-Smith, line 1494: This log message is not appropriate now that this function has been moved to `GlobalState`.

src/models/state/mod.rs
```diff
@@ -1472,6 +1482,92 @@ impl GlobalState {
         // Ok(())
     }
 
+    /// Returns (parent, child) pair of blocks.
+    pub(crate) async fn fetch_block_pair(&self, child_digest: Digest) -> Option<(Block, Block)> {
```
Sword-Smith, line 1486: I think this function crashes if `child_digest` is the genesis block. We check that it's not in the function `response_to_sync_challenge`, but if this function is public, it should probably return `None` in that case.

src/models/state/mod.rs
```diff
@@ -1472,6 +1482,92 @@ impl GlobalState {
...
+                .try_get_leaf(h.into())
+                .await
+            else {
+                bail!("could not get leaf from archival block mmr");
+            };
+            let Some((p, c)) = self.fetch_block_pair(child_digest).await else {
+                bail!("could not fetch indicated block pair");
+            };
+
+            block_mmr_mps.push(
```
Sword-Smith, line 1542: If `tip` in request does not match own tip, then these MMR membership proofs will not be valid. Is that OK? We might as well check for that here and then decide what to do in that case.

Sword-Smith: Addressed with comment in function.

PR #343, opened 2025-01-22
feat(archival_state): Add block's PoW witness to `BlockRecord`

Sword-Smith:
Add a new field to BlockRecord for the block's PoW witness which allows for the calculation of the block hash without needing knowledge of the entire block.

### DIFF
--- a/src/models/state/archival_state.rs
+++ b/src/models/state/archival_state.rs
@@ -22,6 +22,8 @@ use crate::database::storage::storage_schema::traits::*;
 use crate::database::NeptuneLevelDb;
 use crate::database::WriteBatchAsync;
 use crate::models::blockchain::block::block_header::BlockHeader;
+use crate::models::blockchain::block::block_header::BlockHeaderWithBlockHashWitness;
+use crate::models::blockchain::block::block_header::HeaderToBlockHashWitness;
 use crate::models::blockchain::block::block_height::BlockHeight;
 use crate::models::blockchain::block::mutator_set_update::MutatorSetUpdate;
 use crate::models::blockchain::block::Block;
@@ -389,6 +391,7 @@ impl ArchivalState {
             min_aocl_index: new_block.mutator_set_accumulator_after().aocl.num_leafs()
                 - num_additions,
             num_additions,
+            block_hash_witness: HeaderToBlockHashWitness::from(new_block),
         }));
 
         block_index_entries.push((file_record_key, BlockIndexValue::File(file_record_value)));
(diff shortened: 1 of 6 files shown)


src/models/state/archival_state.rs
```diff
@@ -3561,4 +3582,56 @@ mod archival_state_tests {
...
+                    .unwrap()
+                    .as_block_record();
+                assert_eq!(
+                    block.hash(),
+                    stored_record
+                        .pow_witness
+                        .block_hash(stored_record.block_header),
+                    "Block hash from stored witness must agree with block hash for block height {}",
+                    block.header().height
+                );
```
aszepieniec, line 3622: It took me a while to decipher this block of code. I think it is at best tangentially related to the purpose of the test as conveyed by the name. I would recommend moving it to separate test case or, if a test case with that purpose and function already exists, removing it altogether. If you insist on keeping it, please disclaim with comments that it is there for double-checking redundancy and not the main purpose of the test, so that later code readers do not stumble over it like I did.

src/models/state/mod.rs
```diff
@@ -1566,7 +1568,7 @@ impl GlobalState {
         };
 
         let tip_height = tip.header().height;
-        if tip_height < MIN_BLOCK_HEIGHT_FOR_SYNCING.into() {
+        if tip_height < (SYNC_CHALLENGE_POW_WITNESS_LENGTH as u64).into() {
```
aszepieniec, line 1571: Since `SYNC_CHALLENGE_POW_WITNESS_LENGTH` and `SYNC_CHALLENGE_NUM_BLOCK_PAIRS` are independent constants, and since sync mode only makes sense if the claimed tip height is larger than both, shouldn't we check `tip_height` against the minimum of these constants?

```suggestion
        if tip_height < (usize::min(SYNC_CHALLENGE_POW_WITNESS_LENGTH,SYNC_CHALLENGE_NUM_BLOCK_PAIRS) as u64).into() {
```

src/models/state/mod.rs
```diff
@@ -1618,6 +1620,21 @@ impl GlobalState {
...
+        let mut pow_witnesses: Vec<(BlockHeader, BlockPoWWitness)> = vec![];
+        let mut block_hash = tip.hash();
+        while pow_witnesses.len() < SYNC_CHALLENGE_POW_WITNESS_LENGTH {
+            let (block_header, pow_witness) = self
+                .chain
+                .archival_state()
+                .block_header_with_pow_witness(block_hash)
+                .await
+                .unwrap_or_else(|| {
+                    panic!("Pow-witness for block with hash {block_hash} must exist")
```
aszepieniec, line 1632: There is something missing from this statement that makes its truth obvious. Is it because archival nodes always store the proof-of-work witnesses for all blocks stored?

aszepieniec (changes requested):
I left some remarks in-line.

Besides those:

 - `BlockPoWWitness` is always (afaict; perhaps only almost-always) used in combination with `BlockHeader`. This pattern suggests that a single type that combines both pieces of information might be more appropriate. Is there a reason why you avoided that design strategy?
 - There is ambiguity about what a "proof-of-work witness" denotes. For instance, the new struct `BlockPoWWitness` does not include the header; whereas field `pow_witnesses` on `SyncChallengeResponse` does include headers. Please dispel this ambiguity through appropriate renamings and/or refactorings. On that note: is "proof-of-work witness" even the right term to capture the new data that's being encapsulated/stored/transmitted/validated? It is not a witness to proof-of-work unless for a very contrived and non-obvious relation. And its relation to proof-of-work is indirect, as evidenced by the fact that none of the new methods compute a proof-of-work number or verify one.

I think the cleanest way forward is to refactor the types, for instance by defining a new type `AuthenticatedHeader` which contains a block header and its MAST authentication path. Besides avoiding the frequent use of impromptu tuple types and the ambiguity, it would have the benefit of admitting an `Arbitrary` implementation and associated `proptest`s.

PR #352, opened 2025-01-25
refactor!: Seal trait `ConsensusProgram`

jan-ferdinand:
From the functions defined in the trait `ConsensusProgram` it feels like _only_ Neptune core should ever provide implementations of this trait. Is that correct? If so, we can [seal the trait](https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/) (like this PR suggests) and prevent downstream crates from providing additional implementations for `ConsensusProgram`.

If the trait is not even supposed to be _used_ outside of Neptune core I suggest to make it a `pub(crate) trait` instead.

### DIFF
--- a/src/models/blockchain/block/validity.rs
+++ b/src/models/blockchain/block/validity.rs
@@ -15,6 +15,7 @@ use self::correct_mutator_set_update::CorrectMutatorSetUpdate;
 use self::mmr_membership::MmrMembership;
 use self::predecessor_is_valid::PredecessorIsValid;
 use super::Block;
+use crate::models::proof_abstractions::tasm::program;
 use crate::models::proof_abstractions::tasm::program::ConsensusProgram;
 use crate::models::proof_abstractions::SecretWitness;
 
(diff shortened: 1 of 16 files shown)


aszepieniec:
> only Neptune core should ever provide implementations of this trait. Is that correct? 

> the trait is not even supposed to be used outside of Neptune core

> Is that correct? 

Unclear. I can imagine external smart contract / dapp developers working in their own repository instead of hacking on neptune-core. And with an eye to dissemination, publishing their own crate seems to make sense too. At the very least, they will need a trait that is similar to `ConsensusProgram`.

That said, right now we don't need to worry about downstream crates because there aren't any (afaik) and people haven't requested features yet. With an eye on the quality of our own code base, I would say the preferable option is to mark it `pub(crate)` instead. If/when there is demand, we can consider dropping the `(crate)` part or even sealing the trait.

jan-ferdinand:
Superseded by #353

PR #353, opened 2025-01-27
Make trait `ConsensusProgram` private

jan-ferdinand:
The fallout from marking trait `ConsensusProgram` as `pub(crate)` suggests to me that the trait functions `source` and `run_rust` don't really belong in trait `ConsensusProgram`. Perhaps a `#[cfg(test)] trait ShadowedConsensusProgram` or similar could work.

### DIFF
--- a/src/models/blockchain/block/validity.rs
+++ b/src/models/blockchain/block/validity.rs
@@ -77,6 +77,7 @@ impl SecretWitness for PrincipalBlockValidationWitness {
 }
 
 impl ConsensusProgram for PrincipalBlockValidationLogic {
+    #[cfg(test)]
     fn source(&self) {
         todo!()
     }
--- a/src/models/blockchain/block/validity.rs
+++ b/src/models/blockchain/block/validity.rs
@@ -77,11 +77,6 @@ impl SecretWitness for PrincipalBlockValidationWitness {
 }
 
 impl ConsensusProgram for PrincipalBlockValidationLogic {
-    #[cfg(test)]
-    fn source(&self) {
-        todo!()
-    }
-
     fn library_and_code(&self) -> (Library, Vec<LabelledInstruction>) {
         todo!()
     }
@@ -90,5 +85,17 @@ impl ConsensusProgram for PrincipalBlockValidationLogic {
         static HASH: OnceLock<Digest> = OnceLock::new();
 
         *HASH.get_or_init(|| self.program().hash())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::models::proof_abstractions::tasm::program::test::ConsensusProgramSpecification;
+
+    impl ConsensusProgramSpecification for PrincipalBlockValidationLogic {
+        fn source(&self) {
+            todo!()
+        }
     }
 }
--- a/src/models/blockchain/block/validity/block_program.rs
+++ b/src/models/blockchain/block/validity/block_program.rs
@@ -64,61 +64,6 @@ impl BlockProgram {
 }
 
 impl ConsensusProgram for BlockProgram {
-    #[cfg(test)]
-    fn source(&self) {
-        use tasm_lib::triton_vm::prelude::BFieldElement;
-
-        use crate::models::proof_abstractions::tasm::builtins as tasmlib;
-        use crate::models::proof_abstractions::tasm::builtins::verify_stark;
-
-        let block_body_digest: Digest = tasmlib::tasmlib_io_read_stdin___digest();
-        let start_address: BFieldElement = FIRST_NON_DETERMINISTICALLY_INITIALIZED_MEMORY_ADDRESS;
-        let block_witness: BlockProofWitness = tasmlib::decode_from_memory(start_address);
-        let claims: Vec<Claim> = block_witness.claims;
-        let proofs: Vec<Proof> = block_witness.proofs;
-
-        let block_body = &block_witness.block_body;
-
-        let txk_mast_hash: Digest = tasmlib::tasmlib_io_read_secin___digest();
-        let txk_mast_hash_as_leaf = Tip5::hash(&txk_mast_hash);
-        tasmlib::tasmlib_hashing_merkle_verify(
-            block_body_digest,
-            BlockBodyField::TransactionKernel as u32,
-            txk_mast_hash_as_leaf,
-            BlockBody::MAST_HEIGHT as u32,
-        );
-
-        // Verify fee is legal
-        let fee = &block_body.transaction_kernel.fee;
-        let fee_hash = Tip5::hash(fee);
-        tasmlib::tasmlib_hashing_merkle_verify(
-            txk_mast_hash,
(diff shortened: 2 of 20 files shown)


aszepieniec:
> The fallout from marking trait `ConsensusProgram` as `pub(crate)` suggests to me that the trait functions `source` and `run_rust` don't really belong in trait `ConsensusProgram`. Perhaps a `#[cfg(test)] trait ShadowedConsensusProgram` or similar could work.

For context: the reason why the function is called "source" is two-fold, and historical in both cases:
 1. The source is the starting point for a compilation process that results in code. We did use the tasm-lang compiler at some point, but even making the translation manually qualifies as compilation.
 2. The source *defines* consensus (or at least that was my perspective) in the sense that *if* there is an inconsistency between the source and the tasm code then the tasm code is "wrong". And the tasm code merely *implements* consensus and moreover it does that only if it is consistent with the source.

Based on that context, I am of the opinion that "`source`" is a first class citizen, and that the orange squiggly lines merely indicate a failure to recognize the dependency of the tasm code on the source.

None of that applies to "`run_rust`" which probably can be marked with a test flag.

aszepieniec:
Instead of deploying an even bigger refactor, I would suggest to merely suppress the linter complaint warning that `source` is unused. It is used in tests but more importantly it is used in the course of a compilation pipeline that we can't expect rust to pick up on.

Or maybe I just don't understand your question :man_shrugging:

jan-ferdinand:
I can see the perspective that the rust `source` is the specification, the tasm `program` its implementation. I draw different conclusions from it.

Like most specifications, the `source` is never executed. Therefore, like other specifications, it should not reside in the (main part of the) source code. Usually, specifications go into a separate document alltogether. In this case, the specification _can_ be executed on the rust abstract machine, and we can use that advantage by spot-checking semantical equivalence in tests.

aszepieniec:
That summary is concise and accurate. What do you suggest?

aszepieniec:
Are you making this suggestion for our consideration to guard against the possibility that we did not think of this option when writing trait `ConsensusProgram` (you probably guessed right) without an implied follow-up action, or are you making an argument for adopting this change (modulo possible changes to better fit my mental model)? If the latter, please elaborate why you prefer this option to the alternatives.

From my perspective:
 - Two traits is more complicated than one.
 - Traits should not exist for the purpose of testing. On this point I anticipate the counterargument that `TestableConsensusProgram` *also* serves the purpose of specification and to this I rebut that the name and location fail to communicate this dual purpose. Naming it `ConsensusProgramSpecification` reduces the number of rebuttal points to one.
 - Traits unnecessarily marked `pub` or linter warning suppressions are not the end of the world. Orange squiggly lines are bad because those lead to failing pre-commit hooks, but then we have linter warning suppressions for that.

> From my perspective:

I phrase it this way because I want to hear about yours.

jan-ferdinand:
I do in fact suggest to adopt this PR, or something like it.

- Mixing test code and non-test code increases code complexity. “Managing complexity is the most important technical topic in software development.”[^cc_complex] To be more precise, the mixing obscures. Some identify obscurity as one of the two main causes of complexity.[^philo_cause] This mixing constitutes _accidental_ complexity, not _essential_ complexity.[^bullet] It can, and thus should, be avoided.
- I'm convinced that splitting traits (thus increasing the number of traits) can reduce the complexity of the program. In particular, this is the case when one trait has multiple purposes with little overlap: creating one trait per purpose reduces overall complexity. What “one purpose” means is different from case to case, and might depend on perspective. Ultimately, this is a question of design.
- I don't agree with the statement that traits “should not exist for the purpose of testing [alone].”[^test_trait] Good software engineering principles apply as much in test code as they apply outside of it. If a trait is the right tool for the job, then it should be used.
- With very few exceptions, lints highlight issues in the code. I'd much rather resolve such issues than to acknowledge their presence and do nothing about them. Furthermore, suppressing lints sets precedence and can be habit-forming, reducing their usefulness even more. This contributes to worse code or more bad code down the line.
- Suppressing lints in order to satisfy the (pre-commit or other) git hooks or to pass the checks of continuous integration defeats the purpose of either. If hooks and CI are in the way, it's better to be honest about it and remove the hooks and to disable CI.
- As mentioned, I'm also not happy with the trait's name “`TestableConsensusProgram`”. I think your suggestion is an improvement.

Lastly, I don't want to suggest that this PR has the _right_ solution. Probably, better designs exist. I am of the opinion that my suggestion reduces overall code complexity.

[^cc_complex]: Code Complete, 2nd Edition, section “Key Design Concepts”
[^philo_cause]: A Philosophy on Software Design, 2nd Edition, section 2.3 “Causes of complexity”
[^bullet]: https://www.cs.dartmouth.edu/~cs50/Reading/NoSilverBullet.pdf
[^test_trait]:  You seem to have had a different stance at some point, evidenced by the trait ecosystem (!) around the `RustShadow` trait defined in tasm-lib. Granted, that trait is not under the `test` configuration, but its purpose is still very much testing.

aszepieniec:
Thanks for the elaborate counter-argument. Gives me something to think about.

I certainly acknowledge the point that one trait with two purposes is not less complex than two traits with one purpose each. And in fact, after phrasing it in those terms, I prefer the second as well. So let's proceed with this PR (after `TestableConsensusProgram` $\mapsto$ `ConsensusProgramSpecification`) and with the understanding that it is possible for better designs still to present themselves and challenge the incumbent.

In the interest of clarifying the record ...

> You seem to have had a different stance at some point, evidenced by the trait ecosystem (!) around the RustShadow trait defined in tasm-lib. Granted, that trait is not under the test configuration, but its purpose is still very much testing. [↩](https://github.com/Neptune-Crypto/neptune-core/pull/353#user-content-fnref-test_trait-bdeda991ef83881027342f2fa00a082e)

My position was formed in response to struggling with the monstrosity you cite. I do not necessarily think that is a good design; I do think it is an improvement over what was there before.

For context, I think this commandment ("traits should not exist for the purpose of testing") constrains "good" tests more than "good" traits. Traits, or interfaces more generally, serve to minimize the contact surface between separate bodies of code. They should delineate conceptual boundaries. Tests should probe interfaces, not internal mechanics. If you find yourself writing an interface for tests, you're missing the point and you should reconsider your interface architecture.

I realize that these statements are debatable but the point is that they fit together nicely (or so I think!) and between other positions I hold.

jan-ferdinand:
> with the understanding that it is possible for better designs still to present themselves and challenge the incumbent

Absolutely. To quote Code Complete some more:
> [Designs] evolve and improve through design reviews, informal discussions, experience writing the code itself, and experience revising the code.

> My position was formed in response to struggling with the monstrosity you cite. I do not necessarily think that is a good design; I do think it is an improvement over what was there before.

Acknowledged and recorded. Adding to the record, thanks for taking the time to improve it; it is in a workable state now and I don't see easy ways to improve it further.

jan-ferdinand:
I have renamed the trait to `ConsensusProgramSpecification` and rebased on top of the `master` branch. If there are no objections, I'll merge once CI gives the all clear.

PR #354, opened 2025-01-28
Shamir Secret Sharing

aszepieniec:
I think it is rather uncontroversial but it can't hurt to offer the opportunity for review.


docs/src/user-guides/shamir-secret-sharing.md
```diff
@@ -0,0 +1,155 @@
...
+## How to Use It
+
+First, make sure you have a wallet installed.
+
+ - Whenever you run `neptune-core`, it will read the wallet file or create one if none is found. Unless you moved or removed this file, it is still there.
+ - To test if the wallet file is present, run `neptune-cli which-wallet`.
+ - To generate a wallet file without running `neptune-core`, try `neptune-cli generate-wallet`.
+ - To import a wallet from a seed phrase, first make sure there is no wallet file, and then run `neptune-cli import-seed-phrase`.
+
+To generate \\(n\\) shares in a \\(t\\)-out-of-\\(n\\) scheme, run `neptune-cli shamir share t n` and replace `t` and `n` with the values you want. This command generates \\(n\\) seed phrases. **Note:** be sure to record the share index ("`i/n`") along with each share, as you will need this information to reconstruct the original secret.
```
Sword-Smith, line 26: `neptune-cli shamir share t n` should be `neptune-cli shamir-share t n`  (there's a dash `-` missing).

src/bin/neptune-cli.rs
```diff
@@ -497,6 +496,190 @@ async fn main() -> Result<()> {
...
+                    wallet_file.display()
+                );
+                return Ok(());
+            }
+
+            // prompt user for all shares
+            let mut shares = vec![];
+            let capture_integers = Regex::new(r"^(\d+)\/(\d+)$").unwrap();
+            while shares.len() != *t {
+                println!("Enter share index (\"i/n\"): ");
```
Sword-Smith, line 517: You probably intended to use `format!` here, to write the values of `i` and `n`, no?

src/models/state/wallet/mod.rs
```diff
@@ -49,14 +53,153 @@ pub const WALLET_DB_NAME: &str = "wallet";
 pub const WALLET_OUTPUT_COUNT_DB_NAME: &str = "wallout_output_count_db";
 
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
-struct SecretKeyMaterial(XFieldElement);
+pub struct SecretKeyMaterial(XFieldElement);
```
Sword-Smith, line 56: Can we use `pub(crate)` here for `SecretKeyMaterial` and all its new associated functions?

Sword-Smith (changes requested):
Nice tests!

Consider factoring the new functionality into a new file/module, as we suffer from quite big files.

Can we use `pub(crate)` instead of `pub` for the new functions and for `SecretKeyMaterial`?

Also: I identified one typo.

src/models/state/wallet/mod.rs
aszepieniec: The problem is that they are being used from within `neptune-cli.rs`, which lives in its own crate and therefore does not have access to structs and functions marked `pub(crate)`.

docs/src/user-guides/shamir-secret-sharing.md
aszepieniec: :+1:

PR #355, opened 2025-01-28
refactor(Block): Require guesser gets at least 1/4 of block subsidy

Sword-Smith:
Require that guesser gets at least 1/4 of block subsidy.

This change incentivizes the rational guesser to know the preimage of the nonce. If this rule was not there, a mining venture could set the guesser fraction to 0 leaving the entire block subsidy for the composer. With this setup, the composer could skip one invocation of Tip5 since the nonce would not have to be the output of Tip5 but could instead be an arbitrary number. This would give centralized mining ventures and unfair advantage and thus disincentivize Neptune's two-step mining. The hash rate advantage of this approach would be about 15 %.

This change also makes it easier for composers to pick up multiple fee-paying transactions and merge them into an existing block proposal, thus increasing the guesser reward. If mining venture does not bother with using a Tip5-digest for the nonce, they will not have an incentive to pick up and merge fee-paying transactions while PoW guessing is still ongoing. Or, to be more precise, it'll be twice as expensive computationally to pick up fee-paying transactions, as the fee cannot simply be passed onto the transaction fee but must be merged with a transaction with a negative fee.

This change *also* protects against a deterioration of the anonymity set that would happen if the guesser-reward was zero, as those UTXOs would be unlikely to ever be spent.

### DIFF
--- a/src/mine_loop.rs
+++ b/src/mine_loop.rs
@@ -1160,7 +1160,9 @@ pub(crate) mod mine_loop_tests {
             .unwrap();
 
         let mut cli = cli_args::Args::default();
-        for guesser_fee_fraction in [0f64, 0.5, 1.0] {
+        for guesser_fee_fraction in [0.25f64, 0.5, 1.0] {
+            println!("guesser_fee_fraction: {guesser_fee_fraction}");
+
             // Verify constructed coinbase transaction and block template when mempool is empty
             assert!(
                 alice.lock_guard().await.mempool.is_empty(),
--- a/src/models/blockchain/block/mod.rs
+++ b/src/models/blockchain/block/mod.rs
@@ -668,8 +668,9 @@ impl Block {
         //   c) Verify that the mutator set update induced by the block sends
         //      the old mutator set accumulator to the new one.
         //   d) transaction timestamp <= block timestamp
-        //   e) transaction coinbase <= miner reward, and not negative.
-        //   f) 0 <= transaction fee (also checked in block program).
+        //   e) 0 <= transaction fee (also checked in block program).
+        //   f) transaction coinbase <= miner reward, and not negative.
+        //   g) Coinbase, if present, is max 4x the transaction fee.
 
         // DO NOT add explanations to the individual steps below. *All*
         // explanations belong in the recipe above and may not be duplicated or
@@ -844,9 +845,15 @@ impl Block {
         }
 
         // 2.e)
+        let fee = self.kernel.body.transaction_kernel.fee;
+        if fee.is_negative() {
+            warn!("Fee may not be negative when transaction is included in block. Got fee: {fee}.");
+            return false;
+        }
+
+        // 2.f)
         let block_subsidy = Self::block_subsidy(self.kernel.header.height);
-        let coinbase = self.kernel.body.transaction_kernel.coinbase;
-        if let Some(coinbase) = coinbase {
+        if let Some(coinbase) = self.kernel.body.transaction_kernel.coinbase {
             if coinbase > block_subsidy {
                 warn!(
                     "Coinbase exceeds block subsidy. coinbase: {coinbase}; \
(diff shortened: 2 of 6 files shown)


aszepieniec:
Summarizing a discussion we had at the office just now:

*Mechanics of problem.* Right now one guess iteration involves 7 hashing steps. One of them is to convert the nonce preimage into the nonce. The guesser needs to record the preimage because that is the value that will allow him to unlock the guesser fee if the nonce is good enough and the block becomes accepted by the network. Guessers can get away with only 6 hashing steps by sampling a random nonce instead of hashing a nonce-preimage. They won't be able to spend the guesser fees, but to compensate for this they can absorb all rewards while composing the block and leave 0 coins to the guesser. The hash power advantage is of this strategy over the standard one is `(1/6)/(1/7) - 1 = 7/6 - 1 = 0.167`.

*Consequences.*

 1. Miner centralization.
   - a) Mining companies need to reach a certain scale to justify the relative cost of producing this non-standard block composition. After reaching this scale, they can benefit from the 17% hash power advantage. Phrased differently, miners who do not have this scale pay with a 14% disadvantage.
   - b) As an individual guesser, I can choose to direct my guessing power to broadcasted block proposals, or I can choose to direct it to a mining pool that uses this 6-step strategy. Assuming i) that the mining pool operator behaves integrally, and ii) that it takes a small enough fee; then I stand to benefit by directing my guessing power to the mining pool.
 2. Ignore fee-paying transactions. Miners employing this strategy do not stand to benefit from the fee paid by users for priority access to block space -- so why bother including their transactions? Alternatively, they can a) merge the fee-paying transaction with a negative-fee transaction (so as to collect the fee), and then b) merge that into their block transaction. With this strategy that *can* benefit from the fee paid by a transaction but only after incurring the cost of one extra merge step (relative to the conventional strategy). The end result is that the fee needs to be higher so as to exceed the cost of this second merge step. (Bear in mind that electricity and hardware is only one factor, but *time* is another factor especially relevant until we have GPU/ASIC proving.)
 3. Shrinkage of the anonymity set. Guesser fee UTXOs that have zero value are unlikely to be spent. But they still contribute towards shifts of the sliding window Bloom filter in the mutator set. So other UTXOs that are created around the same time have fewer decoys.

*Proposed Solution.* The proposed solution modifies the block program to require that the guesser fee is at least 25% of the block subsidy. Miners can still use the 6-step strategy, but incur a *disadvantage* if they do, at least while the block subsidy dominates the transaction fee.

 - The proposed solution is rather small in scope (see changed files).
 - The proposed solution modifies consensus rules (although it's difficult to see how *any* solution can avoid changing them).
 - The proposed solution is *only* effective while the block subsidy dominates the transaction fees. After 7 or 8 halvings we expect transactions fees to dominate. The proposed solution does not address the dynamics in this regime.
   - Counter: at this point, the problem does not exist anymore. The miner can only afford to ignore transaction fees because of the dominating block subsidy. In the regime where block subsidies are insignificant, the miner cannot afford to ignore transaction fees.

*Alternatives.*

 1. Change the proof-of-work number associated with a block whose transaction fee is zero. For instance, multiply it by 0.85 when this condition is met.
   - Surgical in the sense that addresses the discrepancy close to where it originates from, as opposed to a later place in the pipeline.
   - Rather more complex than the solution proposed in this PR. For instance, right now to get the proof-of-work number you just need the block hash; but with this change you need the block hash and the transaction fee, which lives in the transaction kernel, which lives in the block body.
   - Tinkers with a very fragile thing, which is whatever defines which tips are canonical.

aszepieniec (commented):
I reviewed the code and the changes look adequate to me.

Lets sleep on this PR for a few days. I do not think changes to consensus rules should be taken lightly.

aszepieniec:
Here is another solution I just came up with.

We extend the `BlockHeader` data structure with one more field: `guesser_digest: Digest`. The guesser sets this value along with the `nonce`, but records only the preimage of `guesser_digest`. The `nonce` field has no preimage any more. Instead, the generated guesser fee UTXOs are locked under a `RawHashLock` that can be opened with the preimage to the `guesser_digest`. The `receiver_digest` is set to `guesser_digest` so that the guesser gets privacy when/if he spends the UTXO. However, the `sender_randomness` is set to the `nonce`.

Pros/cons/food-for-thought:
 - Substantial change.
 - Comprehensive solution; in near and far future, 0% guess rate disparity.
 - Surgical: the source of the disparity -- an optional step in a tight loop -- has been addressed directly, namely by moving it outside of the tight loop.
 - One of the benefits of the current solution is that the exact form of the mutator set accumulator after applying a block cannot be known until the guessing puzzle has been solved. As a result, it makes no sense for composers to start composing before the winning guess has been found. This new solution inherits the same feature.

Sword-Smith:
Closed in favor of 0ba5890f57be9cb8b39dad1e548daaaf56c672c4

Issue #339, opened 2025-01-20
Add mdbook docs for RPC API and improve doc-comments.


dan-da:
Note that `cargo doc` already produces basic docs for the RPC API.  The doc-comments should be improved to provide more of an overview of the system and an example for each method.

Also we should add a description of the RPC system in the mdbook.  

Here are some examples from bitcoin-core:

* https://bitcoincoredocs.com/JSON-RPC-interface.html
* https://developer.bitcoin.org/reference/rpc/index.html

elielnfinic:
Can I work on this?

aszepieniec:
> Can I work on this?

Yes, please!

elielnfinic:
Thank you.

dan-da:
@elielnfinic thx for taking this on.    I suggest splitting it into 2 separate PRs:

1. adding doc-comment example(s) to each  RPC method, in the RPC trait in src/rpc_server.rs.
2. adding an mdbook file at docs/src/neptune-core/rpc.md

It doesn't matter which PR comes first, but you might find (2) easier to write after doing (1).

With regards to (2), I think we could use https://bitcoincoredocs.com/JSON-RPC-interface.html as a starting template.

Some key differences are:

1: We use tarpc (rust crate) with the serde_json transport, rather than json-rpc.
2. Because of (1) for now it is will be easiest to write clients in rust.    This can be spun as a good thing, because rust tends to produce quite reliable code, and it will be beneficial to build up a neptune rust ecosystem of library crates and clients.  It should be possible to write a client in some other language as everything is json serialized, but there is presently no code to do so, so that's something we are hoping 3rd parties may build some libs for that in their language of choice.
3. We presently support only cookie auth, not basic user/pass, or digest.  (Digest should be coming soon-ish).
4. We provide a cookie-hint RPC so client can find data-directory in a zero-conf way.  And a flag to disable it.   If disabled, client's should fallback to default location, or user-provided location.  See neptune-cli.rs for an example.
5. error handling.  Clients encounter an outer tarpc result, and an inner method result, with distinct error types.

Mostly this is explained in doc-comments in rpc_server.rs and rpc_auth.rs.   So much of that can be copied / adapted for the mdbook page.

I'm happy to answer q's if anything is unclear.

dan-da:
oh, and just in case it is not obvious:

RPC docs can be generated via `cargo doc` in the root.
[...]

PR #358, opened 2025-02-02
Improving RPC doc comments - addresses #339 

elielnfinic:
Improving RPC doc comments - addresses #339


docs/src/SUMMARY.md
```diff
@@ -14,6 +14,7 @@
   - [Reorganization](./neptune-core/reorganization.md)
   - [Keys and Addresses](./neptune-core/keys.md)
   - [Utxo Notification](./neptune-core/utxo_notification.md)
+  - [RPC Server](./neptune-core/rpc.md)
```
aszepieniec, line 17: I don't see a `rpc.md` file. Is that coming in an upcoming commit?

src/rpc_server.rs
```diff
@@ -204,150 +204,944 @@ pub trait RPC {
...
+    /// # tokio_test::block_on(async {
+    ///
+    /// // create a serde/json transport over tcp.
+    /// let transport = tcp::connect("127.0.0.1:9799", Json::default).await.unwrap();
+    ///
+    /// // create an rpc client using the transport.
+    /// let client = RPCClient::new(client::Config::default(), transport).spawn();
+    ///
+    /// // load the cookie file from disk and assign it to a token
+    /// let token : rpc_auth::Token = rpc_auth::Cookie::try_load(&cookie_hint.data_directory).await.unwrap().into();
```
dan-da, line 279: I don't see `cookie_hint` defined.   Does this actually compile?   I think you are missing a call like:

```
    let cookie_hint = client.cookie_hint(context::current()).await??;
```

anyway, let's comment out all the setup stuff, so only the RPC call displays in the generated docs.

also, please replace the unwrap() with `?` if possible.

dan-da (changes requested):
Thanks for the PR @elielnfinic, it's a good start.

Please make the changes I suggested in the comments and ensure that `cargo test --doc` executes cleanly without warnings/errors.

Also, please squash all the commits into one.
https://www.baeldung.com/ops/git-squash-commits

with that, I think we will be ready to merge it.

let me know if any q's or concerns.

src/rpc_server.rs
```diff
@@ -204,150 +204,944 @@ pub trait RPC {
     /// The CookieHint provides a location for the cookie file used by this
     /// neptune-core instance as well as the [Network].
     ///
+    /// ```no_run
```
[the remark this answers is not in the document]
dan-da: also, can we use `??` instead of `.unwrap().unwrap()`?

src/rpc_server.rs
```diff
@@ -14,24 +14,24 @@
 //! use tarpc::client;
 //! use tarpc::context;
 //!
-//! # tokio_test::block_on(async {
-//!
+//! # #[tokio::main]
+//! # async fn main() -> Result<(), Box<dyn std::error::Error>>{
```
dan-da, line 18: are you using main() instead of tokio_test::block_on() in order to return errors, or for some other reason?    If so, that seems reasonable, but let's just use anyhow::Result<()>, which is already in scope as `Result`.   So it should work to just:

```suggestion
//! # async fn main() -> Result<()> {
```

dan-da (changes requested):
looks good overall.   let's just change the Result type to Result<()>.

Seems we will also need one more rebase to master since some conflicting changes have already snuck in.

PR #366, opened 2025-02-05
trace(ProverJob): More info about job complexity

Sword-Smith:
The padded height is shown before a proving job starts, but it's only the padded height from the processor table (which often but not always dominates). With this PR, the actual proof complexity (dominating padded height) is printed once the job is done, as the proof (if well-formed) will contain this data.

Logging the proving speed would also make a lot of sense, dividing the (real) padded height with the time it took to produce the proof. That's not included in this PR, as I couldn't find a place in the code in the pipeline where both the time and the proof were in scope.

### DIFF
--- a/src/models/proof_abstractions/tasm/prover_job.rs
+++ b/src/models/proof_abstractions/tasm/prover_job.rs
@@ -200,25 +200,25 @@ impl ProverJob {
         assert_eq!(self.claim.program_digest, self.program.hash());
         assert_eq!(self.claim.output, vm_state.public_output);
 
-        let padded_height = vm_state.cycle_count.next_power_of_two();
+        let padded_height_processor_table = vm_state.cycle_count.next_power_of_two();
 
         tracing::info!(
-            "VM program execution finished: padded-height: {}",
-            padded_height
+            "VM program execution finished: padded-height (processor table): {}",
+            padded_height_processor_table
         );
 
         match self.job_settings.max_log2_padded_height_for_proofs {
-            Some(limit) if 2u32.pow(limit.into()) < padded_height => {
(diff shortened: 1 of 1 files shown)


src/models/proof_abstractions/tasm/prover_job.rs
```diff
@@ -14,6 +14,8 @@ use std::process::Stdio;
 use tasm_lib::maybe_write_debuggable_vm_state_to_disk;
 #[cfg(not(test))]
 use tokio::io::AsyncWriteExt;
+#[cfg(not(test))]
+use tracing::info;
```
dan-da, line 18: style: maybe just call `tracing::info!("...")` in the code, and we can avoid this use statement ugly `#[cfg(not(test))]` here.

src/models/proof_abstractions/tasm/prover_job.rs
```diff
@@ -326,7 +328,13 @@ impl ProverJob {
                 let output = result?;
                 match output.status.code() {
                     Some(0) => {
-                        let proof = bincode::deserialize(&output.stdout)?;
+                        let proof: Proof = bincode::deserialize(&output.stdout)?;
+                        info!(
+                            "Generated proof, with padded height: {}",
+                            proof.padded_height()
+                                .map(|x| x.to_string())
+                                .unwrap_or("could not get padded height from proof.".to_string())
```
dan-da, line 336: style: let's prefer .unwrap_or_else(), eg:

```suggestion
                                .unwrap_or_else(|| "could not get padded height from proof.".to_string())
```

That way the code in the closure does not execute for the happy path.

dan-da (approved):
lgtm.   I just had a couple style suggestions.

PR #375, opened 2025-02-07
fix: remove dashboard balance discrepancy

### LINKED ISSUE #346: Dashboard balance discrepancy between overview and history screens.
After mining 8 blocks, I see in dashboard overview:

available balance: 512
unconfrmed balance: 512
time-locked balance: 512

In the history screen I see only the column:
balance after: 1024

This is pretty confusing initially, as 1024 != 512.  And there are many rows with balance > 512, so nothing seems to match up.

Presumably, the history screen "balance after" column is adding available_balance + time-locked balance to obtain 1024.

Perhaps it should have separate columns to match the fields in overview ie: available balance, unconfirmed balance,  time-locked balance.

For space reasons, it might make sense to shorten these column labels, eg:  balance (a), balance (u), balance (t).   And then on a separate legend line:

a = available, u = uconfirmed, t = time-locked.

This would require altering the history RPC method as well, as it returns only a single balance field per row.

Image
Image

dan-da:
fixes #346

This commit refactors how balances are calculated and displayed. It normalizes balance related functions to use the terminology:  confirmed/unconfirmed and available/total.

Balances are displayed in the dashboard overview as 4 fields in 2 rows:
 confirmed balance:    total, available
 unconfirmed balance:  total, available

The unconfirmed balance row is only displayed if either of the two fields differs from its counterpart in the confirmed balance row.  This keeps the default display simpler.

The balance discrepancy with the history screen is fixed because history's "balance after" column displays a total amount (including timelocked utxos) which corresponds to the confirmed balance (total) field.  This PR does not modify the history screen at all, but a follow-up could improve it as well.

In neptune-core, the WalletStatus type is refactored a bit to remove unused fields and methods.  The unsynced field is only used in tests, so is now decorated with #[cfg(test]).

In WalletState, the balance getting functions now take &WalletStatus which avoids duplicate work (iterating over all monitored utxos) by callers that call multiple of these functions, in particular the dashboard_overview_data() RPC.

neptune-core changes:
 * WalletState::confirmed_balance() --> confirmed_available_balance()
 * add WalletState::confirmed_total_balance()
 * WalletState::unconfirmed_balance() --> unconfirmed_available_balance()
 * add WalletState::unconfirmed_total_balance()
 * walletState::get_wallet_status_from_lock() --> get_wallet_status()
 * added a big doc-comment describing WalletStatus
 * WalletStatus::unsynced is now #[cfg(test)]
 * WalletStatus::synced_unspent_liquid_amount() -> synced_unspent_available_amount()
 * add WalletStatus::synced_unspent_total_amount()
 * remove unused WalletStatus::unsynced_amount()
 * remove unused WalletStatus::synced_spent_amount()
 * remove unused impl Display for WalletStatus
 * rpc confirmation_Internal() is now 1 based instead of 0 based. because block that mines tx is 1st confirmation.
 * update tests

dashboard changes:
 * overview screen displays 4 balance fields instead of 3 but in a more compact and readable representation.
 * balance fields have a width of 17 digits. (8 whole coin, 1 decimal point, 8 decimals)
 * "(x confirmations)" --> "(1 confirmation)" (plural aware)
 * "(x confirmations) now starts at 1. block that mines tx is 1st confirmation.
 * only display unconfirmed balance row if needed.

neptune-cli changes:
 * command synced-balance is now confirmed-available-balance
 * command synced-balance-unconfirmed is now unconfirmed-available-balance

(for consistency with naming throughout codebase)


Sword-Smith:
The command `Command::WalletStatus` does not use the Display, but would show the unsynced balance, as it json-encodes `WalletStatus`. With this PR, that information is lost, and any unsynced monitored UTXOs would fail to show up.

So I would argue that we keep the `unsynced` field on the `WalletStatus` struct as it's being used by the `WalletStatus` command. And we're almost certainly going to use this field for future commands/future ways of presenting the wallet status. I could, e.g., imagine that a warning or asterisk was shown next to balance amounts if the `unsynced` field was non-empty.

dan-da:
> we keep the `unsynced` field

addressed this in 064c4c7 and rebased to master.    merging.

PR #378, opened 2025-02-08
chore: normalize --port arg between rpc clients

### LINKED ISSUE #376: align neptune-dashboard and neptune-cli args for specifying server rpc connection
consider that presently we have:

```
$ neptune-dashboard --help | grep server

      --port <PORT>          Sets the server address to connect to [default: 9799]
```

```
$ neptune-cli --help | grep server
      --server-addr <SERVER_ADDR>  Sets the server address to connect to [default: 127.0.0.1:9799]
```

```
$ neptune-explorer --help
      --neptune-rpc-port <port>
          Sets the neptune-core rpc server address to connect to
          
          [default: 9799]
```

I think that all of these should ideally be the same.  Most especially neptune-dashboard and neptune-cli should match eachother.

Considerations:

Since the server only accepts rpc connections on 127.0.0.1 interface, there doesn't seem much point in requiring user to specify 127.0.0.1 as neptune-cli does.

the --neptune-rpc-port used by neptune-explorer is longer, but more descriptive than --port.   For neptune-explorer it makes sense to disambiguate --neptune-rpc-port from --smtp-port, another "port" arg.   That need likely will never arise for neptune-dashboard and neptune-cli.  Also these are much more frequently used by end-users, so the convenience of shorter --port might outweigh the goal to keep the arg the same for all 3 executables.  Additionally, a short flag, -p can be helpful.

So I propose to leave neptune-explorer unchanged and change neptune-cli and neptune-dashboard to:

```
  -p, --port <port>          Sets the neptune-core rpc server localhost port to connect to [default: 9799]
```

created by clap, from:

```
    /// Sets the neptune-core rpc server localhost port to connect to.
    #[clap(short, long, default_value = "9799", value_name = "port")]
    port: u16,
```

If we ever decide we need to support other than 127.0.0.1, we could add a --host arg for each executable.

dan-da:
closes #376.   (see for rationale)

modifies neptune-cli and neptune-dashboard to both use the same argument for specifying neptune-core rpc port:

```
-p, --port <port>          Sets the neptune-core rpc server localhost port to connect to [default: 9799]
```

also updates README and mdbook docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,10 +63,10 @@ neptune-cli --help
 
 To get e.g. the block height of a running daemon, execute
 ```
-neptune-cli --server-addr block-height
+neptune-cli block-height
 ```
 
-If you set up `neptune-core` on a different address or port from the default (127.0.0.1:9799), then the flag `--server-addr [ip_address:port]` is your friend.
+If you set up `neptune-core` to listen for RPC requests on a different port from the default (9799), then the flag `--port <port>` is your friend.
 
 ## Setup for Development (Ubuntu)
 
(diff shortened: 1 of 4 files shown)


Sword-Smith:
(Alan also)

> LGTM

PR #384, opened 2025-02-09
Improving RPC inline documentation

elielnfinic:
Improving RPC inline documentation - addresses #339


dan-da:
looks good to me.   thanks for the PR @elielnfinic.

note: obsoletes #358.

PR #385, opened 2025-02-09
trace: Reduce severity of duration-check log message

Sword-Smith:
When running the client, I want WARN log messages to catch my attention. But these warnings from the duration-checker completely spam my log output so I don't notice other warnings.

With this change, the messages are still printed but only with the DEBUG severity.


dan-da (approved):
yeah that's fine. 

We'll probably want to remove/refine some of these anyway as we go forward.

PR #400, opened 2025-02-11
fix: import-seed-phrase check if wallet db exists

### LINKED ISSUE #398: neptune-cli import-seed-phrase allows modifying secret although wallet db exists.
I was just reviewing neptune-cli code.  I haven't tested this, so I'm uncertain what issues will arise.

What I notice is that the code for import-seed-phrase checks for existence of wallet-secret file and only allows to proceed if it does not exist.  that's fine.

On the happy path, it creates a new wallet-secret file.   But it never checks for existence of wallet-db file.   So a new wallet-secret can be created (by our tooling) along-side an old wallet-db.

Is that a problem?   I'm not certain.  But I feel like it will be, because the wallet-db contains monitored-utxos, spent-transactions (if #397 gets merged) and other data that does not correspond to the keys/addresses generated from the wallet-secret.  So the wallet will likely present wrong information.  It might even encounter some unexpected state and crash... hard to say until we test this scenario.

It certainly would be possible for the neptune-core code to detect this situation and rebuild the wallet-db (as best as possible) by reading the entire blockchain, but afaik that code does not exist.   (it probably should in the future, eg in case a user manually does this, or loses wallet db).

I think the simplest fix right now would be to modify the import-seed-phrase command to also check for existence of the wallet-db file, and error out if it exists.

dan-da:
fixes #398

Checks if wallet-dir or wallet-db-dir exists before importing seed phrase.  If either exist, an error message is printed and the action aborted.

The old behavior was to check only if the wallet-secret file exists.

------

note: the wallet-dir contains file incoming_randomness.dat  outgoing_randomness.dat  wallet.dat.   I'm unsure the implications of replacing wallet.dat (the wallet secret) and keeping the other 2 files.    For this reason, I (errored?) on the safe side to just error out if the wallet-dir exists.

### DIFF
--- a/src/bin/neptune-cli.rs
+++ b/src/bin/neptune-cli.rs
@@ -429,15 +429,23 @@ async fn main() -> Result<()> {
             return Ok(());
         }
         Command::ImportSeedPhrase { network } => {
-            let wallet_dir =
-                DataDirectory::get(args.data_dir.clone(), *network)?.wallet_directory_path();
-            let wallet_file = WalletSecret::wallet_secret_path(&wallet_dir);
+            let data_directory = DataDirectory::get(args.data_dir.clone(), *network)?;
+            let wallet_dir = data_directory.wallet_directory_path();
+            let wallet_db_dir = data_directory.wallet_database_dir_path();
+            let wallet_secret_path = WalletSecret::wallet_secret_path(&wallet_dir);
 
-            // if the wallet file already exists,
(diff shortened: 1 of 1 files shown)


aszepieniec (approved):
LGTM but let me add it to the queue for mergers-into-master

aszepieniec:
Merged in 09c3a2ca48702b124996b204380115c57c91673e.

PR #402, opened 2025-02-11
doc : updating synchronizing to synchronized

### LINKED ISSUE #392: Dashboard: Rename "synchronizing" to "sync mode"
It is rather easy to misread "synchronizing: false" as "synchronized: false" and come to the conclusion that one's node is synchronized already. The interface can be improved if it instead says "sync mode: false". More difficult to interpret that wrongly.

elielnfinic:
Fixes #392

### DIFF
--- a/src/bin/dashboard_src/overview_screen.rs
+++ b/src/bin/dashboard_src/overview_screen.rs
@@ -365,7 +365,7 @@ impl Widget for OverviewScreen {
 
         lines.push(format!("network: {}", data.network));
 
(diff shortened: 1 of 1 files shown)


aszepieniec:
I'm afraid you got the meaning of the variable wrong, and fell into the very trap the issue was supposed to disambiguate.

`sync_mode` denotes whether the node is currently synchronizing. Probably better to rename the UI label to "sync mode".

elielnfinic:
Lol, thanks @aszepieniec

PR #429, opened 2025-02-13
Refuse connections from recent disconnects

jan-ferdinand:
In order to reduce the potential connection flood, rapid re-connection attempts after a disconnect are ignored.

If node A decides to (gracefully) terminate the connection to some node B, then node A blocks any connection attempts from node B for some time. The cool-down time can be configured through the command-line interface's argument “reconnect_cooldown”. The argument is given in seconds. The default is 30 seconds.

~~This introduces a breaking change because the public enum `ConnectionRefusedReason` gets a new variant `RecentlyDisconnected`. This enum is now marked `#[non_exhaustive]` to prevent future variant additions from being breaking again.~~

This introduces a breaking change because the public structure `NetworkingState` gains a new, private field.

### DIFF
--- a/src/config_models/cli_args.rs
+++ b/src/config_models/cli_args.rs
@@ -260,6 +260,18 @@ pub struct Args {
     /// Exposing the data directory leaks some privacy. Disable to prevent.
     #[clap(long)]
     pub disable_cookie_hint: bool,
+
+    /// The duration (in seconds) during which new connection attempts from peers
+    /// are ignored after a connection to them was closed.
+    ///
+    /// Does not affect abnormally closed connections. For example, if a connection
+    /// is dropped due to networking issues, an immediate reconnection attempt is
+    /// not affected by this cooldown.
+    //
+    // The default should be larger than the default interval between peer discovery
+    // to meaningfully suppress rapid reconnection attempts.
+    #[clap(long, default_value = "1800", value_parser = duration_from_seconds_str)]
+    pub reconnect_cooldown: Duration,
 }
 
 impl Default for Args {
@@ -278,6 +290,10 @@ fn fraction_validator(s: &str) -> Result<f64, String> {
     } else {
         Err(format!("Fraction must be between 0 and 1, got {value}"))
     }
+}
+
+fn duration_from_seconds_str(s: &str) -> Result<Duration, std::num::ParseIntError> {
+    Ok(Duration::from_secs(s.parse()?))
 }
 
 impl Args {
--- a/src/connect_to_peers.rs
+++ b/src/connect_to_peers.rs
@@ -1,5 +1,6 @@
 use std::fmt::Debug;
 use std::net::SocketAddr;
+use std::time::SystemTime;
 
 use anyhow::bail;
 use anyhow::Result;
@@ -106,6 +107,27 @@ async fn check_if_connection_is_allowed(
         let ip = peer_address.ip();
         warn!("Peer {ip}, banned because of bad standing, attempted to connect. Disallowing.");
         return InternalConnectionStatus::Refused(ConnectionRefusedReason::BadStanding);
+    }
+
+    if let Some(time) = global_state
+        .net
+        .last_disconnection_time_of_peer(other_handshake.instance_id)
+    {
+        if SystemTime::now()
+            .duration_since(time)
+            .is_ok_and(|d| d < cli_arguments.reconnect_cooldown)
+        {
+            info!(
+                "Refusing connection with {peer_address} \
+                 due to reconnect cooldown ({cooldown} seconds).",
+                cooldown = cli_arguments.reconnect_cooldown.as_secs(),
+            );
+
+            // A “wrong” reason is given because of backwards compatibility.
+            // todo: Use next breaking release to give a more accurate reason here.
+            let reason = ConnectionRefusedReason::MaxPeerNumberExceeded;
+            return InternalConnectionStatus::Refused(reason);
+        }
     }
 
     // Disallow connection if max number of peers has been reached or
@@ -259,9 +281,9 @@ where
     peer.send(PeerMessage::ConnectionStatus(connection_status.into()))
         .await?;
     if let InternalConnectionStatus::Refused(reason) = connection_status {
-        let refusal_reason = format!("Refusing incoming connection. Reason: {:?}", reason);
-        warn!(refusal_reason);
-        bail!(refusal_reason);
+        let reason = format!("Refusing incoming connection. Reason: {reason:?}");
+        warn!("{reason}");
+        bail!("{reason}");
     }
 
     // Whether the incoming connection comes from a peer in bad standing is
@@ -512,6 +534,7 @@ pub(crate) async fn close_peer_connected_callback(
 
 #[cfg(test)]
 mod connect_tests {
+    use std::time::Duration;
     use std::time::SystemTime;
 
     use anyhow::bail;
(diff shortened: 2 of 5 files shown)


jan-ferdinand:
@aszepieniec, @Sword-Smith, I'm struggling to create a test for the new feature. The various state variables and channels wielded by other tests pertaining to peer connection seem to require an intimate understanding of the inner workings of the various tasks (which I lack), and seem to come with varying amounts of implicit behavior (for which documentation is lacking).

I have pushed my best attempt so far in commit 847d1b1. Clearly, there are some fundamentals I don't understand. I'd be grateful for some assistance.

aszepieniec:
> This introduces a breaking change because the public enum ConnectionRefusedReason gets a new variant RecentlyDisconnected. This enum is now marked #[non_exhaustive] to prevent future variant additions from being breaking again.

I did not realize this change would be breaking, but it is. Let's avoid modifying existing messages and add new ones instead. (Actually, I'm assuming that new messages are simply ignored by old clients, but this assumption is worth verifying.)

aszepieniec:
> Actually, I'm assuming that new messages are simply ignored by old clients, but this assumption is worth verifying.

It sounds like *any* change to `PeerMessage` would be backwards-incompatible :facepalm: So for the time being, we are not allowed to change this struct *at all*. We can speculate about how to get around this in issue #440 but that's probably out of scope for this PR, which about getting the internals to work.

jan-ferdinand:
> Clearly, there are some fundamentals I don't understand.

As far as I can tell, the problem is that the channel `from_peer_to_main_tx` gets closed after calling `answer_peer_inner`. This only seems to happen the second time `answer_peer_inner` is called. I think I need to keep this channel open somehow, or create an equivalent channel. I'd be grateful for some assistance.

src/connect_to_peers.rs
```diff
@@ -760,6 +780,53 @@ mod connect_tests {
...
+
+        let (_broadcast_tx, broadcast_rx, to_main_tx, _to_main_rx, mut state_lock, handshake) =
+            get_test_genesis_setup(network, 0, args).await?;
+        let node_0_address = get_dummy_socket_address(0);
+
+        state_lock
+            .lock_guard_mut()
+            .await
+            .net
+            .register_peer_disconnection(node_0_address, SystemTime::now());
```
aszepieniec, line 801: Consider adding a dual test case, or iterating over variants within this test case, to verify that the connection *is* allowed after the cooldown period has exceeded.

I highlight these lines because I think that's where you can change the effective elapsed time.

src/models/channel.rs
```diff
@@ -142,14 +142,19 @@ impl MainToPeerTask {
...
         }
         .to_string()
     }
 }
 
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub(crate) enum InternalDisconnectReason {
+    OutOfConnectionCapacity,
+}
+
```
aszepieniec, line 157: What's the motivation for having an enum with one variant? Isn't it cleaner to drop the enum altogether?

src/models/state/networking_state.rs
```diff
@@ -94,6 +94,12 @@ pub struct NetworkingState {
     /// record latest successful upgrade, merely latest attempt. This is to
     /// prevent excessive runs of the proof-upgrade functionality.
     pub last_tx_proof_upgrade_attempt: std::time::SystemTime,
+
+    /// Disconnection times of past peers. Can be used to determine if a connection
+    /// request should be accepted or rejected.
+    ///
+    /// Both the main task and the peer tasks may update this map.
```
aszepieniec, line 101: Should the main task be allowed to update this map? As far as I can tell, it only happens in `register_peer_disconnection` which, as far as I can tell, is only called from `peer_loop.rs` (and a test in `connect_to_peers.rs`).

src/models/channel.rs
jan-ferdinand: I'm assuming that this enum will grow in the future. For example, we could add an RPC to terminate some connection, giving rise to a new variant here.

src/models/state/networking_state.rs
jan-ferdinand: Does the main task only terminate connections by communicating to the peer tasks? If so, you are right, the main task should not directly change the map.

I worded it this way to not accidentally scare future us when stumbling across this.

src/connect_to_peers.rs
jan-ferdinand: I have expanded the existing test case.

src/models/state/networking_state.rs
aszepieniec: It looks like there is a third escape option:

 3. Trigger an error (instead of a `Ok` return code).

This escape path is taken when the peer misbehaves, resulting in a ban. And possibly other error cases also. In this case we also do not care about recording disconnection times, but even that point is moot as it all takes place inside `peer_loop.rs`.

src/models/channel.rs
aszepieniec: That sounds like a useful command to have. However, it might take a while before its priority is high enough to justify implementation and until then we are dragging around needless complexity. Where do you stand on that?

aszepieniec (approved):

src/models/channel.rs
jan-ferdinand: It's certainly a trade off. I think building for extensibility is usually a good idea, but I'm not even sure the PR does that. For example, I have now realised that the `DisconnectAll` does not state a reason, which is inconsistent. I have reverted the introduction of `InternalDisconnectReason`, favoring a clean solution now over a maybe-working-maybe-broken extensibility solution.

src/config_models/cli_args.rs
```diff
@@ -254,6 +254,15 @@ pub struct Args {
...
     #[clap(long)]
     pub disable_cookie_hint: bool,
+
+    /// The duration (in seconds) during which new connection attempts from peers
+    /// are ignored after a connection to them was closed.
+    ///
+    /// Does not affect abnormally closed connections. For example, if a connection
+    /// is dropped due to networking issues, an immediate reconnection attempt is
+    /// not affected by this cooldown.
+    #[clap(long, default_value = "30", value_parser = duration_from_seconds_str)]
```
Sword-Smith, line 264: Peer discovery runs every 2 minutes which is where a reconnect would happen. So with such a low default value, I don't think current behavior will change noticeably. I would lean towards a much longer default value. How about 30 minutes?

src/connect_to_peers.rs
```diff
@@ -757,6 +780,80 @@ mod connect_tests {
         Ok(())
     }
 
+    #[traced_test]
+    #[tokio::test]
+    async fn bootstrap_node_refuses_reconnects_within_disconnect_cooldown_period() -> Result<()> {
```
Sword-Smith, line 785: Test seems independent of "bootstrap" setting. Consider renaming.

src/connect_to_peers.rs
```diff
@@ -757,6 +780,80 @@ mod connect_tests {
...
+        let network = Network::Main;
+        let reconnect_cooldown = Duration::from_secs(3);
```
Sword-Smith, line 787: It looks to me like there's a risk that this test becomes flaky, because of your assumptions about how fast it's run, which may not hold up in all scenarios. But we can just fix that if it becomes a problem IMO.

src/peer_loop.rs
```diff
@@ -1632,15 +1632,31 @@ impl PeerLoopHandler {
...
+
+                Ok(DISCONNECT_CONNECTION)
+            }
+            MainToPeerTask::DisconnectAll() => {
+                // Disconnect from this peer, no matter what.
+                self.global_state_lock
+                    .lock_guard_mut()
+                    .await
+                    .net
+                    .register_peer_disconnection(self.peer_address, SystemTime::now());
```
Sword-Smith, line 1656: I would *only* add to this hash map in the scenario where you disconnect from a specified peer. But that's not a strong preference. Consider your opinion, then go with that.

src/connect_to_peers.rs
```diff
@@ -108,6 +109,27 @@ async fn check_if_connection_is_allowed(
...
+    {
+        if SystemTime::now()
+            .duration_since(time)
+            .is_ok_and(|d| d < cli_arguments.reconnect_cooldown)
+        {
+            info!(
+                "Refusing incoming connection from {peer_address} \
+                 due to reconnect cooldown ({cooldown} seconds).",
+                cooldown = cli_arguments.reconnect_cooldown.as_secs(),
+            );
```
Sword-Smith, line 124: The function `check_if_connection_is_allowed` that you're modifying here is invoked for both incoming and outgoing connections. But the log messages makes the assumption that you're handling an incoming connection.

src/models/state/networking_state.rs
```diff
@@ -94,6 +94,17 @@ pub struct NetworkingState {
...
+    /// Disconnection times of past peers. Can be used to determine if a connection
+    /// request should be accepted or rejected.
+    ///
+    /// Only records times of _graceful_ disconnections that were triggered by
+    /// _this_ node. That is, times of the following events are _not_ recorded:
+    /// - Graceful disconnect initiated by the peer.
+    /// - Abrupt disconnections, for example due to network failures.
+    ///
+    /// Only the peer tasks may update this map.
+    disconnection_times: HashMap<SocketAddr, SystemTime>,
```
Sword-Smith, line 107: I think the key here should be `InstanceId`, not `SocketAddr`, since `SocketAddr` is unique for each new outgoing connection that a machine makes.

src/connect_to_peers.rs
jan-ferdinand: I've increased the cooldown to 8 seconds, which should be acceptable for a test and should cut CI and other potentially-slow environments some more slack.

src/peer_loop.rs
```diff
@@ -1882,6 +1889,19 @@ impl PeerLoopHandler {
         // feature to have for testing purposes.
         res
     }
+
+    /// Register graceful peer disconnection in the global state.
```
Sword-Smith, line 1893: Comment here could be a bit more descriptive maybe. I like how the field on the networking state is very explicit.

Sword-Smith (approved):
I think you mention the wrong default value in the commit message. Otherwise LGTM.

PR #439, opened 2025-02-14
feat(CLI): Show wallet status as table

aszepieniec:
The CLI command `wallet-status` produces unwieldy output even for a modest number of wallet status elements. This commit adds another option for exporting: formatted as a markdown table. Specifically, the CLI command now takes an optional flag `--table`, in which case the table is outputted instead of the JSON data dump. Optional flag `--json` produces the original JSON data dump. Not including either flag defaults to the original JSON data dump, so in particular no old behaviors of the CLI tool have changed.

I thought I would let this commit live as a PR for a while before merging. I don't think it's particularly controversial but you never know what kind of review feedback you get.


dan-da (approved):
I just gave it a quick look, and seems fine.

One of these days we should go through all the neptune-cli commands and standardize some output formats.  ie any command that displays data should have a json option and a human option.  This seems like a good start towards that goal.

PR #445, opened 2025-02-16
refactor(peer_loop): Extra check of double spending txs

Sword-Smith:
Ensure that double spending transactions never make it into the mempool, or are relayed to other peers. 

This is not a consensus change, as the same check happens in `Block::is_valid`. But it moves the check earlier up in the pipeline. Previously, it was only checked if all removal records have valid MMR membership proofs, not that they also set indices that were not already set.

This error message is reported:
```
2025-02-16T05:45:22.184331881Z  INFO ThreadId(06) neptune_cash::peer_loop: Got block proposal from peer.                                                                                
2025-02-16T05:45:22.323280118Z  WARN ThreadId(06) neptune_cash::models::blockchain::block: mutator set update must be possible 
```

I suspect the problem is that double-spending transactions are relayed on the network. And block proposals are invalid because of this, meaning that composers are wasting resources and missing out on block rewards. And guessers are idling because they don't have any work.

### DIFF
--- a/src/models/blockchain/transaction/transaction_kernel.rs
+++ b/src/models/blockchain/transaction/transaction_kernel.rs
@@ -1,7 +1,7 @@
-use std::collections::HashSet;
 use std::sync::OnceLock;
 
 use get_size2::GetSize;
+use itertools::Itertools;
 use serde::Deserialize;
 use serde::Serialize;
 use strum::EnumCount;
@@ -20,7 +20,6 @@ use crate::models::proof_abstractions::timestamp::Timestamp;
 use crate::prelude::twenty_first;
 use crate::util_types::mutator_set::addition_record::AdditionRecord;
 use crate::util_types::mutator_set::mutator_set_accumulator::MutatorSetAccumulator;
-use crate::util_types::mutator_set::removal_record::AbsoluteIndexSet;
 use crate::util_types::mutator_set::removal_record::RemovalRecord;
 
 /// TransactionKernel is immutable and its hash never changes.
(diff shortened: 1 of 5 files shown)


dan-da:
> I suspect the problem is that double-spending transactions are relayed on the network. And block proposals are invalid because of this, meaning that composers are wasting resources and missing out on block rewards. And guessers are idling because they don't have any work.

I would feel better about the fix if this hypothesis were tested and duplicated.

If I understand correctly, the problem this is attempting to fix is that block proposals are presently taking about 9 minutes, up from about 4 minutes.   And we observe that along with these slow blocks there are often (always?) 1 or two transactions in the mempool.    And from that, the hypothesis is that the composition is taking longer because the composer first attempts to prove the invalid tx, fails, kicks it out, and starts over with a new proposal, approximately doubling the composing time.

If that hypothesis is correct, it should be possible to duplicate in a testnet, and/or unit test.    has either been done?

regardless, the fix sounds generally correct in that only valid tx should be relayed across the network.   I have a dim memory that bitcoin-core has a large set of validation checks that tx must pass before they can be relayed.   and then the receiving node must duplicate them as well before adding to mempool, because they can't trust external input.

It sounds like if we are already checking for this in Block::is_valid(), but not before getting into the mempool, that we do not have a comprensive set of relay checks in place, and should focus on that.   Maybe this or the following PR does that... I haven't looked at the code yet.   These are general observations.

src/models/blockchain/transaction/transaction_kernel.rs
```diff
@@ -98,6 +101,25 @@ impl From<PrimitiveWitness> for TransactionKernel {
     }
 }
 
+impl TransactionKernel {
+    /// Determine if the transaction double spends in input, either by spending an input that was
+    /// already spent, or by attempting to spend the same input multiple times. Relative to a given
+    /// mutator set. Will also return true if any removal records have invalid membership proofs. So
+    /// the caller should check for that before calling this method.
```
dan-da, line 108: why should caller check it?   that sounds like duplicate work.   I believe either caller should check it, or this method, not both.

Sword-Smith:
Running with this path I *do* get warnings about double-spending tx. To me, that confirms the hypothesis. See log

image

```
2025-02-16T07:04:44.83130509Z  WARN ThreadId(190) neptune_cash::peer_loop: Received double-spending tx                                                                                  
2025-02-16T07:04:44.831332823Z  WARN ThreadId(190) neptune_cash::peer_loop: Punishing peer 2001:bc8:17c0:41e:46a8:42ff:fe22:e8e9 for DoubleSpendingTransaction
```

dan-da (commented):
It's a bit strange to me that Block::is_valid() is apparently already doing this check, but now we are doing the same check in TransactionKernel.

Shouldn't Block::is_valid() call this new TransactionKernel method on the merge block tx, so we are not duplicating crtical validation logic?

Sword-Smith:
Alternatively, we could do the same check as in the block validity check: Check if the mutator set update derived from the transaction can be applied to the current tip's mutator set.

Sword-Smith:
These two extra checks on incoming transactions in the `peer_loop` has fixed the problem I had with invalid composition. I don't fully understand how the 2nd check can be hit. @aszepieniec do you have a clue?

src/peer_loop.rs
```diff
@@ -1300,17 +1301,44 @@ impl PeerLoopHandler {
...
+
+                // 7. If transaction cannot be applied to mutator set, punish. I'm not sure how this
+                // could happen when above checks pass.
+                let ms_update = MutatorSetUpdate::new(
+                    transaction.kernel.inputs.clone(),
+                    transaction.kernel.outputs.clone(),
+                );
+                let can_apply = ms_update
+                    .apply_to_accumulator(&mut mutator_set_accumulator_after.clone())
+                    .is_ok();
```
aszepieniec, line 1323: I don't see what's wrong with these six lines but I just want to point out the (afaik) new and unorthodox use of `MutatorSetUpdate`. Previously it was only ever used in relation to blocks, not individual standalone transactions. (Note that `mutator_set_update.rs` lives in `block/` not `mutator_set/`.)

aszepieniec (changes requested):
This PR adds missing logic to the validation checklist for transactions transferred by peers. I believe the missing-now-added logic is correct.

That said, I would like to structure this change differently. I focus on three steps:
 5. check if transaction is confirmable,
 6. check if transaction double-spends inputs,
 7. check if transaction-induced mutator set update can be applied.

*Confirmability* is defined as the feature of transactions (in this context) that can be confirmed in a block without making that block invalid.

Point 5 currently does not check confirmability; it *only* checks removal record validity, which in turn boils down to a) all necessary SWBFI membership proofs are present and b) valid. The function `is_confirmable_relative_to` does not to what its name suggests. So this point is better articulated as
 5. check if all removal records are valid.

Point 5, 6, and 7 are all features of confirmability. I suggest factoring out these points by expanding what `is_confirmable_relative_to` does to match its name. Since the log messages are useful, it is worth changing the return value to `Result<(), ErrorCode>` for some to-be defined `ErrorCode`.

> These two extra checks on incoming transactions in the peer_loop has fixed the problem I had with invalid composition. I don't fully understand how the 2nd check can be hit. @aszepieniec do you have a clue?

It points to a bug in `can_apply_to_accumulator`, or at least to an inconsistency between 5+6 and 7, which should be equivalent.

Sword-Smith:
It would be very nice if this patch printed the txid of transactions that are double spending.

Sword-Smith:
Superseded by #449.

PR #446, opened 2025-02-16
refactor: Check validity of own block proposal

Sword-Smith:
Before sharing own block proposal, verify that it is valid.

Acts as a sanity check that the composition is valid. If it isn't, that would indicate an error in the application.


dan-da (approved):
lgtm.

PR #449, opened 2025-02-16
refactor: Factor out unconfirmability

aszepieniec:
Funnel all tests related to unconfirmability to one function, with optional error code. Improve log messaging when receiving unconfirmable transactions from peers.


Sword-Smith:
Supersedes #445.

Looks pretty good to me. Let's see if we can't add a test or two to reach all possible error codes.

PR #469, opened 2025-02-26
style: Factor out common logic

jan-ferdinand:
Don't repeat the logic for testing whether peer standing is bad. Use one canonical method instead. Also:

- improve documentation of command line argument `peer_tolerance`
- prevent non-sensical peer tolerances from entering the application
- improve code flow by staying on the happy path more
- adhere to max line length more often
- move test code into test module

PR #474, opened 2025-02-27
Remove timestamps from `PeerStanding`

jan-ferdinand:
The struct `PeerStanding` records the most recent reward and the most recent punishment. Previously, it recorded the time of the respective event. This time was never used. Now, the time is no longer recorded.

This is a breaking change because struct `PeerStanding` is public, and so are the modified fields.

Note that this PR builds on top of #469.

### DIFF
--- a/src/bin/dashboard_src/peers_screen.rs
+++ b/src/bin/dashboard_src/peers_screen.rs
@@ -361,11 +361,11 @@ impl Widget for PeersScreen {
                 self.sort_order.compare(
                     a.standing()
                         .latest_punishment
-                        .map(|p| p.0.to_string())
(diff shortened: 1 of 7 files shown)


Sword-Smith:
I think the timestamp could be very valuable for debugging purposes -- as it allows you to go into the logs and see what caused the negative sanction. So I'd rather open an issue to make use of this timestamp in e.g. the dashboard.

Also: The CLI command `neptune-cli peer-info` prints a JSON formatted string of all peer standings. So you can actually see this data if you want to. Type checker didn't catch that because JSON.

jan-ferdinand:
Closing in favor of #475, which suggests that the timestamps should be used.

PR #478, opened 2025-02-28
Thv/external guessers

Sword-Smith:
Add infrastructure to allow guessing by other programs than neptune-core. Still missing: CLI endpoints for this feature, and callbacks when a new block proposal is received. But this is a good start.

### DIFF
--- a/src/main_loop.rs
+++ b/src/main_loop.rs
@@ -34,6 +34,7 @@ use crate::macros::log_slow_scope;
 use crate::models::blockchain::block::block_header::BlockHeader;
 use crate::models::blockchain::block::block_height::BlockHeight;
 use crate::models::blockchain::block::difficulty_control::ProofOfWork;
+use crate::models::blockchain::block::Block;
 use crate::models::blockchain::transaction::Transaction;
 use crate::models::blockchain::transaction::TransactionProof;
 use crate::models::channel::MainToMiner;
@@ -485,6 +486,41 @@ impl MainLoopHandler {
         self.main_to_miner_tx.send(MainToMiner::Continue);
     }
 
+    async fn handle_self_guessed_block(
+        &mut self,
+        main_loop_state: &mut MutableMainLoopState,
+        new_block: Box<Block>,
+    ) -> Result<()> {
+        // Store block in database
+        // This block spans global state write lock for updating.
+        let mut global_state_mut = self.global_state_lock.lock_guard_mut().await;
+
+        if !global_state_mut.incoming_block_is_more_canonical(&new_block) {
+            drop(global_state_mut); // don't hold across send()
+            warn!("Got new block from miner that was not child of tip. Discarding.");
+            self.main_to_miner_tx.send(MainToMiner::Continue);
+            return Ok(());
+        } else {
+            info!("Locally-mined block is new tip: {}", new_block.hash());
+        }
+
+        // Share block with peers first thing.
+        info!("broadcasting new block to peers");
+        self.main_to_peer_broadcast_tx
+            .send(MainToPeerTask::Block(new_block.clone()))
+            .expect("Peer handler broadcast channel prematurely closed. This should never happen.");
+
+        let update_jobs = global_state_mut
+            .set_new_tip(new_block.as_ref().clone())
+            .await?;
+        drop(global_state_mut);
+
+        self.spawn_mempool_txs_update_job(main_loop_state, update_jobs)
+            .await;
+
+        Ok(())
+    }
+
     /// Locking:
     ///   * acquires `global_state_lock` for write
     async fn handle_miner_task_message(
@@ -499,38 +535,8 @@ impl MainLoopHandler {
                 let new_block = new_block_info.block;
 
                 info!("Miner found new block: {}", new_block.kernel.header.height);
-
-                // Store block in database
-                // This block spans global state write lock for updating.
-                let mut global_state_mut = self.global_state_lock.lock_guard_mut().await;
-
-                if !global_state_mut.incoming_block_is_more_canonical(&new_block) {
-                    drop(global_state_mut); // don't hold across send()
-                    warn!("Got new block from miner task that was not child of tip. Discarding.");
-                    self.main_to_miner_tx.send(MainToMiner::Continue);
-                    return Ok(None);
-                } else {
-                    info!(
-                        "Block from miner is new canonical tip: {}",
-                        new_block.hash(),
-                    );
-                }
-
-                // Share block with peers first thing.
-                info!("broadcasting new block to peers");
-                self.main_to_peer_broadcast_tx
-                    .send(MainToPeerTask::Block(new_block.clone()))
-                    .expect(
-                        "Peer handler broadcast channel prematurely closed. This should never happen.",
-                    );
-
-                let update_jobs = global_state_mut
-                    .set_new_tip(new_block.as_ref().clone())
+                self.handle_self_guessed_block(main_loop_state, new_block)
                     .await?;
-                drop(global_state_mut);
-
-                self.spawn_mempool_txs_update_job(main_loop_state, update_jobs)
-                    .await;
             }
             MinerToMain::BlockProposal(boxed_proposal) => {
                 let (block, expected_utxos) = *boxed_proposal;
@@ -1526,7 +1532,7 @@ impl MainLoopHandler {
 
                 // Handle messages from rpc server task
                 Some(rpc_server_message) = rpc_server_to_main_rx.recv() => {
-                    let shutdown_after_execution = self.handle_rpc_server_message(rpc_server_message.clone()).await?;
+                    let shutdown_after_execution = self.handle_rpc_server_message(rpc_server_message.clone(), &mut main_loop_state).await?;
                     if shutdown_after_execution {
                         break SUCCESS_EXIT_CODE
                     }
@@ -1616,7 +1622,11 @@ impl MainLoopHandler {
 
     /// Handle messages from the RPC server. Returns `true` iff the client should shut down
     /// after handling this message.
-    async fn handle_rpc_server_message(&mut self, msg: RPCServerToMain) -> Result<bool> {
+    async fn handle_rpc_server_message(
+        &mut self,
+        msg: RPCServerToMain,
+        main_loop_state: &mut MutableMainLoopState,
+    ) -> Result<bool> {
         match msg {
             RPCServerToMain::BroadcastTx(transaction) => {
                 debug!(
@@ -1677,6 +1687,13 @@ impl MainLoopHandler {
                 }
 
                 // do not shut down
+                Ok(false)
+            }
+            RPCServerToMain::SolvePow(new_block) => {
+                info!("Handling PoW solution from RPC call");
+
+                self.handle_self_guessed_block(main_loop_state, new_block)
+                    .await?;
                 Ok(false)
             }
             RPCServerToMain::PauseMiner => {
--- a/src/rpc_server.rs
+++ b/src/rpc_server.rs
@@ -1735,6 +1735,14 @@ pub trait RPC {
     /// ```
     async fn restart_miner(token: rpc_auth::Token) -> RpcResult<()>;
 
+    /// Provide a PoW-solution to the current block proposal.
+    ///
+    /// Returns true iff the provided solution satisfies the proof-of-work
+    /// threshold of the next block. False otherwise. False can be returned if
+    /// the solution comes in too late in which case either a new block
+    /// proposal is stored by the client, or no block proposal is known at all.
+    async fn pow_solution(token: rpc_auth::Token, nonce: Digest) -> RpcResult<bool>;
+
     /// mark MUTXOs as abandoned
     ///
     /// ```no_run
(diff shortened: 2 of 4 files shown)


src/rpc_server.rs
```diff
@@ -1696,6 +1735,14 @@ pub trait RPC {
     /// ```
     async fn restart_miner(token: rpc_auth::Token) -> RpcResult<()>;
 
+    /// Provide a PoW-solution to the current block proposal.
+    ///
+    /// Returns true iff the provided solution satisfies the proof-of-work
+    /// threshold of the next block. False otherwise. False can be returned if
+    /// the solution comes in too late in which case either a new block
+    /// proposal is stored by the client, or no block proposal is known at all.
```
jan-ferdinand, line 1743: I'm confused by this explanation. Maybe something like the following works better? I might also have misinterpreted this function's purpose due to my confusion.

```rust
/// Suggest a solution for the current block proposal's proof-of-work puzzle.
///
/// In case the suggestion is no valid solution, of if there is no current block
/// proposal, or if a new block proposal has been received in the meantime, the
/// suggestion is rejected and `false` is returned.
```

I'm not sure about the “new block proposal has been received in the meantime” part. I understand it neither in the original docstring nor in my suggestion.

src/rpc_server.rs
```diff
@@ -1696,6 +1735,14 @@ pub trait RPC {
...
+    async fn pow_solution(token: rpc_auth::Token, nonce: Digest) -> RpcResult<bool>;
```
jan-ferdinand, line 1744: With this function name, I'd expect the return type to be the pow_solution.

Maybe one of the following works better?
```suggestion
    async fn is_pow_solution(token: rpc_auth::Token, nonce: Digest) -> RpcResult<bool>;
```
```suggestion
    async fn pow_solution(token: rpc_auth::Token, nonce: Digest) -> RpcResult<Result<(), ()>>;
```
```suggestion
    async fn solve_pow_puzzle(token: rpc_auth::Token, nonce: Digest) -> RpcResult<Result<(), ()>>;
```

src/rpc_server.rs
```diff
@@ -3064,6 +3111,64 @@ impl RPC for NeptuneRPCServer {
...
+            warn!(
+                "Got claimed PoW solution but PoW threshold was not met.\n\
+            Claimed solution: {solution_digest};\nthreshold: {threshold}"
+            );
+            return Ok(false);
+        }
+
+        // No time to waste! Inform main_loop!
+        let solution = Box::new(proposal);
+        let _ = self
```
jan-ferdinand, line 3164: Should a fail to send the message actually be ignored? A closed channel might indicate the application is in a critical state and should be terminated. If this is not case, it might be worth explaining why a send failure can be ignored.

src/rpc_server.rs
```diff
@@ -3129,6 +3234,43 @@ impl RPC for NeptuneRPCServer {
         Ok(Self::cpu_temp_inner())
     }
 
+    // documented in trait. do not add doc-comment.
```
jan-ferdinand, line 3237: Same as before.

src/rpc_server.rs
```diff
@@ -3064,6 +3111,64 @@ impl RPC for NeptuneRPCServer {
...
+        token: rpc_auth::Token,
+        nonce: Digest,
+    ) -> RpcResult<bool> {
+        log_slow_scope!(fn_name!());
+        token.auth(&self.valid_tokens)?;
+
+        let Some(mut proposal) = self
+            .state
+            .lock_guard()
+            .await
```
jan-ferdinand, line 3127: The lock guard is acquired multiple times. Can this be factored out into a variable?

src/rpc_server.rs
```diff
@@ -3129,6 +3234,43 @@ impl RPC for NeptuneRPCServer {
...
+        _context: tarpc::context::Context,
+        token: rpc_auth::Token,
+    ) -> RpcResult<Option<GuessChallengeInfo>> {
+        log_slow_scope!(fn_name!());
+        token.auth(&self.valid_tokens)?;
+
+        let Some(proposal) = self
+            .state
+            .lock_guard()
+            .await
```
jan-ferdinand, line 3249: The lock guard is acquired multiple times. Can this be factored out into a variable?

src/rpc_server.rs
```diff
@@ -3129,6 +3234,43 @@ impl RPC for NeptuneRPCServer {
...
+        let (guesser_key_after_image, latest_block_header) = {
+            let state = self.state.lock_guard().await;
+            let guesser_key_after_image = state
+                .wallet_state
+                .wallet_secret
+                .guesser_spending_key(proposal.header().prev_block_digest);
+            let latest_block_header = *state.chain.light_state().header();
+
+            (guesser_key_after_image.after_image(), latest_block_header)
+        };
```
jan-ferdinand, line 3265: Why so roundabout?

```suggestion
        let state = self.state.lock_guard().await;
        let guesser_key_after_image = state
            .wallet_state
            .wallet_secret
            .guesser_spending_key(proposal.header().prev_block_digest)
            .after_image();
        let latest_block_header = *state.chain.light_state().header();
```

src/rpc_server.rs
```diff
@@ -3129,6 +3234,43 @@ impl RPC for NeptuneRPCServer {
...
+
+        let Some(proposal) = self
+            .state
+            .lock_guard()
+            .await
+            .block_proposal
+            .map(|x| x.to_owned())
+        else {
+            return Ok(None);
+        };
```
jan-ferdinand, line 3254:
```suggestion
        let state = self.state.lock_guard().await;
        let Some(proposal) = state.block_proposal.map(|x| x.to_owned()) else {
            return Ok(None);
        };
```

jan-ferdinand (commented):
I can't comment on the architectural design since I'm too unfamiliar with the architecture of Neptune core. The construction can benefit from a few additional iterations.

src/rpc_server.rs
```diff
@@ -3064,6 +3111,64 @@ impl RPC for NeptuneRPCServer {
         Ok(())
     }
 
+    // documented in trait. do not add doc-comment.
```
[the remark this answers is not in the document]
dan-da: because previously doc-comments were being added to fn in trait impl, and it was becoming a bit of a mess in the impl RpcServer, which is a huge file worked on by many people.    This comment serves as a reminder.

src/rpc_server.rs
dan-da: > Why so roundabout?

just some background:

The first way drops the lock guard when the end of (inner) scope is reached, releasing the lock.

Your suggestion keeps the lock around until the end of outer scope.   It can be explicitly dropped with `drop(state)` but the suggestion does not do that, so it actually changes behavior (assuming there is subsequent code in the outer scope).

We always want to drop the locks as quickly as possible after usage, and never hold them across lengthy operations, or a second attempt to acquire the lock.    There are basically 3 ways to acquire/use/drop:

1. in a single statement:  `global_state_lock.lock_guard().do_something()`
2. in a new scope: `{let state = global_state_lock.lock_guard(); do_something()}`
3. with a drop: `let state = global_state_lock.lock_guard(); do_something(); drop(state);`

I tend to prefer (1) and (2) and use (3) somewhat sparingly.    I find that (3) is more likely to lead to drop() being forgotten or later removed, while (2) makes the drop automatic and impossible to forget.

src/rpc_server.rs
dan-da: again, be mindful that the suggested change is not dropping the acquired lock.  I haven't checked subsequent code, so maybe it's ok, but it is a logic change.

src/rpc_server.rs
jan-ferdinand: The advantage of an explicit `drop(…)` is that is communicates what is happening instead of relying on implicit behavior. I think this helps readability and maintainability; scoping might simplify initial construction.

I personally am less likely to remove a `drop(…)` than I am to rewrite an oddly scoped block that looks like an artifact of a refactoring tool (in this case, inlining a method).

aszepieniec (changes requested):
The word "challenge" to describe the information needed to completely specify a computational task and test candidate solutions is a poor choice of words and I strongly suggest finding an alternative.

In computational complexity and in cryptography the word used to describe this information is *problem*. More specifically it is a *search* problem (as opposed to a decision problem; because the output is more than one bit). Opinions differ about what the thing being searched for ought to be called, but the feature of hash functions that makes them suitable for use in proof-of-work is *correlation intractability*. That said, I'm not sure a scientifically accurate term is the best choice of words for type, variable, and function names.

In the context of proof-of-work, and particularly in less scientific and more layperson-friendly literature, I've seen "proof-of-work puzzle" used often. I suggest to go with this phrase.

The reason why "challenge" is a poor choice is because in cryptography and decentralized protocols it denotes one step of an interactive challenge and response protocol. A verifier sets the challenge and scrutinizes the response in order to verify the authenticity of prior or subsequent messages, or the authorization of the responder. These features do not describe proof-of-work.

src/rpc_server.rs
dan-da: Those are valid points.    However removing a drop() during maintenance is only half of it.    The other half is remembering to put it there in the first place, and to keep it in the right place as code is refactored.    If one uses the convention that every time we acquire a lock guard for use by 2 or more statements we start a new scope, then there is less chance of forgetting to add the drop, and unwittingly hold it across a lengthy operation.

Either way, it is pattern/convention within the codebase that requires some familiarity to get right.  Comments are helpful, especially for new eyes to the codebase.    And we should always pay extra attention to any lock changes in PRs.

src/main_loop.rs
```diff
@@ -485,6 +486,41 @@ impl MainLoopHandler {
         self.main_to_miner_tx.send(MainToMiner::Continue);
     }
 
+    async fn handle_self_guessed_block(
```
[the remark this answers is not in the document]
Sword-Smith: Let's also try to write a test for this newly-factored out function.

src/rpc_server.rs
```diff
@@ -199,6 +203,36 @@ impl MempoolTransactionInfo {
...
+/// the minting of the next block.
+#[derive(Clone, Debug, Copy, Serialize, Deserialize)]
+pub struct GuessChallengeInfo {
+    kernel_auth_path: [Digest; BlockKernel::MAST_HEIGHT],
+    header_auth_path: [Digest; BlockHeader::MAST_HEIGHT],
+    threshold: Digest,
+    total_guesser_reward: NativeCurrencyAmount,
+}
+
+impl GuessChallengeInfo {
```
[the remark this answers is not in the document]
Sword-Smith: `ProofOfWorkPuzzle` is OK, although the data structure also informs about the guesser's potential reward.

src/rpc_server.rs
Sword-Smith: I'm happy with `provide_pow_solution`

src/rpc_server.rs
```diff
@@ -3064,6 +3111,64 @@ impl RPC for NeptuneRPCServer {
...
+        else {
+            warn!(
+                "Got claimed PoW solution but no challenge was known. \
+            Did solution come in too late?"
+            );
+            return Ok(false);
+        };
+
+        // A proposal was found. Check if solution works.
+        let (guesser_key_after_image, latest_block_header) = {
```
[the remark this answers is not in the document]
Sword-Smith: Agree with the variable-renaming. The code, though, is structured this way to minimize the time that read locks are held.

src/rpc_server.rs
```diff
@@ -4183,6 +4326,152 @@ mod rpc_server_tests {
...
+                .await
+                .unwrap();
+            assert!(
+                !accepted,
+                "Must reject PoW solution when no proposal exists"
+            );
+        }
+
+        #[tokio::test]
+        async fn exported_guess_challenge_is_consistent_with_block_hash() {
```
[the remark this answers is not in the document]
Sword-Smith: `guess_challenge_info_is_consistent_with_block_hash` (two tests above) tests the logic of the data structure holding the data of the puzzle. This test tests the entire pipeline.

src/rpc_server.rs
Sword-Smith: That's intentional. To return as quickly as possible and release all locks as quickly as possible.

Sword-Smith:
Integrated most reviewer feedback. Still releasing locks as soon as possible, though.

src/rpc_server.rs
jan-ferdinand: I don't have all the context that comes into play here. With that caveat out of the way:

- I'm generally in favor of owning up to inherent complexity instead of delegating difficult decisions by default.[^difficult] If we, with years of accumulated knowledge, don't know what's right (or usually right), can we really expect other people to know?
- If there are odd cases or rough edges that are not dealt with, they should at least be outlined as clearly as possible, for example in the documentation.

[^difficult]: _Allowing_ different decisions at the same time is even better.

src/rpc_server.rs
```diff
@@ -1294,6 +1333,11 @@ pub trait RPC {
     /// ```
     async fn cpu_temp(token: rpc_auth::Token) -> RpcResult<Option<f32>>;
 
+    /// Get the proof-of-work challenge for the next block.
```
aszepieniec, line 1336:
```suggestion
    /// Get the proof-of-work puzzle for the current block proposal.
```

src/rpc_server.rs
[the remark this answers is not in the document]
aszepieniec: The variable rename dispels all the confusion I was struggling with. :+1:

aszepieniec (changes requested):
The miners potentially expend a lot of resources on guessing infrastructure. So when a a successful guess cannot be sent to the main thread (for whatever reason) it would be costly to let that information disappear. Rather than letting it disappear, I suggest to log an error message that contains the proof-of-work solution.

If I'm not mistaken the test for `handle_self_guessed_block` is yet to be written, is that correct?

> What to do with the raise condition where the client registers a new block proposal (not a new block) and the external guesser (simultaneously) finds the block for the old block proposal that no longer is stored in the client's state.

Well phrased.

> The solution I've come up with is to store exported block proposals to a file somewhere.

That's a fine solution.

Here are two alternatives worth considering:
 - Store it in RAM, on the state of the RPC server.
 - Send the block proposal and send the block back, instead of the the proof-of-work puzzle and the solution.

src/main_loop.rs
```diff
@@ -485,6 +486,45 @@ impl MainLoopHandler {
...
 
+    /// Process a block whose PoW solution was solved by this client (or an
+    /// external program) and has not been seen by the rest of the network yet.
+    ///
+    /// Shares block with all connected peers, updates own state, and updates
+    /// any mempool transactions to be valid under this new block.
+    ///
+    /// Locking:
+    ///  * acquires `global_state_lock` for write
+    async fn handle_self_guessed_block(
```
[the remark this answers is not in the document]
Sword-Smith: Good question! I just added one in c48b2f1ddf6c8a9c4a67484bcd71f66d05f7332f.

Sword-Smith:
Ready for review again. Now all block exported block proposals are stored in RAM. This set is cleared, though, whenever a new tip is set. Also added test of `handle_self_guessed_block`. Added new field to `GlobalState`: `mining_state`.

aszepieniec (changes requested):
> Now all block exported block proposals are stored in RAM. This set is cleared, though, whenever a new tip is set. Also added test of handle_self_guessed_block. Added new field to GlobalState: mining_state.

I'm confused. I can't find any code fitting this description. Did you forget to add a file or something?

Sword-Smith:
> I'm confused. I can't find any code fitting this description. Did you forget to add a file or something?

Yes. Sorry. Force-pushed from wrong machine. Fixed and added ability to get proposal without knowing guesser preimage. Please review.

src/main_loop.rs
```diff
@@ -485,6 +486,45 @@ impl MainLoopHandler {
...
+        self.main_to_peer_broadcast_tx
+            .send(MainToPeerTask::Block(new_block.clone()))
+            .expect("Peer handler broadcast channel prematurely closed.");
+
+        let update_jobs = global_state_mut.set_new_tip(*new_block).await?;
+        drop(global_state_mut);
+
+        self.spawn_mempool_txs_update_job(main_loop_state, update_jobs)
+            .await;
+
```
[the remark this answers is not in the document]
aszepieniec: Aha, I see the new lines for doing that in `set_new_tip`.

Is that the right place to put it though? I think `set_new_tip` is also called when in sync mode. In that case, the list of exported block proposals will be repeatedly cleared; and the current block proposal will be repeatedly reset to `None`. This seems like a waste of resources. (I'm not saying its a concretely noticeable performance degradation but instead that it points to an imperfection in the design.)

The alternative would be to move those lines here. And repeat that code in handlers of foreign blocks. Okay, maybe that's not ideal either: code repetition also points to imperfect design.

Either way, there is something worth considering here.

And: whatever design solution we come up with, clearing the list of exported proposals and setting the current one to `None` is *definitely* behavior that's expected from this function so please expand the test to check for it.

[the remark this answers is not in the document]
aszepieniec: Okay, it looks like it's already being tested in a separate test. So no need to expand the test.

aszepieniec (approved):
Just the one design question to think about. And if the conclusion is that no change is necessary, then go ahead and merge.

PR #484, opened 2025-03-05
Make clippy more pedantic

jan-ferdinand:
I'm flagging all of you for review because
- this change affects the entire codebase, and
- since preferences differ, we might want to discuss the benefits and drawbacks of certain lints in greater detail.

## Enabled and fixed lints

I have enabled and fixed the following lints. Each commit fixes one lint (or a small number of identical-in-spirit lints) to simplify their reversion in case there are controversies. 

- [`cloned_instead_of_copied`](https://rust-lang.github.io/rust-clippy/master/index.html#cloned_instead_of_copied)
- [`explicit_into_iter_loop`](https://rust-lang.github.io/rust-clippy/master/index.html#explicit_into_iter_loop)
- [`explicit_iter_loop`](https://rust-lang.github.io/rust-clippy/master/index.html#explicit_iter_loop)
- [`filter_map_next`](https://rust-lang.github.io/rust-clippy/master/index.html#filter_map_next)
- [`if_not_else`](https://rust-lang.github.io/rust-clippy/master/index.html#if_not_else)
- [`inefficient_to_string`](https://rust-lang.github.io/rust-clippy/master/index.html#inefficient_to_string)
- [`items_after_statements`](https://rust-lang.github.io/rust-clippy/master/index.html#items_after_statements)
- [`manual_assert`](https://rust-lang.github.io/rust-clippy/master/index.html#manual_assert)
- [`manual_let_else`](https://rust-lang.github.io/rust-clippy/master/index.html#manual_let_else)
- [`match_wildcard_for_single_variants`](https://rust-lang.github.io/rust-clippy/master/index.html#match_wildcard_for_single_variants)
- [`missing_copy_implementations`](https://dev-doc.rust-lang.org/stable/rustc/lints/listing/allowed-by-default.html#missing-copy-implementations)
- [`missing_debug_implementations`](https://dev-doc.rust-lang.org/stable/rustc/lints/listing/allowed-by-default.html#missing-debug-implementations)
- [`needless_for_each`](https://rust-lang.github.io/rust-clippy/master/index.html#needless_for_each)
- [`no_effect_underscore_binding`](https://rust-lang.github.io/rust-clippy/master/index.html#no_effect_underscore_binding)
- [`redundant_else`](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_else)
- [`ref_option_ref`](https://rust-lang.github.io/rust-clippy/master/index.html#ref_option_ref)
- [`unnecessary_wraps`](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_wraps)
- [`unnested_or_patterns`](https://rust-lang.github.io/rust-clippy/master/index.html#unnested_or_patterns)
- [`used_underscore_binding`](https://rust-lang.github.io/rust-clippy/master/index.html#used_underscore_binding)

## Lints I'd like to enable

The following lints are not fulfilled, but I think we would benefit from them:

- [`cast_lossless`](https://rust-lang.github.io/rust-clippy/master/index.html#cast_lossless) – I strongly prefer using `as` casting _only_ if necessary, relying on `from` and `try_from` otherwise. This way, encountering an `as` cast clearly communicates that possible truncation has been considered and accepted by the developer. If you also see things this way, I'm happy to enable this lint. _edit_: This lint is now enabled
- [`let_underscore_drop`](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#let-underscore-drop) – Currently, `let _ = foo();` is most commonly used to drop types `Result<(), _>` when it doesn't matter if the function succeeds. I think this is a legit way of using `let _ = …`. Yet, allowing this lint obscures cases where the result should not be dropped, like `let _ = mutex.lock();` or `let (tx, _) = channel(…);`. We might benefit from enforcing a pattern like `let _unused = foo()`. _edit_: I have changed my mind on this lint; see below.
- [`struct_field_names`](https://rust-lang.github.io/rust-clippy/master/index.html#struct_field_names) – Fixing this requires renaming (private) fields, which is not completely mechanical.
- [`uninlined_format_args`](https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args) – I'm a fan of `format!("x is {x} and y is {y}")` and try to avoid `format!("x is {} and y is {}", x, y)`. If you also prefer this, I'm happy to enable this lint. _edit_: The style this lint tries to enforce is not generally favored.
- [`unused_async`](https://rust-lang.github.io/rust-clippy/master/index.html#unused_async) – I might not understand all the reasons why a function that seemingly does not need to be `async` does, in fact, need to be `async`. _edit_: this lint is now enabled

## Enabled lints

Additionally, I have enabled the following lints, all of which were already fulfilled:

- [`copy_iterator`](https://rust-lang.github.io/rust-clippy/master/index.html#copy_iterator)
- [`doc_link_with_quotes`](https://rust-lang.github.io/rust-clippy/master/index.html#doc_link_with_quotes)
- [`enum_glob_use`](https://rust-lang.github.io/rust-clippy/master/index.html#enum_glob_use)
- [`expl_impl_clone_on_copy`](https://rust-lang.github.io/rust-clippy/master/index.html#expl_impl_clone_on_copy)
- [`explicit_deref_methods`](https://rust-lang.github.io/rust-clippy/master/index.html#explicit_deref_methods)
- [`flat_map_option`](https://rust-lang.github.io/rust-clippy/master/index.html#flat_map_option)
- [`fn_params_excessive_bools`](https://rust-lang.github.io/rust-clippy/master/index.html#fn_params_excessive_bools)
- [`from_iter_instead_of_collect`](https://rust-lang.github.io/rust-clippy/master/index.html#from_iter_instead_of_collect)
- [`invalid_upcast_comparisons`](https://rust-lang.github.io/rust-clippy/master/index.html#invalid_upcast_comparisons)
- [`large_digit_groups`](https://rust-lang.github.io/rust-clippy/master/index.html#large_digit_groups)
- [`macro_use_imports`](https://rust-lang.github.io/rust-clippy/master/index.html#macro_use_imports)
- [`manual_instant_elapsed`](https://rust-lang.github.io/rust-clippy/master/index.html#manual_instant_elapsed)
- [`manual_ok_or`](https://rust-lang.github.io/rust-clippy/master/index.html#manual_ok_or)
- [`many_single_char_names`](https://rust-lang.github.io/rust-clippy/master/index.html#many_single_char_names)
- [`match_wild_err_arm`](https://rust-lang.github.io/rust-clippy/master/index.html#match_wild_err_arm)
- [`mismatching_type_param_order`](https://rust-lang.github.io/rust-clippy/master/index.html#mismatching_type_param_order)
- [`mut_mut`](https://rust-lang.github.io/rust-clippy/master/index.html#mut_mut)
- [`needless_continue`](https://rust-lang.github.io/rust-clippy/master/index.html#needless_continue)
- [`needless_if`](https://rust-lang.github.io/rust-clippy/master/index.html#needless_if)
- [`option_option`](https://rust-lang.github.io/rust-clippy/master/index.html#option_option)
- [`ref_binding_to_reference`](https://rust-lang.github.io/rust-clippy/master/index.html#ref_binding_to_reference)
- [`same_functions_in_if_condition`](https://rust-lang.github.io/rust-clippy/master/index.html#same_functions_in_if_condition)
- [`unchecked_duration_subtraction`](https://rust-lang.github.io/rust-clippy/master/index.html#unchecked_duration_subtraction)
- [`unnecessary_box_returns`](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_box_returns)
- [`unnecessary_join`](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_join)
- [`verbose_bit_mask`](https://rust-lang.github.io/rust-clippy/master/index.html#verbose_bit_mask)


src/database/storage/storage_schema/schema.rs
```diff
@@ -72,6 +72,7 @@ use crate::locks::tokio::LockCallbackFn;
 /// It works equally well to put them in a `struct`.  If the tables
 /// are all of the same type (including generics), they could be
 /// placed in a collection type such as `Vec`, or `HashMap`.
+#[derive(Debug, Clone)]
 pub struct DbtSchema {
```
Sword-Smith, line 76: I'm not sure what `Debug` and `Clone` could be used for, or even means, for these structures. They also probably shouldn't be `pub` I think.

src/main_loop.rs
```diff
@@ -87,6 +87,7 @@ const TX_UPDATER_CHANNEL_CAPACITY: usize = 1;
 ///  a) the miner might not exist in which case there would be no-one to empty
 ///     the channel; and
 ///  b) contrary to other channels, transmission failures here are not critical.
+#[derive(Debug, Clone)]
```
Sword-Smith, line 90: Not sure this *should* be clonable.

src/main_loop.rs
```diff
@@ -1778,6 +1776,8 @@ mod test {
     }
 
     async fn setup(num_init_peers_outgoing: u8, num_peers_incoming: u8) -> TestSetup {
+        const CHANNEL_CAPACITY: usize = 10;
```
Sword-Smith, line 1779: Bad name. Maybe rename to `CHANNEL_CAPACITY_MINER_TO_MAIN`? I know you didn't make the name, but when you're moving it further from where it's being used, context is less clear than before.

src/mine_loop.rs
```diff
@@ -623,13 +624,15 @@ pub(crate) async fn mine(
...
-    tokio::time::sleep(Duration::from_secs(INITIAL_MINING_SLEEP_IN_SECONDS)).await;
-    let cli_args = global_state_lock.cli().clone();
 
     // Set PoW guessing to restart every N seconds, if it has been started. Only
     // the guesser task may set this to actually resolve, as this will otherwise
     // abort e.g. the composer.
     const GUESSING_RESTART_INTERVAL_IN_SECONDS: u64 = 20;
+
+    tokio::time::sleep(Duration::from_secs(INITIAL_MINING_SLEEP_IN_SECONDS)).await;
+    let cli_args = global_state_lock.cli().clone();
```
Sword-Smith, line 634: Not sure why we need an owned version of CLI args here. Would a reference do?

src/models/blockchain/transaction/utxo.rs
```diff
@@ -163,7 +163,7 @@ impl Utxo {
 
     /// If the UTXO has a timelock, find out what the release date is.
     pub fn release_date(&self) -> Option<Timestamp> {
-        self.coins.iter().filter_map(Coin::release_date).next()
+        self.coins.iter().find_map(Coin::release_date)
```
Sword-Smith, line 166: Might be a bug here (unrelated to your change), we probably want to take the "biggest" release data here, in case URXO has multiple instances of the same timelock type script.

src/models/blockchain/type_scripts/native_currency_amount.rs
```diff
@@ -473,7 +471,7 @@ impl NativeCurrencyAmount {
             match sign {
                 num_bigint::Sign::Minus => -nau_rational.numer(),
                 num_bigint::Sign::Plus => nau_rational.numer().to_owned(),
-                _ => unreachable!(),
+                num_bigint::Sign::NoSign => unreachable!(),
```
Sword-Smith, line 474: I like these rewrites, as it gives a compiler error if a new variant is added.

src/models/blockchain/type_scripts/time_lock.rs
```diff
@@ -48,8 +48,7 @@ impl TimeLock {
     pub fn extract_release_date(utxo: &Utxo) -> Timestamp {
         utxo.coins()
             .iter()
-            .filter_map(Coin::release_date)
-            .next()
+            .find_map(Coin::release_date)
             .unwrap_or_else(Timestamp::zero)
```
Sword-Smith, line 52: Cf. https://github.com/Neptune-Crypto/neptune-core/issues/487

src/util_types/archival_mmr.rs
```diff
@@ -21,6 +21,7 @@ use crate::prelude::twenty_first;
 ///
 /// Merkle Mountain Ranges only know about hashes. When values are to be associated with
 /// MMRs, these values must be stored by the caller, or in a wrapper to this data structure.
+#[derive(Debug, Clone)]
 pub struct ArchivalMmr<Storage: StorageVec<Digest>> {
```
Sword-Smith, line 25: I think of this as a structured database. What does it mean to clone a database? I guess it's the connection to the database that's being cloned, which might be unproblematic. Still, not completely clear to me, and I see no good reason to allow cloning here.

Sword-Smith (changes requested):
I'm neutral or positive towards most changes here. I'm a bit skeptical about deriving `Clone` for all the database structures. If you're only ever supposed to have one instance of them, I think it's nicer to just not be able to clone them.

jan-ferdinand:
@Sword-Smith, thanks for your feedback. Where do you stand with regards to the lints that I would like to enable but haven't yet (see OP)?

aszepieniec:
I disagree with `if_not_else`. A frequent pattern that occurs in the course of validation is

```rust
if !ok {
    gracefully_abort();
}
```

which makes perfect sense in and of itself.

Sometimes you also want to inform the user that control flow proceeds to the next step by dropping a log message. The snippet then becomes

```rust
if !ok {
    gracefully_abort();
}

// control can't get here
log!("test ok; proceeding to next step");
```

So far so good. Sometimes you don't want to abort but inject a new step. And you want to log the message to indicate that the extra injected step is skipped.

```rust
if !ok {
    extra_step();
} else {
    log!("test ok; no need for extra step");
}
```

IIuc, this linter will insist that the "correct" way to write the same logic is:

```rust
if ok {
    log!("test ok; no need for extra step");
} else {
    extra_step();
}
```

This is counter-intuitive:
 - the extra step is called after the log message about it sent;
 - the "else" branch communicates lesser importance because it is optional -- but the importance here is the exact opposite: things don't really break down if you forget about the log message, but if you forget the extra step when it's necessary Bad Things Happen.

Otherwise I'm happy with the selection of linter rules.

jan-ferdinand:
@aszepieniec, thanks for your feedback. Where do you stand with regards to the lints that I would like to enable but haven't yet (see OP)?

Regarding `if_not_else`: I see what you mean. Applied to the concrete codebase at hand, the argument strikes me as a theoretic one: In my opinion, the commit that introduces and fixes the lint (0c34fa6e) improves readability in 4 out of the 5 cases, and readability remains unchanged in the [fifth case](https://github.com/Neptune-Crypto/neptune-core/commit/0c34fa6ef3204fdec7d4b22f8eb1a98011d03f3a#diff-a65778f9ae3c4c27b3b4b89732db26ea328a108482bf7e03eec89953f796995aL1527-L1530).[^difficult] I can remove the lint, but would like to include the fixes for it.

[^difficult]: Here, the surrounding code makes overall readability low already; refactoring the entire section seems like a good idea to me. This might remove this fifth case entirely.

aszepieniec:
> Where do you stand with regards to the lints that I would like to enable but haven't yet (see OP)?

:+1: 

I agree with your analysis about 0c34fa6 and think the changes therein are an improvement. Let's keep the changes and remove the lint.

I can be more specific about my  objection: flagging code of the form `if !condition { a } else { b }` is a good heuristic that *usually, but not always,* indicates suboptimal code, but the recommendation to fix it by writing `if condition { b } else { a }` instead is terrible blanket advice because it *sometimes* leads to worse code.

IMHO, enabled lints should satisfy the stronger criterion that they *always* lead to better code.

src/database/storage/storage_schema/schema.rs
jan-ferdinand: Regarding `pub` visibility: #496.

src/util_types/archival_mmr.rs
jan-ferdinand: This is not necessarily a structured database; whether it is depends on the `Storage` type. As I see it, it is perfectly valid to clone an archival Merkle Mountain Range in general.

Also, note the types. `Storage` is a generic type argument. As you surely know, `#[derive(Clone)]` only applies if the generic type implements `Clone`. Here is a [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=4321f807fbf6bafad20c54e8f27c6239) in case you want a refresher.

src/mine_loop.rs
jan-ferdinand: Maybe. It might cause borrowing issues later, when `global_state_lock.lock` is used while `cli_args` are still in scope. Note that (a) even though the diff suggests it, I did not introduce this line, and (b) `CliArgs` is pretty cheap to clone. Since this is not about any of the lints, I'll mark this conversation as resolved. Fee free to open a corresponding issue.

jan-ferdinand:
> > Where do you stand with regards to the lints that I would like to enable but haven't yet (see OP)?
> 
> 👍
> 
Is that signifying unconditional agreement with me enabling those five lints?

aszepieniec:
> > > Where do you stand with regards to the lints that I would like to enable but haven't yet (see OP)?
> > 
> > 
> > 👍
> 
> Is that signifying unconditional agreement with me enabling those five lints?

👍

dan-da:
> I have enabled and fixed the following lints. Each commit fixes one lint (or a small number of identical-in-spirit lints) to simplify their reversion in case there are controversies.
> 
> * [`cloned_instead_of_copied`](https://rust-lang.github.io/rust-clippy/master/index.html#cloned_instead_of_copied)
> * [`explicit_into_iter_loop`](https://rust-lang.github.io/rust-clippy/master/index.html#explicit_into_iter_loop)
> * [`explicit_iter_loop`](https://rust-lang.github.io/rust-clippy/master/index.html#explicit_iter_loop)
> * [`filter_map_next`](https://rust-lang.github.io/rust-clippy/master/index.html#filter_map_next)
> * [`if_not_else`](https://rust-lang.github.io/rust-clippy/master/index.html#if_not_else)
> * [`inefficient_to_string`](https://rust-lang.github.io/rust-clippy/master/index.html#inefficient_to_string)
> * [`items_after_statements`](https://rust-lang.github.io/rust-clippy/master/index.html#items_after_statements)
> * [`manual_assert`](https://rust-lang.github.io/rust-clippy/master/index.html#manual_assert)
> * [`manual_let_else`](https://rust-lang.github.io/rust-clippy/master/index.html#manual_let_else)
> * [`match_wildcard_for_single_variants`](https://rust-lang.github.io/rust-clippy/master/index.html#match_wildcard_for_single_variants)
> * [`missing_copy_implementations`](https://dev-doc.rust-lang.org/stable/rustc/lints/listing/allowed-by-default.html#missing-copy-implementations)
> * [`missing_debug_implementations`](https://dev-doc.rust-lang.org/stable/rustc/lints/listing/allowed-by-default.html#missing-debug-implementations)
> * [`needless_for_each`](https://rust-lang.github.io/rust-clippy/master/index.html#needless_for_each)
> * [`no_effect_underscore_binding`](https://rust-lang.github.io/rust-clippy/master/index.html#no_effect_underscore_binding)
> * [`redundant_else`](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_else)
> * [`ref_option_ref`](https://rust-lang.github.io/rust-clippy/master/index.html#ref_option_ref)
> * [`unnecessary_wraps`](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_wraps)
> * [`unnested_or_patterns`](https://rust-lang.github.io/rust-clippy/master/index.html#unnested_or_patterns)
> * [`used_underscore_binding`](https://rust-lang.github.io/rust-clippy/master/index.html#used_underscore_binding)

I'm ok with enabling these.

I vote yay or nay for each below.

> The following lints are not fulfilled, but I think we would benefit from them:
> 
> * [`cast_lossless`](https://rust-lang.github.io/rust-clippy/master/index.html#cast_lossless) – I strongly prefer using `as` casting _only_ if necessary, relying on `from` and `try_from` otherwise. This way, encountering an `as` cast clearly communicates that possible truncation has been considered and accepted by the developer. If you also see things this way, I'm happy to enable this lint.

yay.  should eliminate a possible cause of panics.

> * [`let_underscore_drop`](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#let-underscore-drop) – Currently, `let _ = foo();` is most commonly used to drop types `Result<(), _>` when it doesn't matter if the function succeeds. I think this is a legit way of using `let _ = …`. Yet, allowing this lint obscures cases where the result should not be dropped, like `let _ = mutex.lock();` or `let (tx, _) = channel(…);`. We might benefit from enforcing a pattern like `let _unused = foo()`.

yay.  

> * [`struct_field_names`](https://rust-lang.github.io/rust-clippy/master/index.html#struct_field_names) – Fixing this requires renaming (private) fields, which is not completely mechanical.

nay.    since enabling it seems to add work/headache.  (maybe later?)

> * [`uninlined_format_args`](https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args) – I'm a fan of `format!("x is {x} and y is {y}")` and try to avoid `format!("x is {} and y is {}", x, y)`. If you also prefer this, I'm happy to enable this lint.

nay.  I prefer the latter style because it works for both variables and expressions, while the former only works for variables, and also because it is ingrained habit going back decades from other langs. 

> * [`unused_async`](https://rust-lang.github.io/rust-clippy/master/index.html#unused_async) – I might not understand all the reasons why a function that seemingly does not need to be `async` does, in fact, need to be `async`.

yay.    as code gets refactored, an async fn (a) that previously called async fn (b) might remove the call to (b) and thus no longer need to be async.  It would be nice to be notified about this.    note however that changing (a) also necessitates that all callers of (a) remove the a.await, which could chain react even higher.    So this might be one to #[allow()] in some cases.

dan-da (approved):
lgtm!   (the lints list.  I haven't reviewed the 128 files).

let's merge it before it gets stale.

jan-ferdinand:
@dan-da, thanks for your feedback. :blush: 

>> [cast_lossless](https://rust-lang.github.io/rust-clippy/master/index.html#cast_lossless)
>
> yay. should eliminate a possible cause of panics.

I don't think `as` casting ever panics. The lint is mostly about drawing attention to those `as` casts that are left once enabled by removing some of the [many meanings](https://doc.rust-lang.org/reference/expressions/operator-expr.html#type-cast-expressions) an `as` cast has from the codebase.

>> [struct_field_names](https://rust-lang.github.io/rust-clippy/master/index.html#struct_field_names)
>
> nay. since enabling it seems to add work/headache. (maybe later?)

This lint triggers on only 2 types (`AuthenticateCoinbaseFields` and `ClaimUtxoData`) so in some sense, it's not a lot of work. However, I couldn't find better field (and type) names, so in this sense, it is a lot of work, and maybe a small headache.

>> [unused_async](https://rust-lang.github.io/rust-clippy/master/index.html#unused_async)
>
> yay. as code gets refactored […]

Great, I had hoped that this was the reason. :ok_hand:

jan-ferdinand:
Regarding [`let_underscore_drop`](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#let-underscore-drop), I have learned:

> It is idiomatic to use a [let statement](https://doc.rust-lang.org/reference/statements.html#let-statements) with a pattern of `_` when a must-used value is purposely discarded.[^idiom]

In addition, the lints [`let_underscore_future`](https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_future) and [`let_underscore_lock`](https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_lock) are already enabled.[^lock]

Taken together, I have changed my mind on the [`let_underscore_drop`](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#let-underscore-drop) lint.

[^idiom]: https://doc.rust-lang.org/reference/attributes/diagnostics.html#r-attributes.diagnostics.must_use.trait-impl-function
[^lock]: The lint `let_underscore_lock` only supports `mutex` and `rwlock` in `parking_lot`. For `std` locks, there is the deny-by-default `rustc` lint [`let_underscore_lock`](https://doc.rust-lang.org/nightly/rustc/lints/listing/deny-by-default.html#let-underscore-lock). I'm not sure there is an equivalent for `tokio` locks.

dan-da:
> > yay. should eliminate a possible cause of panics.
> 
> I don't think `as` casting ever panics. The lint is mostly about drawing attention to those `as` casts that are left once enabled by removing some of the [many meanings](https://doc.rust-lang.org/reference/expressions/operator-expr.html#type-cast-expressions) an `as` cast has from the codebase.

oh right.  It doesn't panic but it can wrap and cause logic bugs, eg when casting u64 to u32, i32 to u32, etc.   So it's useful to allow safe casts  but not potentially bug causing casts.

https://users.rust-lang.org/t/safe-casting-of-primitives/24240/10

> > > [struct_field_names](https://rust-lang.github.io/rust-clippy/master/index.html#struct_field_names)
> > 
> > 
> > nay. since enabling it seems to add work/headache. (maybe later?)
> 
> This lint triggers on only 2 types (`AuthenticateCoinbaseFields` and `ClaimUtxoData`) so in some sense, it's not a lot of work. However, I couldn't find better field (and type) names, so in this sense, it is a lot of work, and maybe a small headache.

I would leave it up to you, if you want to deal with it or not.

jan-ferdinand:
@Sword-Smith, do you have any further concerns with this PR?

PR #488, opened 2025-03-06
Leave bootstrap nodes alone if well connected

jan-ferdinand:
In order to help newcomers to connect to peers on the network, certain nodes act as bootstrap nodes. Those nodes tend to be well known, and users tend to specify them as potential peers via command line arguments.

Previously, connections to potential peers specified as command line arguments were always re-initiated when dropped. This can cause an involuntary denial-of-service of those well-known bootstrap nodes.

With this PR, a well-connected node does not initiate connections with bootstrap nodes.

### DIFF
--- a/src/connect_to_peers.rs
+++ b/src/connect_to_peers.rs
@@ -23,6 +23,7 @@ use tracing::warn;
 
 use crate::models::channel::MainToPeerTask;
 use crate::models::channel::PeerTaskToMain;
+use crate::models::peer::bootstrap_info::BootstrapStatus;
 use crate::models::peer::ConnectionRefusedReason;
 use crate::models::peer::InternalConnectionStatus;
 use crate::models::peer::NegativePeerSanction;
@@ -309,12 +310,23 @@ where
     info!("Connection accepted from {peer_address}");
 
     // If necessary, disconnect from another, existing peer.
-    if connection_status == InternalConnectionStatus::AcceptedMaxReached && state.cli().bootstrap {
+    let self_is_bootstrap = state.cli().bootstrap;
+    if connection_status == InternalConnectionStatus::AcceptedMaxReached && self_is_bootstrap {
         info!("Maximum # peers reached, so disconnecting from an existing peer.");
         peer_task_to_main_tx
             .send(PeerTaskToMain::DisconnectFromLongestLivedPeer)
             .await?;
     }
+
+    // inform the new peer about our bootstrap status
+    let bootstrap_status = if self_is_bootstrap {
+        BootstrapStatus::Bootstrap
+    } else {
+        BootstrapStatus::Ordinary
+    };
+    peer.send(PeerMessage::BootstrapStatus(bootstrap_status))
+        .await?;
+    debug!("Informing {peer_address} of our bootstrap status: {bootstrap_status}");
 
     let peer_distance = 1; // All incoming connections have distance 1
     let mut peer_loop_handler = PeerLoopHandler::new(
@@ -557,6 +569,9 @@ mod connect_tests {
 
     use anyhow::bail;
     use anyhow::Result;
+    use bytes::Bytes;
+    use proptest::prelude::*;
+    use test_strategy::proptest;
     use tokio_test::io::Builder;
     use tracing_test::traced_test;
     use twenty_first::math::digest::Digest;
@@ -583,9 +598,12 @@ mod connect_tests {
     #[traced_test]
     #[tokio::test]
     async fn test_outgoing_connection_succeed() -> Result<()> {
-        let network = Network::Alpha;
-        let other_handshake = get_dummy_handshake_data_for_genesis(network);
-        let own_handshake = get_dummy_handshake_data_for_genesis(network);
+        let network = Network::Main;
+        let (_tx, main_to_peer_rx, peer_to_main_tx, _rx, state, own_handshake) =
+            get_test_genesis_setup(network, 0, cli_args::Args::default()).await?;
+        let (peer_handshake, peer_socket_address) =
+            get_dummy_peer_connection_data_genesis(network, 0);
+
         let mock = Builder::new()
             .write(&to_bytes(&PeerMessage::Handshake(Box::new((
                 MAGIC_STRING_REQUEST.to_vec(),
@@ -593,7 +611,7 @@ mod connect_tests {
             ))))?)
             .read(&to_bytes(&PeerMessage::Handshake(Box::new((
                 MAGIC_STRING_RESPONSE.to_vec(),
-                other_handshake,
+                peer_handshake,
             ))))?)
             .read(&to_bytes(&PeerMessage::ConnectionStatus(
                 TransferConnectionStatus::Accepted,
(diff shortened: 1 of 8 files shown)


src/peer_loop.rs
```diff
@@ -1544,6 +1559,10 @@ impl PeerLoopHandler {
 
                 Ok(KEEP_CONNECTION_ALIVE)
             }
+            PeerMessage::BootstrapStatus(status) => {
+                self.set_bootstrap_status(status).await;
+                Ok(KEEP_CONNECTION_ALIVE)
+            }
```
Sword-Smith, line 1565: This status should only be set *once* per connection. So I would first check if it is set for the peer in question, and only then acquire a write-lock. If the status is already set, punish peer. We should be very stingy about write locks.

jan-ferdinand: I want to give nodes a chance to change their bootstrap status.

Once per connection is probably fine – after all, you need to restart a node to change its bootstrap status – but the way you suggest to go about that is more restrictive than once per connection. Your suggestion means that a node can never update its own bootstrap status, it can only ever set it once. Only when the peer is restarted, and forgets the bootstrap status of all its past peers, can the status be set again.

src/main_loop.rs
```diff
@@ -228,137 +231,156 @@ impl SyncState {
...
-            .iter()
-            .choose(&mut rng)
-            .map(|x| (x.0.to_owned(), x.1.distance))
+    ///   * acquires `global_state_lock` for read
+    async fn is_well_connected(&self) -> bool {
+        /// A node with more peers than this ratio of its maximum number of
+        /// peers is considered well-connected.
+        const WELL_CONNECTED_RATIO: f32 = 0.8;
+
+        let curr_num_peers = self.global_state_lock.lock_guard().await.net.peer_map.len();
```
Sword-Smith, line 352: This is the 2nd time a read lock is being grabbed by  this task. This will lead to a deadlock if any task requests a write lock between the 1st lock being given and the 2nd being given. We *really* should have a way to test that the same task is never requested twice.

This complicated pitfall is why I prefer to get all relevant data from global state upfront, as far out in the callstack as possible. It does lead to more complicated function interfaces, but on the upside, you functions can become purely functional and locking issues are much easier to reason about.

src/peer_loop.rs
```diff
@@ -203,6 +206,87 @@ impl PeerLoopHandler {
         sanction_result.map_err(|err| anyhow::anyhow!("Cannot reward banned peer: {err}"))
     }
 
+    /// Update a peer's bootstrap status.
+    ///
+    /// Per connection, the bootstrap status can be set once without
+    /// incurring [punishment](Self::punish). After a cooldown period, the
+    /// bootstrap status can be set again. This is to prevent the peer from
```
Sword-Smith, line 213: Please disambiguate what "This" refers to.

src/peer_loop.rs
```diff
@@ -203,6 +206,87 @@ impl PeerLoopHandler {
...
+    /// forcing a write-lock acquisition on the global state lock.
+    ///
+    /// This method only acquires a write lock for the global state lock if
+    /// necessary.
+    ///
+    /// # Locking:
+    ///   * acquires `global_state_lock` for read
+    ///   * might acquire `global_state_lock` for write
+    async fn set_bootstrap_status(&mut self, status: BootstrapStatus) -> Result<()> {
```
Sword-Smith, line 222: As with the issue in `main_loop`, this method takes locks a bit too far into the callstack for my taste. However, what this method actually is, is the handling of `PeerMessage::Bootstrap`, so if we were to factor out the handling of all peer messages (please don't do that in this PR) to separate methods, this pattern would make more sense, I think. I think just renaming this method to "handle_boostrap_message" or something similar would fix my concern -- that might even provide a nice template for a later rewrite of `handle_peer_message`.

Sword-Smith (changes requested):

src/main_loop.rs
```diff
@@ -228,137 +231,170 @@ impl SyncState {
...
         }
     }
 }
 
-/// holds information about a set of potential peers in the process of peer discovery
+/// The potential peers in the process of peer discovery.
+#[derive(Debug, Clone)]
 struct PotentialPeersState {
-    potential_peers: HashMap<SocketAddr, PotentialPeerInfo>,
+    global_state_lock: GlobalStateLock,
```
Sword-Smith, line 255: Very few structs get to have a copy of this lock. It's `PeerLoopHandler`, the main state-carrying struct of `peer_loop`, the `mine_loop` (a spawned infinite loop), the `RpcServer`, and `MainLoopHandler`, the state-carrying struct of `main_loop`. Carrying around this lock in less central structures than these main ones, make it very hard to reason about locks, which we really need to be able to do, as it's one of the biggest weaknesses and difficulties with neptune-core. Also notice that your suggested change gives `MutableMainLoopState` a copy of the lock, something it hasn't had earlier.

So what can we do instead of giving `PotentialPeersState` a copy of the lock? I guess we can pass all the required arguments which does, admittedly lead to some big function signatures. Alternatively, we can pass the acquired lock (`GlobalState`, not `GlobalStateLock`) to the function that makes (the most) use of it `peer_candidate` and pass. That way, the lock is acquired further out in the call stack, and it's much easier to reason about potential deadlocks. Or maybe the function `peer_candidate` gets a reference to the `NetworkState`, which would imply that a read lock is held throughout its execution.

I prefer the getting the values you need in the `discover_peers` method on `MainLoopHandler` and then

Sword-Smith (changes requested):

src/main_loop.rs
jan-ferdinand: I've changed the fields of struct `PotentialPeersState` and the parameters to function `peer_candidate` to not include the `GlobalStateLock` or the `GlobalState`. Instead, `PotentialPeersState::new` takes a reference to the `cli_args::Args`, and function `peer_candidate` takes a reference to `NetworkingState`.

src/connect_to_peers.rs
```diff
@@ -1251,4 +1269,61 @@ mod connect_tests {
...
+            main_to_peer_rx,
+            peer_to_main_tx,
+            own_handshake,
+        )
+        .await?;
+
+        let bootstrap_status = &state.lock_guard().await.net.bootstrap_status;
+        assert_eq!(1, bootstrap_status.len());
+        let peer_status = bootstrap_status.get(&peer_socket_address).unwrap();
+        assert_eq!(BootstrapStatus::Ordinary, peer_status.status);
```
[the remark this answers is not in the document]
Sword-Smith: Also: If I understand this correctly, it checks if bootstrap status of the newly connected peer is recorded when the transaction is incoming. But a more relevant scenario is that our node initiates the transaction, the called node reveals its bootstrap status, and we record it. We could have a test for that also.

src/main_loop.rs
```diff
@@ -228,137 +232,177 @@ impl SyncState {
...
-        max_distance_candidates
-            .iter()
-            .choose(&mut rng)
-            .map(|x| (x.0.to_owned(), x.1.distance))
+        self.candidates
+            .values()
+            .filter(|candidate| !is_self(candidate))
+            .filter(|candidate| !same_id_is_connected(candidate))
+            .filter(|candidate| !same_socket_is_connected(candidate))
+            .filter(|candidate| !self_is_well_connected || !is_bootstrap_node(candidate))
```
Sword-Smith, line 358: Rewrite with De Morgan's law to `!(self_is_well_connected && is_bootstrap_node(candidate))`? I find that simpler to parse. If you disagree, leave it as it is.

src/connect_to_peers.rs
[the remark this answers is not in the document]
jan-ferdinand: I have generalized this test over the connection's direction and the two involved bootstrap status.

src/main_loop.rs
```diff
@@ -2374,15 +2422,74 @@ mod test {
...
+            let (_, peer_state) = set_up_peer_candidate_test(global_state_lock.clone()).await;
+
+            let net_state = &global_state_lock.lock_guard().await.net;
+            let chosen_candidate = peer_state.peer_candidate(net_state);
+            assert!(chosen_candidate.is_none());
+        }
+
+        /// Create a [PotentialPeersState] and [add](PotentialPeersState::add)
+        /// one [peer candidate](PeerCandidate). The added peer candidate as
+        /// well as the resulting [PotentialPeersState] are returned.
```
Sword-Smith, line 2471: Maybe mention that the added peer is a bootstrap node?

src/main_loop.rs
```diff
@@ -2374,15 +2422,74 @@ mod test {
...
+        ///
+        ///   * acquires `global_state_lock` for write
```
Sword-Smith, line 2473: In general, this comment is not needed for test helpers, I think. As we don't need to reason so carefully about deadlocks there.

jan-ferdinand: In my opinion, a deadlocking test is also pretty annoying. If a simple comment helps to figure out the problem, that line is well spent.

jan-ferdinand: More generally, all code is code we need to maintain. This includes test code. The standards for writing test code should be at least as high as for writing non-test code.

jan-ferdinand:
ping @Sword-Smith

Sword-Smith (approved):

PR #490, opened 2025-03-06
Thv/ 481 pw to pc upgrade race condition

### LINKED ISSUE #481: Proof upgrading PW->PC / block race condition
Users report the following scenario:
A `ProofCollection`-capable computer initiates a transaction. It is first backed by a `PrimitiveWitness` which then needs to be upgraded to `ProofCollection`. This upgrade is made by the transaction initiator since the primitive witness contains secret data and may never be shared with peers. But prior to the completion of this upgrade, a new block is received, so when the upgrade is done, the mutator set data related to the transaction is deprecated. The protocol (currently) does not allow for the sharing of transactions with deprecated mutator set data, so a message is logged, and the transaction discarded, [as can be seen here](https://github.com/Neptune-Crypto/neptune-core/blob/master/src/main_loop/proof_upgrader.rs#L379). 

I see two potential solutions:

1. Transaction initiator retains the original `PritmitiveWitness`, upgrades it to the new block, and then attempts to make a new `ProofCollection` that can then be shared.
2. The protocol is made more lenient and we allow for the sharing of `ProofCollection` with deprecated mutator set data. A machine upgrading from proof collection to single proof can add a `Update` step to this proof upgrading after completing the `SingleProof`.

Sword-Smith:
Close #481 by way of (1): "Transaction initiator retains the original PritmitiveWitness, upgrades it to the new block, and then attempts to make a new ProofCollection that can then be shared."

Fixes broken `PrimitiveWitness::validate`.

### DIFF
--- a/src/main_loop/proof_upgrader.rs
+++ b/src/main_loop/proof_upgrader.rs
@@ -298,6 +298,7 @@ impl UpgradeJob {
 
             // No locks may be held here!
             let (upgraded, expected_utxos) = match upgrade_job
+                .clone()
                 .upgrade(
                     triton_vm_job_queue,
                     job_options,
@@ -326,7 +327,7 @@ impl UpgradeJob {
                 }
             };
 
-            let new_update_job: UpdateMutatorSetDataJob = {
+            upgrade_job = {
                 let mut global_state = global_state_lock.lock_guard_mut().await;
                 // Did we receive a new block while proving? If so, perform an
                 // update also, if this was requested (and we have a single proof)
@@ -375,11 +376,6 @@ impl UpgradeJob {
                     return;
                 }
 
-                let TransactionProof::SingleProof(single_proof) = upgraded.proof else {
-                    info!("Cannot perform update, as we don't have a SingleProof");
-                    return;
-                };
-
                 let Some(ms_update) = global_state
                     .chain
                     .archival_state_mut()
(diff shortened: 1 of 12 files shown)


src/models/blockchain/transaction/primitive_witness.rs
```diff
@@ -410,6 +428,111 @@ impl PrimitiveWitness {
...
+            let membership_proofs = &mut primitive_witness
+                .input_membership_proofs
+                .iter_mut()
+                .collect_vec();
+
+            MsMembershipProof::batch_update_from_remove(membership_proofs, removal_record)
+                .expect("MS MP update from add must succeed in wallet handler");
+
+            msa_state.remove(removal_record);
+        }
```
aszepieniec, line 503: I've seen this logic before. Are you sure this not repeated code? If so, I would strongly suggest to factor out a function for it (if it doesn't exist already).

src/models/blockchain/transaction/primitive_witness.rs
```diff
@@ -1546,11 +1671,121 @@ mod test {
...
+        prop_assert!(
+            TransactionProof::Witness(updated_pw)
+                .verify(new_kernel_hash)
+                .await
+        );
+    }
+
+    #[traced_test]
+    #[tokio::test]
+    async fn arb_is_valid_unit_test_small() {
```
aszepieniec, line 1714: `arb` in the context of proptest and test-strategy means something different from the thing you're testing. Why not spell it out?

src/main_loop/proof_upgrader.rs
```diff
@@ -326,7 +327,7 @@ impl UpgradeJob {
                 }
             };
 
-            let new_update_job: UpdateMutatorSetDataJob = {
+            upgrade_job = {
                 let mut global_state = global_state_lock.lock_guard_mut().await;
                 // Did we receive a new block while proving? If so, perform an
                 // update also, if this was requested (and we have a single proof)
```
aszepieniec, line 333: This policy is outdated.

aszepieniec (changes requested):
Basic structure looks good to me. I do have some superficial comments inline.

PR #491, opened 2025-03-06
Streamline Transaction Creation Interface

aszepieniec:
Prior to this PR, the transaction creation tools were rather unwieldy because
 - there were many constructors, offering various levels of customization
 - in some cases extra data was needed.

This PR refactors this interface. Now you call `create_transaction` with a `TxCreationConfig` object which you prepare prior to the call. The function returns a `TxCreationArtifacts` object which definitely contains the transaction, but possibly also (depending on the config) other data.

This PR also contains a new test which doesn't really test anything but whose main purpose is to present the block size and relative composition of blocks with progressively larger numbers of inputs and outputs. (To make this test work, it was necessary to avoid re-selecting UTXOs for the next transaction, which led to the entire interface refactor.)

I'm looking for feedback on the new interface. What kind of tests make sense?

### DIFF
--- a/src/main_loop.rs
+++ b/src/main_loop.rs
@@ -1963,7 +1963,7 @@ mod test {
                 .use_job_queue(&dummy_queue);
             let global_state = global_state_lock.lock_guard().await;
             global_state
-                .create_transaction_with_config(
+                .create_transaction(
                     vec![].into(),
                     fee,
                     in_seven_months,
--- a/src/main_loop.rs
+++ b/src/main_loop.rs
@@ -1963,7 +1963,7 @@ mod test {
                 .use_job_queue(&dummy_queue);
             let global_state = global_state_lock.lock_guard().await;
             global_state
-                .create_transaction_with_prover_capability(
+                .create_transaction_with_config(
                     vec![].into(),
                     fee,
                     in_seven_months,
--- a/src/main_loop.rs
+++ b/src/main_loop.rs
@@ -1963,14 +1963,10 @@ mod test {
                 .use_job_queue(&dummy_queue);
             let global_state = global_state_lock.lock_guard().await;
             global_state
-                .create_transaction(
-                    vec![].into(),
-                    fee,
-                    in_seven_months,
-                    &config,
-                )
+                .create_transaction(vec![].into(), fee, in_seven_months, &config)
                 .await
                 .unwrap()
+                .transaction
         }
 
         #[tokio::test]
--- a/src/main_loop.rs
+++ b/src/main_loop.rs
@@ -1934,7 +1934,7 @@ mod test {
         use crate::models::blockchain::type_scripts::native_currency_amount::NativeCurrencyAmount;
         use crate::models::peer::transfer_transaction::TransactionProofQuality;
         use crate::models::proof_abstractions::timestamp::Timestamp;
-        use crate::models::state::wallet::utxo_notification::UtxoNotificationMedium;
+        use crate::models::state::tx_initiation_config::TxInitiationConfig;
 
         async fn tx_no_outputs(
             global_state_lock: &GlobalStateLock,
@@ -1956,20 +1956,21 @@ mod test {
                 .timestamp
                 + Timestamp::months(7);
 
+            let dummy_queue = TritonVmJobQueue::dummy();
+            let config = TxInitiationConfig::default()
+                .recover_change_off_chain(change_key.into())
+                .with_prover_capability(tx_proof_type)
+                .use_job_queue(&dummy_queue);
             let global_state = global_state_lock.lock_guard().await;
             global_state
                 .create_transaction_with_prover_capability(
                     vec![].into(),
-                    change_key.into(),
-                    UtxoNotificationMedium::OffChain,
                     fee,
                     in_seven_months,
-                    tx_proof_type,
-                    &TritonVmJobQueue::dummy(),
+                    &config,
                 )
                 .await
                 .unwrap()
-                .0
         }
 
         #[tokio::test]
--- a/src/main_loop.rs
+++ b/src/main_loop.rs
@@ -1934,7 +1934,7 @@ mod test {
         use crate::models::blockchain::type_scripts::native_currency_amount::NativeCurrencyAmount;
         use crate::models::peer::transfer_transaction::TransactionProofQuality;
         use crate::models::proof_abstractions::timestamp::Timestamp;
-        use crate::models::state::tx_initiation_config::TxInitiationConfig;
+        use crate::models::state::tx_creation_config::TxCreationConfig;
 
         async fn tx_no_outputs(
             global_state_lock: &GlobalStateLock,
@@ -1957,7 +1957,7 @@ mod test {
                 + Timestamp::months(7);
 
             let dummy_queue = TritonVmJobQueue::dummy();
-            let config = TxInitiationConfig::default()
+            let config = TxCreationConfig::default()
                 .recover_change_off_chain(change_key.into())
                 .with_prover_capability(tx_proof_type)
                 .use_job_queue(&dummy_queue);
--- a/src/main_loop.rs
+++ b/src/main_loop.rs
@@ -1963,7 +1963,7 @@ mod test {
                 .use_job_queue(&dummy_queue);
             let global_state = global_state_lock.lock_guard().await;
             global_state
-                .create_transaction(vec![].into(), fee, in_seven_months, &config)
+                .create_transaction(vec![].into(), fee, in_seven_months, config)
                 .await
                 .unwrap()
                 .transaction
--- a/src/models/state/mempool.rs
+++ b/src/models/state/mempool.rs
@@ -835,7 +835,7 @@ mod tests {
     use crate::models::blockchain::transaction::Transaction;
     use crate::models::blockchain::type_scripts::native_currency_amount::NativeCurrencyAmount;
     use crate::models::shared::SIZE_20MB_IN_BYTES;
-    use crate::models::state::tx_initiation_config::TxInitiationConfig;
+    use crate::models::state::tx_creation_config::TxCreationConfig;
     use crate::models::state::tx_proving_capability::TxProvingCapability;
     use crate::models::state::wallet::expected_utxo::UtxoNotifier;
     use crate::models::state::wallet::transaction_output::TxOutput;
@@ -988,7 +988,7 @@ mod tests {
         let in_seven_months = genesis_block.kernel.header.timestamp + Timestamp::months(7);
         let high_fee = NativeCurrencyAmount::coins(15);
         let dummy_job_queue = TritonVmJobQueue::dummy();
-        let config = TxInitiationConfig::default()
+        let config = TxCreationConfig::default()
             .recover_change(bob_spending_key.into(), UtxoNotificationMedium::OnChain)
             .with_prover_capability(TxProvingCapability::ProofCollection)
             .use_job_queue(&dummy_job_queue);
@@ -1144,7 +1144,7 @@ mod tests {
         let in_seven_months = now + Timestamp::months(7);
         let in_eight_months = now + Timestamp::months(8);
         let dummy_queue = TritonVmJobQueue::dummy();
-        let config_bob = TxInitiationConfig::default()
+        let config_bob = TxCreationConfig::default()
             .recover_change(bob_spending_key.into(), UtxoNotificationMedium::OnChain)
             .with_prover_capability(TxProvingCapability::SingleProof)
             .use_job_queue(&dummy_queue);
@@ -1187,7 +1187,7 @@ mod tests {
             true,
         )];
         let dummy_job_queue = TritonVmJobQueue::dummy();
-        let config_alice = TxInitiationConfig::default()
+        let config_alice = TxCreationConfig::default()
             .recover_change(alice_key.into(), UtxoNotificationMedium::OffChain)
             .with_prover_capability(TxProvingCapability::SingleProof)
             .use_job_queue(&dummy_job_queue);
@@ -1482,7 +1482,7 @@ mod tests {
         let now = genesis_block.kernel.header.timestamp;
         let in_seven_years = now + Timestamp::months(7 * 12);
         let dummy_queue = TritonVmJobQueue::dummy();
-        let config = TxInitiationConfig::default()
+        let config = TxCreationConfig::default()
             .recover_change(alice_key.into(), UtxoNotificationMedium::OffChain)
             .with_prover_capability(proving_capability)
             .use_job_queue(&dummy_queue);
@@ -1612,7 +1612,7 @@ mod tests {
                 );
                 let tx_outputs: TxOutputList = vec![receiver_data.clone()].into();
                 let dummy_queue = TritonVmJobQueue::dummy();
-                let config = TxInitiationConfig::default()
+                let config = TxCreationConfig::default()
                     .recover_change_on_chain(premine_spending_key.into())
                     .with_prover_capability(TxProvingCapability::ProofCollection)
                     .use_job_queue(&dummy_queue);
@@ -1817,7 +1817,7 @@ mod tests {
             .await;
             let in_seven_months = genesis_block.kernel.header.timestamp + Timestamp::months(7);
             let dummy_queue = TritonVmJobQueue::dummy();
-            let config = TxInitiationConfig::default()
+            let config = TxCreationConfig::default()
                 .recover_change_on_chain(bob_spending_key.into())
                 .with_prover_capability(proof_type)
                 .use_job_queue(&dummy_queue);
--- a/src/models/state/mempool.rs
+++ b/src/models/state/mempool.rs
@@ -835,6 +835,7 @@ mod tests {
     use crate::models::blockchain::transaction::Transaction;
     use crate::models::blockchain::type_scripts::native_currency_amount::NativeCurrencyAmount;
     use crate::models::shared::SIZE_20MB_IN_BYTES;
+    use crate::models::state::tx_initiation_config::TxInitiationConfig;
     use crate::models::state::tx_proving_capability::TxProvingCapability;
     use crate::models::state::wallet::expected_utxo::UtxoNotifier;
     use crate::models::state::wallet::transaction_output::TxOutput;
@@ -986,17 +987,19 @@ mod tests {
         .await;
         let in_seven_months = genesis_block.kernel.header.timestamp + Timestamp::months(7);
         let high_fee = NativeCurrencyAmount::coins(15);
-        let (tx_by_bob, _, _maybe_change_output) = bob
+        let dummy_job_queue = TritonVmJobQueue::dummy();
+        let config = TxInitiationConfig::default()
+            .recover_change(bob_spending_key.into(), UtxoNotificationMedium::OnChain)
+            .with_prover_capability(TxProvingCapability::ProofCollection)
+            .use_job_queue(&dummy_job_queue);
+        let tx_by_bob = bob
             .lock_guard()
             .await
             .create_transaction_with_prover_capability(
                 vec![].into(),
-                bob_spending_key.into(),
-                UtxoNotificationMedium::OnChain,
                 high_fee,
                 in_seven_months,
-                TxProvingCapability::ProofCollection,
-                &TritonVmJobQueue::dummy(),
+                &config,
             )
             .await
             .unwrap();
@@ -1140,24 +1143,26 @@ mod tests {
         let now = genesis_block.kernel.header.timestamp;
         let in_seven_months = now + Timestamp::months(7);
         let in_eight_months = now + Timestamp::months(8);
-        let (tx_by_bob, _, maybe_change_output) = bob
+        let dummy_queue = TritonVmJobQueue::dummy();
+        let config_bob = TxInitiationConfig::default()
+            .recover_change(bob_spending_key.into(), UtxoNotificationMedium::OnChain)
+            .with_prover_capability(TxProvingCapability::SingleProof)
+            .use_job_queue(&dummy_queue);
+        let tx_by_bob = bob
             .lock_guard()
             .await
             .create_transaction_with_prover_capability(
                 utxos_from_bob.clone(),
-                bob_spending_key.into(),
-                UtxoNotificationMedium::OnChain,
                 NativeCurrencyAmount::coins(1),
                 in_seven_months,
-                TxProvingCapability::SingleProof,
-                &TritonVmJobQueue::dummy(),
+                &config_bob,
             )
             .await
             .unwrap();
 
         // inform wallet of any expected utxos from this tx.
         let expected_utxos = bob.lock_guard().await.wallet_state.extract_expected_utxos(
-            utxos_from_bob.concat_with(maybe_change_output),
+            utxos_from_bob.concat_with(config_bob.change_output()),
             UtxoNotifier::Myself,
         );
         bob.lock_guard_mut()
@@ -1181,17 +1186,19 @@ mod tests {
             alice_address.into(),
             true,
         )];
-        let (tx_from_alice_original, _, _maybe_change_output) = alice
+        let dummy_job_queue = TritonVmJobQueue::dummy();
+        let config_alice = TxInitiationConfig::default()
+            .recover_change(alice_key.into(), UtxoNotificationMedium::OffChain)
+            .with_prover_capability(TxProvingCapability::SingleProof)
+            .use_job_queue(&dummy_job_queue);
+        let tx_from_alice_original = alice
             .lock_guard()
             .await
             .create_transaction_with_prover_capability(
                 utxos_from_alice.into(),
-                alice_key.into(),
-                UtxoNotificationMedium::OffChain,
                 NativeCurrencyAmount::coins(1),
                 in_seven_months,
-                TxProvingCapability::SingleProof,
-                &TritonVmJobQueue::dummy(),
+                &config_alice,
             )
             .await
             .unwrap();
@@ -1474,17 +1481,19 @@ mod tests {
             .to_owned();
         let now = genesis_block.kernel.header.timestamp;
         let in_seven_years = now + Timestamp::months(7 * 12);
-        let (unmined_tx, _, _maybe_change_output) = alice
+        let dummy_queue = TritonVmJobQueue::dummy();
+        let config = TxInitiationConfig::default()
+            .recover_change(alice_key.into(), UtxoNotificationMedium::OffChain)
+            .with_prover_capability(proving_capability)
+            .use_job_queue(&dummy_queue);
+        let unmined_tx = alice
             .lock_guard()
             .await
             .create_transaction_with_prover_capability(
                 vec![tx_receiver_data].into(),
-                alice_key.into(),
-                UtxoNotificationMedium::OffChain,
                 NativeCurrencyAmount::coins(1),
                 in_seven_years,
-                proving_capability,
-                &TritonVmJobQueue::dummy(),
+                &config,
             )
             .await
             .unwrap();
@@ -1602,22 +1611,23 @@ mod tests {
                     false,
                 );
                 let tx_outputs: TxOutputList = vec![receiver_data.clone()].into();
-                let (tx, _, _maybe_change_output) = preminer_clone
+                let dummy_queue = TritonVmJobQueue::dummy();
+                let config = TxInitiationConfig::default()
+                    .recover_change_on_chain(premine_spending_key.into())
+                    .with_prover_capability(TxProvingCapability::ProofCollection)
+                    .use_job_queue(&dummy_queue);
+                preminer_clone
                     .clone()
                     .lock_guard()
                     .await
                     .create_transaction_with_prover_capability(
                         tx_outputs.clone(),
-                        premine_spending_key.into(),
-                        UtxoNotificationMedium::OnChain,
                         fee,
                         in_seven_months,
-                        TxProvingCapability::ProofCollection,
-                        &TritonVmJobQueue::dummy(),
+                        &config,
                     )
                     .await
-                    .expect("producing proof collection should succeed");
-                tx
+                    .expect("producing proof collection should succeed")
             };
 
         assert_eq!(0, preminer.lock_guard().await.mempool.len());
@@ -1806,17 +1816,19 @@ mod tests {
             )
             .await;
             let in_seven_months = genesis_block.kernel.header.timestamp + Timestamp::months(7);
-            let (tx_by_bob, _, _maybe_change_output) = bob
+            let dummy_queue = TritonVmJobQueue::dummy();
+            let config = TxInitiationConfig::default()
+                .recover_change_on_chain(bob_spending_key.into())
+                .with_prover_capability(proof_type)
+                .use_job_queue(&dummy_queue);
+            let tx_by_bob = bob
                 .lock_guard()
                 .await
                 .create_transaction_with_prover_capability(
                     vec![].into(),
-                    bob_spending_key.into(),
-                    UtxoNotificationMedium::OnChain,
                     fee,
                     in_seven_months,
-                    proof_type,
-                    &TritonVmJobQueue::dummy(),
+                    &config,
                 )
                 .await
                 .unwrap();
--- a/src/models/state/mempool.rs
+++ b/src/models/state/mempool.rs
@@ -995,7 +995,7 @@ mod tests {
         let tx_by_bob = bob
             .lock_guard()
             .await
-            .create_transaction(vec![].into(), high_fee, in_seven_months, &config)
+            .create_transaction(vec![].into(), high_fee, in_seven_months, config)
             .await
             .unwrap()
             .transaction;
@@ -1151,7 +1151,7 @@ mod tests {
                 utxos_from_bob.clone(),
                 NativeCurrencyAmount::coins(1),
                 in_seven_months,
-                &config_bob,
+                config_bob,
             )
             .await
             .unwrap();
@@ -1195,7 +1195,7 @@ mod tests {
                 utxos_from_alice.into(),
                 NativeCurrencyAmount::coins(1),
                 in_seven_months,
-                &config_alice,
+                config_alice,
             )
             .await
             .unwrap()
@@ -1491,7 +1491,7 @@ mod tests {
                 vec![tx_receiver_data].into(),
                 NativeCurrencyAmount::coins(1),
                 in_seven_years,
-                &config,
+                config,
             )
             .await
             .unwrap()
@@ -1619,7 +1619,7 @@ mod tests {
                     .clone()
                     .lock_guard()
                     .await
-                    .create_transaction(tx_outputs.clone(), fee, in_seven_months, &config)
+                    .create_transaction(tx_outputs.clone(), fee, in_seven_months, config)
                     .await
                     .expect("producing proof collection should succeed")
             };
@@ -1821,7 +1821,7 @@ mod tests {
             let tx = bob
                 .lock_guard()
                 .await
-                .create_transaction(vec![].into(), fee, in_seven_months, &config)
+                .create_transaction(vec![].into(), fee, in_seven_months, config)
                 .await
                 .unwrap()
                 .transaction;
--- a/src/models/state/mempool.rs
+++ b/src/models/state/mempool.rs
@@ -995,7 +995,7 @@ mod tests {
         let tx_by_bob = bob
             .lock_guard()
             .await
-            .create_transaction_with_config(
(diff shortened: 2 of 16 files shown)


dan-da:
Sorry, I don't have time to review just now.

However I wanted to drop a pointer to branch [offchain_serialized_pr](https://github.com/Neptune-Crypto/neptune-core/blob/offchain_serialized_pr) which never got merged in its entirety.   It might have some ideas/features that can be incorporated into present streamlining (which I fully support any effort in that direction).  

In that branch, create_transaction() has been heavily refactored and simplified.  It supports creation of any transaction type, ie non-native coins.  So does the send() RPC call, which supports any number of outputs, so send_to_many() was removed..  Change is calculated in a previous step by generate_tx_params(), a new RPC API, so there is no need to pass in a change spending key, get an extra change utxo, etc.

https://github.com/Neptune-Crypto/neptune-core/blob/70f40bce44655c901db5c1459a1c3bff83b2de74/src/models/state/mod.rs#L816

https://github.com/Neptune-Crypto/neptune-core/blob/70f40bce44655c901db5c1459a1c3bff83b2de74/src/models/state/mod.rs#L226

https://github.com/Neptune-Crypto/neptune-core/blob/70f40bce44655c901db5c1459a1c3bff83b2de74/src/rpc_server.rs#L728

https://github.com/Neptune-Crypto/neptune-core/blob/70f40bce44655c901db5c1459a1c3bff83b2de74/src/rpc_server.rs#L692

jan-ferdinand (commented):
Great change. :+1: I like the interface simplification.

I have also skimmed the branch Dan mentions in an earlier comment. The single `send` function with one parameter of type `TxParams` seems a lot cleaner than having both `send` and `send_to_many` with a list of 5 and 4 parameters, respectively.[^params] I'm aware you're not changing the `send` RPC in this PR; I mention it because Dan's branch also simplifies the interface for `create_transaction` by using the same `TxParams`. I'm wondering whether the `TxCreationConfig` you introduce can also be used for `send` (and to remove `send_many`)? If no, can `TxParams` and `TxCreationConfig` be consolidated?

Merging `send` and `send_to_many` and using `TxCreationConfigUnionTxParams` for both `send` and `create_transaction` can then be done in a future PR.

[^params]: I'm not counting arguments `Context` and `rpc_auth::Token` as those seem to be additions more recent than the tip of `offchain_serialized_pr`.

aszepieniec:
> Great change. 👍 I like the interface simplification.
> 
> I have also skimmed the branch Dan mentions in an earlier comment. The single `send` function with one parameter of type `TxParams` seems a lot cleaner than having both `send` and `send_to_many` with a list of 5 and 4 parameters, respectively.[1](#user-content-fn-params-87ddbffa3e9bafd1a5f3f2215043f372) I'm aware you're not changing the `send` RPC in this PR; I mention it because Dan's branch also simplifies the interface for `create_transaction` by using the same `TxParams`. I'm wondering whether the `TxCreationConfig` you introduce can also be used for `send` (and to remove `send_many`)? If no, can `TxParams` and `TxCreationConfig` be consolidated?
> 
> Merging `send` and `send_to_many` and using `TxCreationConfigUnionTxParams` for both `send` and `create_transaction` can then be done in a future PR.
> ## Footnotes
> 
>     1. I'm not counting arguments `Context` and `rpc_auth::Token` as those seem to be additions more recent than the tip of `offchain_serialized_pr`. [↩](#user-content-fnref-params-87ddbffa3e9bafd1a5f3f2215043f372)

I had a look as well and came away with the impression that `TxParams` is about capturing the payload of the transaction, not the configuration options for the process that creates it. If so, it should be compared with `TransactionDetails` on current `master` / this PR.

aszepieniec:
Ready for review @jan-ferdinand  @Sword-Smith @dan-da.

src/models/state/mod.rs
```diff
@@ -725,88 +730,75 @@ impl GlobalState {
...
-                anyhow::anyhow!("overflow subtracting total_spend from input_amount")
-            })?;
-
-            let change_utxo =
-                self.create_change_output(amount, change_key, change_utxo_notify_medium)?;
-            tx_outputs.push(change_utxo.clone());
-            maybe_change_output = Some(change_utxo);
-        }
+        // Add change output, if required to balance transaction
+        let nonzero_change = total_cost < total_unlocked;
```
jan-ferdinand, line 776: `nonzero_change` sounds like it's of type `NonZero<Change>` or similar.

```suggestion
        let change_is_nonzero = total_cost < total_unlocked; // or “is_nonzero_change”
```

src/models/state/tx_creation_artifacts.rs
```diff
@@ -0,0 +1,16 @@
+use crate::models::blockchain::transaction::Transaction;
+use crate::models::state::transaction_details::TransactionDetails;
+use crate::models::state::wallet::transaction_output::TxOutput;
+
+#[derive(Debug, Clone)]
+pub(crate) struct TxCreationArtifacts {
+    pub(crate) transaction: Transaction,
+    pub(crate) details: Option<TransactionDetails>,
```
jan-ferdinand, line 8: Maybe add a (dev-)comment explaining why this isn't always present? For example, do the creation artifacts go over the wire and this saves bandwidth?

src/rpc_server.rs
```diff
@@ -2013,6 +2011,11 @@ impl NeptuneRPCServer {
                 return Err(e.into());
             }
         };
+        let mut transaction = transaction_creation_artifacts.transaction;
+        let transaction_details = transaction_creation_artifacts
+            .details
+            .expect("details should be some when configured to track details");
+        let maybe_change_output = transaction_creation_artifacts.change_output;
         drop(state);
```
jan-ferdinand, line 2019: Can you drop the state earlier than this? It seems like it's not needed in the added lines.

src/models/state/wallet/wallet_state.rs
```diff
@@ -1656,25 +1669,11 @@ impl WalletState {
...
-            bail!(
-                "Insufficient funds. Requested: {}, Available: {}",
-                total_spend,
-                confirmed_available_amount_without_mempool_spends,
-            );
-        }
-
+        // Start selecting inputs.
+        // If there aren't enough inputs, we will discover after the next loop.
+        // We handle this error gracefully where we catch it, if we catch it.
```
jan-ferdinand, line 1674: Maybe:

```suggestion
        // If there aren't enough inputs, we will discover after the next loop
        // and handle the corresponding error gracefully where we catch it.
```

src/models/state/wallet/wallet_state.rs
```diff
@@ -1693,37 +1694,49 @@ impl WalletState {
...
-            "UTXO allocation failed. This should not be possible. Requested: {}, Available: {}",
-            total_spend,
-            confirmed_available_amount_without_mempool_spends,
-        );
+        // If there aren't enough funds, catch and report error gracefully
+        if allocated_amount < total_spend {
+            bail!(
+                "UTXO allocation failed.\n\
+                Requested: {}\n\
+                Avaialble, ignoring mempool: {}\n\
```
jan-ferdinand, line 1733:
```suggestion
                Available, ignoring mempool: {}\n\
```

src/models/state/wallet/wallet_state.rs
```diff
@@ -1693,37 +1694,49 @@ impl WalletState {
...
-        // check that also takes mempool into account.
-        assert!(
-            allocated_amount >= total_spend,
-            "UTXO allocation failed. This should not be possible. Requested: {}, Available: {}",
-            total_spend,
-            confirmed_available_amount_without_mempool_spends,
-        );
+        // If there aren't enough funds, catch and report error gracefully
+        if allocated_amount < total_spend {
+            bail!(
```
jan-ferdinand, line 1730: Maybe, but that's definitely personal preference and not an outright recommendation:

```rust
            bail!(
                "UTXO allocation failed.\n\
                Requested: {total_spend}\n\
                Available, ignoring mempool: {total_available_amount_ignoring_mempool}\n\
                Allocated: {allocated_amount}",
            )
```

jan-ferdinand (approved):

jan-ferdinand:
I have left some comments, questions, and nits inline. I'm aware that some of them are purely personal preference and I don't expect the opinionated parts to be changed. :slightly_smiling_face:

src/main_loop.rs
```diff
@@ -433,7 +434,7 @@ impl MainLoopHandler {
     /// Sends the result back through the provided channel.
     async fn update_mempool_jobs(
         update_jobs: Vec<UpdateMutatorSetDataJob>,
-        job_queue: &TritonVmJobQueue,
```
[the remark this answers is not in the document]
aszepieniec: 28ae478:

> The job queue was already being passed everywhere by reference. This
commit replaces the `&` reference with an `Arc<..>`.
> 
> Having an `Arc` reference is more conducive to storing the reference,
for instance in configuration objects such as `TxCreationConfig`.
Previously, that struct required complex lifetime annotations.
Moreover, the reference there was already (and still) wrapped in an
`Option` but because of the difficulty of storing `&`-reference, there
was no way of defaulting to a sensible default in case of `None`.
> 
> As a consequence, this change obviates the need to supply the `dummy`
job queue in all tests that use that. Less lines yay!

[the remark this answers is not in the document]
jan-ferdinand: Thanks for the pointer. :+1:

src/models/state/mempool.rs
```diff
@@ -1800,22 +1797,20 @@ mod tests {
             )
             .await;
             let in_seven_months = genesis_block.kernel.header.timestamp + Timestamp::months(7);
-            let (tx_by_bob, _, _maybe_change_output) = bob
+            let config = TxCreationConfig::default()
+                .recover_change_on_chain(bob_spending_key.into())
+                .with_prover_capability(proof_type);
+
+            // Clippy is wrong here. You can *not* eliminate the binding.
+            #[allow(clippy::let_and_return)]
```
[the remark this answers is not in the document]
jan-ferdinand: You need to annotate the `fn` instead:

```rust
#[expect(clippy::let_and_return)]
fn foo() -> T {
    let x = T::new();
    x
}

File: src/models/blockchain/transaction/lock_script.rs [hunk 9]
```diff
@@ -195,6 +185,18 @@ mod test {
     use crate::models::blockchain::transaction::primitive_witness::PrimitiveWitness;
     use crate::models::blockchain::type_scripts::native_currency_amount::NativeCurrencyAmount;
 
+    impl LockScript {
+        /// A lock script that is guaranteed to fail
+        #[cfg(test)]
```
jan-ferdinand, line 190: You can drop this now that it lives in the test module.

```suggestion
```

src/models/state/tx_creation_config.rs
```diff
@@ -139,3 +112,34 @@ impl TxCreationConfig {
...
 }
+
+#[cfg(test)]
+pub(crate) mod text {
+    use super::*;
+
+    impl TxCreationConfig {
+        /// Enable change-recovery with the given key, and set the medium to
+        /// `OnChain`.
+        #[cfg(test)]
```
jan-ferdinand, line 123: You can drop these now that the functions live in the the module.

dan-da:
> I had a look as well and came away with the impression that `TxParams` is about capturing the payload of the transaction, not the configuration options for the process that creates it. If so, it should be compared with `TransactionDetails` on current `master` / this PR.

TxParams in the cited branch is primarily about breaking tx creation into two steps:

1. select tx inputs and outputs, possibly including change.   eg, generate_tx_params()
2. actually create the transaction.  create_transaction()

In my view, this is a more flexible and better approach because:

1. implementation of create_transaction() becomes simpler.  It no longer needs to select input or calc change.
2. the logic of selecting inputs and calcing change is out-sourced to the caller, which then can use any logic it wants for both input selection and change.  generate_tx_params() is provided as an easy to use helper that satisfies common usages, but caller can impl their own logic instead.
3. create_transaction() is agnostic as to the type of Coin used in the inputs and outputs.  meaning that it can be used for other tokens, smart-contracts, etc.   (this may be true of present PR as well, I haven't had time yet to review closely).
4. points (3) and (4) extend to the RPC API as well.    So there is a consistent approach to tx creation at both layers.

devs familiar with bitcoin's RPC API and its history would immediately grok the approach.  bitcoin-core initially had only the `send` RPC call which was easy to use but limited.  ecosystem participants demanded more flexibility/control over tx creation, so around 2013-2014 iirc `sendrawtransaction` was added along with `createrawtransaction` and `signrawtransactionwithkey`.   These allowed RPC clients full control over all tx details, enabling things like input selection, multisig, advanced scripts, full fee control, etc.

So the changes I made were aimed at providing neptune-core RPC clients that level of flexibility from the very beginning rather than with ad-hoc changes down the road.

sadly, that part of the PR never got merged, and we are already implementing changes after launch.  It would be nice if we can make the APIs flexible enough now, so we don't have to do it again, possibly breaking 3rd party apps in the process.

From what I can tell the present PR is a nice improvement over the master branch, but does not yet enable all the above points.  Thus, I would advocate further refactoring to ensure they are all addressed.

dan-da:
My previous PR did *not* include any equivalent of bitcoin-core's `signrawtransaction`.   We may wish to consider something like that for a future PR.   Here's an AI generated summary of its purpose:  

[signrawtransaction.md](https://github.com/user-attachments/files/19235230/signrawtransaction.md)

Additionally it occurs to me that create_transaction() should accept a proof argument, meaning the proof would be created prior to calling create_transaction().

We would of course provide a neptune-core API (and RPC endpoint) to generate a proof from `TransactionDetails`.

This would enable clients to generate proofs out-of-process, possibly even on different hardware.

So as I see it, create_transaction() signature should be something like

```
pub async fn create_transaction(
        &self,
        tx_params: TxParams,
        proof:  Proof,
        timestamp: Timestamp,
    ) -> Result<Transaction>
```

notes:

1. removed pub(crate).    This should be a public API.
2. replaced mut tx_outputs with immutable tx_params: TxParams.
3. removed fee.   included in TxParams.
4. remove tx_creation_config.   Something like it could be an input to creation of TxParams.
5. changed Result from TransactionArtifacts to Transaction.  (If it doesn't return a `Transaction` it should not be named create_transaction)


In short, this PR is a concept/design nack from me as it stands.    Doubtless it has improvements over master and is a starting point, but I think we need to iterate on the API more until we have something really clean for the API, whittled down to its essence.

The above fn signature is a proposed starting point.   With these sorts of things, it is sometimes helpful to begin with definining what the set of API endpoints should be, and their signatures.  Ie, what is the cleanest way we can express this functionality for the caller.   Start with primary endpoints and signatures, then implement bodies and supporting fns.

note that even cleaner would be to break tx creation out of GlobalState into its own module.  I'm not advocating we go that far necessarily, but would be happy if we did.

src/models/state/tx_creation_config.rs
```diff
@@ -139,3 +112,34 @@ impl TxCreationConfig {
         self.proof_job_options.clone()
     }
 }
+
+#[cfg(test)]
+pub(crate) mod text {
```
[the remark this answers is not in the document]
aszepieniec: Right! So `texts` it is then.

aszepieniec:
> In short, this PR is a concept/design nack from me as it stands. Doubtless it has improvements over master and is a starting point, but I think we need to iterate on the API more until we have something really clean for the API, whittled down to its essence.

In general, I do not subscribe to the principle that the perfect should be the enemy of the good. If this PR constitutes an improvement over current master, then I would like to get it merged -- *even if it falls short of a comprehensive rearchitecture addressing all the issues of the in-scope modules*. I would like it to avoid merging it if
 - it does not constitute an improvement, or
 - (implied, but specifically) it makes a future step towards an architecture that does comprehensively address all issues more difficult.

That question will be in the back of my mind as I engage in this discussion: to merge or not to merge. However, I'm happy to use this discussion to also jumpstart and fuel an interactive design process for finding the right comprehensive architecture.

> TxParams in the cited branch is primarily about breaking tx creation into two steps:
>
> -   select tx inputs and outputs, possibly including change. eg, generate_tx_params()
> -   actually create the transaction. create_transaction()

> Additionally it occurs to me that create_transaction() should accept a proof argument, meaning the proof would be created prior to calling create_transaction().

I'm sorry, but if selecting transactions is not part of transaction creation, producing change is not either, and nor is the proof, then what's left?

> In my view, this is a more flexible and better approach because:

I think I do not understand which two things you are comparing. Is it `TxParams` versus `TransactionDetails`, or one-shot versus two-stage? Because as far as I can tell, the two-stage approach works with both `TxParams` and `TransactionDetails` and is already present in the codebase -- see `create_raw_transaction`.

> removed pub(crate). This should be a public API.

Fun fact. Prior to this PR, `create_transaction` was marked `pub` and was not being called at all, anywhere. Instead, `create_transaction_with_prover_capability` was used. If that function had been marked `pub(crate)` instead, the orange squiggly lines and CI errors would have clued us in earlier to the fact that there was something off with this design.

I'm not arguing to keep the `pub(crate)` marker, but instead cautioning against prematurely exposing functions. Let's get the interface right first -- and use the usage pattern within the codebase as a gauge for this metric -- and then expose it.

> changed Result from TransactionArtifacts to Transaction. (If it doesn't return a Transaction it should not be named create_transaction)

Well then certainly it shouldn't return a *triple* of which only the *first* item is a `Transaction`. Compared to that abhorrence, returning `TxCreationArtifacts` from a function called `create_transaction` is a minor source of confusion (*and* an improvement).

Changing the output type to `Transaction` is cleaner yes but it doesn't address the cases where more information is needed. But I think you're on to something here, especially combined with the two-stage approach. Specifically, currently the optional fields on `TxCreationArtifacts` are:
 - the change `TxOutput`, which the wallet needs to expect to recover;
 - the `TransactionDetails`, but in fact I think it's only needed by a test.

Two-stage transaction creation can address this funkiness like so:
 - The change `TxOutput` is created before the call to `create_transaction`, and retained across the call.
 - The `TransactionDetails` are the input to `create_transaction`, and so they are just cloned and the clone is retained across the call.

What I haven't figured out yet is how to avoid passing the chore of producing these objects onto the caller. Right now the caller calls one (appropriately configured) function and then receives everything he needs.

> note that even cleaner would be to break tx creation out of GlobalState into its own module. I'm not advocating we go that far necessarily, but would be happy if we did.

Agreed!

dan-da:
> > In short, this PR is a concept/design nack from me as it stands. Doubtless it has improvements over master and is a starting point, but I think we need to iterate on the API more until we have something really clean for the API, whittled down to its essence.
> 
> In general, I do not subscribe to the principle that the perfect should be the enemy of the good. If this PR constitutes an improvement over current master, then I would like to get it merged -- _even if it falls short of a comprehensive rearchitecture addressing all the issues of the in-scope modules_. I would like it to avoid merging it if

I see both sides on this.  While I agree with the general principle, I also think that now we are on mainnet we need to begin establishing a culture of avoiding change to our public API unless absolutely necessary, and if we are going to change it, then its better to take our time and get it right, so as to avoid breaking 3rd parties.

Admittedly, the community is small now so we have some flexibility, but its generally not good practice to be changing such an important API in one release after another.  So that is a counterargument, for consideraiton.

> > TxParams in the cited branch is primarily about breaking tx creation into two steps:
> > 
> > * select tx inputs and outputs, possibly including change. eg, generate_tx_params()
> > * actually create the transaction. create_transaction()
> 
> > Additionally it occurs to me that create_transaction() should accept a proof argument, meaning the proof would be created prior to calling create_transaction().
> 
> I'm sorry, but if selecting inputs is not part of transaction creation, producing change is not either, and nor is the proof, then what's left?

taking the components that have been built in prior steps and actually creating a `Transaction`.    Perhaps it could just be Transaction::new() instead of involving GlobalState.

> > In my view, this is a more flexible and better approach because:
> 
> I think I do not understand which two things you are comparing. Is it `TxParams` versus `TransactionDetails`, or one-shot versus two-stage? Because as far as I can tell, the two-stage approach works with both `TxParams` and `TransactionDetails` and is already present in the codebase -- see `create_raw_transaction`.

one-shot vs two-stage. 

I just took a look at create_raw_transaction() in the PR and it is indeed closer to what I'm suggesting, with these caveats:

1. the method is pub(crate) rather than pub.  (minor)

2. it accepts a TxCreationConfig arg, which has some un-necessary things:    eg if the caller has already obtained TransactionDetails then change Utxo is already present and change_policy is irrelevant.  and if we modify the API to accept a Proof arg (as I suggest), then the various proof-generation options also become unnecessary.  In short, I don't see any need for TxCreationConfig, once we boil the API down to its essence.

> > removed pub(crate). This should be a public API.
> 
> Fun fact. Prior to this PR, `create_transaction` was marked `pub` and was not being called at all, anywhere. Instead, `create_transaction_with_prover_capability` was used. If that function had been marked `pub(crate)` instead, the orange squiggly lines and CI errors would have clued us in earlier to the fact that there was something off with this design.

not being called anywhere in our crate is arguably fine (so long as tests are written for it).  It can still be called by third parties using the crate.

However, I will not argue that the code in master is/was elegant.  Rather I'm not a fan and indeed I did a lot of work to improve it in the previous discussed PR.  So I prefer to discuss difference with the code in that PR as I won't defend what is in master now.

> I'm not arguing to keep the `pub(crate)` marker, but instead cautioning against prematurely exposing functions. Let's get the interface right first -- and use the usage pattern within the codebase as a gauge for this metric -- and then expose it.

I can agree with that up to a point.   I still think it is a useful exercise though to think about things from the RPC client perspective as a starting point.  Eg, if I'm writing an exchange or a wallet, or a smart contract, or a multi-token DEX, what do I need?     And for me, the answer is the same for all those use cases:  I need a way to have full control over inputs, outputs, and fee.  And even proof generation may be desirable.

> > changed Result from TransactionArtifacts to Transaction. (If it doesn't return a Transaction it should not be named create_transaction)
> 
> Well then certainly it shouldn't return a _triple_ of which only the _first_ item is a `Transaction`. Compared to that abhorrence, returning `TxCreationArtifacts` from a function called `create_transaction` is a minor source of confusion (_and_ an improvement).

If you look at the branch with the refactoring I did, it not return a triple.  It was very simple and clean:

```
pub async fn create_transaction(&self, tx_params: TxParams) -> Result<Transaction>
```

It has been a source of frustration to me what has happened since.   So again, I won't defend present master, and let's not compare with it further.

> Changing the output type to `Transaction` is cleaner yes but it doesn't address the cases where more information is needed.

Logically, if the method is "create_transaction", why should any more information be needed in the result?   If it is, that seems to point to an API design problem.

> But I think you're on to something here, especially combined with the two-stage approach. Specifically, currently the optional fields on `TxCreationArtifacts` are:
> 
> * the change `TxOutput`, which the wallet needs to expect to recover;
> * the `TransactionDetails`, but in fact I think it's only needed by a test.
> 
> Two-stage transaction creation can address this funkiness like so:
> 
> * The change `TxOutput` is created before the call to `create_transaction`, and retained across the call.

exactly!  that's what I've been trying to get across.   It was a light-bulb moment for me when I realized that.  And I iterated on the API several tmes before settling on what is in the branch. 

> * The `TransactionDetails` are the input to `create_transaction`, and so they are just cloned and the clone is retained across the call.

That seems about right.  TransactionDetails appears pretty close to [my TxParams](https://github.com/Neptune-Crypto/neptune-core/blob/70f40bce44655c901db5c1459a1c3bff83b2de74/src/models/blockchain/transaction/transaction_params.rs#L50C12-L50C20), but with addition of `MutatorSetAccumulator`, which iirc became necessary after "real proving" got merged in.

note that TxParams does not have a fee field, because it is calculated as sum(inputs) - sum(outputs).  If fee field is included, then it just needs to be validated against this.

TxParams does not include a coinbase field because it is intended for usage by external callers, and didn't appear to me to make any sense in that context.   I would suggest that our eventual `pub` create_transaction() or create_raw_transaction() use a datatype that does not include a coinbase field, *unless* we are expecting that external mining software might need to call create_transaction, and I suppose that makes sense... hmm.   so maybe it's fine then.

> What I haven't figured out yet is how to avoid passing the chore of producing these objects onto the caller. Right now the caller calls one (appropriately configured) function and then receives everything he needs.

Take a look at generate_tx_params() which I linked in https://github.com/Neptune-Crypto/neptune-core/pull/491#issuecomment-2705293139.

The point of that method is to perform the heavy lifting of selecting inputs, making change output, etc.   And it can handle the common cases.  Basically I extracted and refactored the "stage one" part of create_transaction() into create_tx_params().    But the key point is that it is *optional* for the caller to use it.   They can perform their own logic if they need to deviate from behavior of this method.    or also, we could provide other methods for other use cases, eg sending tokens.

What I realized as I was writing my earlier comment is that we should be able to apply the same model to Proof generation.   Ie, we provide an easy to use, but *optional* API for proof-generation.   More advanced clients may wish to perform the proof generation themselves.

jan-ferdinand:
I like where this discussion is going. It seems like there's at least one open question regarding requirements of the “create transaction” API, namely whether it should be able to handle coinbase UTXOs. I think it's a good idea to figure out the requirements as early as possible, since changing requirements usually lead to updates in the design. I'm aware this is usually not a linear process. Are there other open questions with regards to the requirements?

aszepieniec:
> >   I'm sorry, but if selecting inputs is not part of transaction creation, producing change is not either, and nor is the proof, then what's left?
>
> taking the components that have been built in prior steps and actually creating a Transaction. Perhaps it could just be Transaction::new() instead of involving GlobalState.

Sure, if we are considering only the step where existing data is rearranged into an object of the right type, then a thin constructor called `new` is preferable or even struct literal syntax is. That limited focus does not contribute to the discussion about how to organize the rest of the pipeline which includes how the data in question is being produced. If you want the label "transaction creation" to denote only the object construction step, that's fine but then please suggest an alternative label that captures the rest of the pipeline -- which the function called `create_transaction` right now does.

In particular, if you want to suggest refactoring `create_transaction` so that it takes a proof as a given argument, then you're implicitly suggesting a new interface to produce proofs. Please shed light on that. Likewise, if you are suggesting refactoring `create_transaction` so that it doesn't need to produce a change UTXO or select inputs, then you are implicitly suggesting that these tasks are done by some other bodies of code, and please shed light on how their interfaces would work. Well I guess that [`generate_tx_params`](https://github.com/Neptune-Crypto/neptune-core/blob/70f40bce44655c901db5c1459a1c3bff83b2de74/src/rpc_server.rs#L728) is a very concrete proposal (that's great, let's discuss that) but the point is that these factored out methods are an integral part of the question and cannot be abstracted over.

> `TransactionDetails` appears pretty close to [my `TxParams`](https://github.com/Neptune-Crypto/neptune-core/blob/70f40bce44655c901db5c1459a1c3bff83b2de74/src/models/blockchain/transaction/transaction_params.rs#L50C12-L50C20), but with addition of `MutatorSetAccumulator`, which iirc became necessary after "real proving" got merged in.

I'm not married to the name `TransactionDetails` by the way, although I'm not sure `TxParams` is an improvement. Now would be a good time to suggest alternatives.

Distilling all the above into a suggested way forward, how is this interface?

```rust
// we have tx_outputs : TxOutputList, change_policy : ChangePolicy, proof_config
let total_outbound_amount : NativeCurrencyAmount = tx_outputs.total_native_currency_amount();
let tx_inputs = global_state.select_inputs(total_outbound_amount)?;
let total_unlocked_amount : NativeCurrencyAmount = tx_inputs.total_native_currency_amount();
let change_amount = total_unlocked_amount.checked_sub(total_outbound_amount).expect("guaranteed to be positive because otherwise `select_inputs` fails");
let change_output = if change_amount == 0 { None } else {
    let output  = TxOutput::create_change(change_amount, change_policy);
    tx_outputs.push(output.clone())
    Some(output)
};
let transaction_details = TransactionDetails::new(tx_inputs, tx_outputs);
let proof = transaction_details.prove(proof_config);
let transaction = Transaction::new(transaction_details, proof);
```

dan-da:
> In particular, if you want to suggest refactoring `create_transaction` so that it takes a proof as a given argument, then you're implicitly suggesting a new interface to produce proofs. Please shed light on that. Likewise, if you are suggesting refactoring `create_transaction` so that it doesn't need to produce a change UTXO or select inputs, then you are implicitly suggesting that these tasks are done by some other bodies of code, and please shed light on how their interfaces would work. Well I guess that [`generate_tx_params`](https://github.com/Neptune-Crypto/neptune-core/blob/70f40bce44655c901db5c1459a1c3bff83b2de74/src/rpc_server.rs#L728) is a very concrete proposal (that's great, let's discuss that) but the point is that these factored out methods are an integral part of the question and cannot be abstracted over.

At a low level, they can be abstracted over, and that is the point.  Consider:

At core, what I am suggesting is that the caller has complete freedom to create the TxParams however they wish, and (now) the Proof as well.   So all that our public API _truly needs to care about_ is that we receive a valid `Transaction` object when sending/broadcasting.

Ok, so once we have that level of flexibility in the public interface as our starting point, then we can provide higher level helpers or builders for the common cases where caller does not need such a high degree of flexibility and just wants simplicity.

Tmw I can dedicate some time to creating a specific API proposal to provide something more concrete to discuss, especially with regards to creating a proof, and also the RPC APIs. 

> ```rust
> // we have tx_outputs : TxOutputList, change_policy : ChangePolicy, proof_config
> let total_outbound_amount : NativeCurrencyAmount = tx_outputs.total_native_currency_amount();
> let tx_inputs = global_state.select_inputs(total_outbound_amount)?;
> let total_unlocked_amount : NativeCurrencyAmount = tx_inputs.total_native_currency_amount();
> let change_amount = total_unlocked_amount.checked_sub(total_outbound_amount).expect("guaranteed to be positive because otherwise `select_inputs` fails");
> let change_output = if change_amount == 0 { None } else {
>     let output  = TxOutput::create_change(change_amount, change_policy);
>     tx_outputs.push(output.clone())
>     Some(output)
> };
> let transaction_details = TransactionDetails::new(tx_inputs, tx_outputs);
> let proof = transaction_details.prove(proof_config);
> let transaction = Transaction::new(transaction_details, proof);
> ```

I like the direction.

So expanding on what I said above I'm thinking of multiple levels of abstraction.  eg:

1. raw level.   Just give us a `Transaction`
2. low level.  what create_tx_params() and create_tx_proof() would use internally (or any callers that need to construct a `Transaction` directly)
3. common-usage level.  call create_tx_params(), create_tx_proof(), Transaction::new(), send()
4. simple_send().   just provide a receiving address, amount, fee.

The above seems to represent (2) and seems reasonable for that level.

Tmw, I think I will spend some time designing first a set of RPC APIs (fn signatures) around these levels.  I think its helpful to visualize how it can be used by clients.   The RPC APIs would exist for levels 1, 3, and 4 for sure.  Level 2 I'm not sure about.  perhaps not, for now at least, though it may be necessary for RPC clients to do advanced things, so its an area for more analysis.

src/models/state/tx_creation_config.rs
```diff
@@ -0,0 +1,145 @@
...
+        self
+    }
+
+    /// Configure the proving capacity.
+    pub(crate) fn with_prover_capability(mut self, prover_capability: TxProvingCapability) -> Self {
+        self.prover_capability = prover_capability;
+        self
+    }
+
+    /// Configure which job queue to use.
```
dan-da, line 61: "which job queue to use?"

The intent here (multiple queues) appears to be at odds with the job-queue design.

the purpose of the job-queue is/was to have a *single* serialized queue so that proving jobs do not compete with eachother for machine resources (cpu, ram).

src/models/state/tx_creation_config.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /// Get the transaction proving capability.
+    pub(crate) fn prover_capability(&self) -> TxProvingCapability {
+        self.prover_capability
+    }
+
+    /// Get (a smart pointer to) the job queue, or create a new job queue and
+    /// return a pointer to that.
+    pub(crate) fn job_queue(&self) -> Arc<TritonVmJobQueue> {
```
dan-da, line 104: there is already a TritonVmJobQueue in GlobalStateLock.  This would create a second job queue instance.   That then permits 2 triton-vm jobs to execute simultaneously.   Which the job-queue was designed to prevent.

Unless something fundamental has changed with regards to our resource requirements, I think this is problematic and should probably be removed, so we have only a single vm job queue.

Remember also that the job-queue is used when composing and for upgrade jobs.  not only during tx initiation.

If this were to be maintained, I think it requires justification/rationale and a description of how it has been tested on different hardware configs while simultaneously composing, etc, etc.

src/models/state/tx_creation_config.rs
aszepieniec: Then why do we pass around a pointer to the job queue through a bunch of function interfaces? If the purpose is to have a single serialized queue that guarantees that jobs do not compete with each other for resources, then surely a static `OnceLock<TritonVmJobQueue>` is a more fitting solution?

src/models/state/tx_creation_config.rs
aszepieniec: I think the right solution is to refactor the interface to the `TritonVmJobQueue`.

That said, I'm not sure this criticism is accurate. I'm never building two instances of the job-queue; merely copying and storing references to the same instance.

src/models/state/tx_creation_config.rs
dan-da: It might be more fitting.  The fact remains the raison d'être for the job-queue is and remains serializing jobs running in the VM so as not to exceed available resources.

It's possible we might want to change that at some point, but if so it should be done carefully.

src/models/state/tx_creation_config.rs
dan-da: > I think the right solution is to refactor the interface to the TritonVmJobQueue

Afaik, the job-queue interface is just fine.   ???

I'm _guessing_ from another comment you mean how the job-queue handle gets passed around inside neptune-core?   I agree that could use improvement, perhaps the static OnceLock trick, but that is a matter of usage by the job-queue's caller(s), not of the job-queue interface itself.

It would be like saying the HashMap interface needs to be refactored because we pass a HashMap around all over the place.

> I'm never building two instances of the job-queue; merely copying and storing references to the same instance.

But your code is creating a new instance.  And an instance already exists in GlobalStateLock.  So the one created here becomes the second instance.

```
    /// Get (a smart pointer to) the job queue, or create a new job queue and
    /// return a pointer to that.
    pub(crate) fn job_queue(&self) -> Arc<TritonVmJobQueue> {
        self.triton_vm_job_queue
            .as_ref()
            .cloned()
            .unwrap_or_else(|| Arc::new(TritonVmJobQueue::start()))
    }
```

See GlobalStateLock::new().

dan-da:
> Tmw, I think I will spend some time designing first a set of RPC APIs

first draft: https://github.com/Neptune-Crypto/neptune-core/issues/510#issuecomment-2731749143

src/models/state/tx_creation_config.rs
aszepieniec: > Afaik, the job-queue interface is just fine. ???
>
> I'm guessing from another comment you mean how the job-queue handle gets passed around inside neptune-core? I agree that could use improvement, perhaps the static OnceLock trick, but that is a matter of usage by the job-queue's caller(s), not of the job-queue interface itself.

I was tempted to agree that the issue is isolated to how the handle gets passed around, but then I wanted to prove you wrong about two instances being created. Long story short, you are right. Two instances *are* being created, but it's not obvious. One instance is spawned in `GlobalStateLock::new` and one in `create_transaction_from_data_worker` via the `config.job_queue()` call because at this point the config does not store an instance so it creates one on the fly. So if you initiate a transaction while upgrading someone else's transaction or while composing, you will generate two proof jobs that simultaneously compete for resources. Good catch!

Circling back, armed with this knowledge, to the point about the job-queue interface ... I have to say the interface is not fine. The intention is that only one instances of this type ever exists. Therefore, the interface should be such that messy implementations (such as the one I am responsible for) that accidentally generate two instances are impossible.

I propose to resolve it as follows:
 - [ ] on a new branch and in a new PR
 - [ ] make `start` private
 - [ ] use the `OnceLock` trick
 - [ ] store instances of the job queue nowhere, except once in the static memory location designated for it
 - [ ] eliminate or reduce the percolation of passed references

aszepieniec: Cfr. #519

Sword-Smith:
A real-world need is the ability to restrict the number of inputs that a new transaction is being built with. If the transaction-initiator is `ProofCollection`-capable, they cannot expect other nodes to upgrade their proofs if it contains too many inputs, as the resulting `SingleProof` program execution becomes too big. I consider padded height $2^{23}$ the practical limit with the computers that on the network today. The same, or perhaps an even smaller, limit applies when the initiating amchine is `SingleProof`-capable. Putting this limit into `WalletState` seems compatible with @dan-da's suggested transaction creation logic.

dan-da:
superceded by #544 which has been merged.  closing.

PR #492, opened 2025-03-06
Initial general-purpose RPC boilerplate/base

KaffinPX:
It's base of a general-purpose RPC over HTTP/JSON: users are supposed to handle their safety by proxies like nginx etc.
For now it is read-only but in future it should allow submitting txs and PoW(external guessers).
To be extended based on requirements...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -278,25 +278,59 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3b829e4e32b91e643de6eafe82b1d90675f5874230191a4ffbc1b336dec4d6bf"
 dependencies = [
  "async-trait",
- "axum-core",
+ "axum-core 0.3.4",
  "bitflags 1.3.2",
  "bytes",
  "futures-util",
- "http",
- "http-body",
- "hyper",
+ "http 0.2.12",
+ "http-body 0.4.6",
+ "hyper 0.14.32",
  "itoa",
- "matchit",
+ "matchit 0.7.3",
  "memchr",
  "mime",
  "percent-encoding",
  "pin-project-lite",
  "rustversion",
  "serde",
- "sync_wrapper",
- "tower",
+ "sync_wrapper 0.1.2",
+ "tower 0.4.13",
  "tower-layer",
  "tower-service",
+]
+
+[[package]]
+name = "axum"
+version = "0.8.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6d6fd624c75e18b3b4c6b9caf42b1afe24437daaee904069137d8bab077be8b8"
+dependencies = [
+ "axum-core 0.5.0",
+ "bytes",
+ "form_urlencoded",
+ "futures-util",
+ "http 1.2.0",
+ "http-body 1.0.1",
+ "http-body-util",
+ "hyper 1.6.0",
+ "hyper-util",
+ "itoa",
+ "matchit 0.8.4",
+ "memchr",
+ "mime",
+ "percent-encoding",
+ "pin-project-lite",
+ "rustversion",
+ "serde",
+ "serde_json",
+ "serde_path_to_error",
+ "serde_urlencoded",
+ "sync_wrapper 1.0.2",
+ "tokio",
+ "tower 0.5.2",
+ "tower-layer",
+ "tower-service",
+ "tracing",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


dan-da:
thanks for the PR.   I'm sure significant effort went into this, and its an interesting approach.

That said, I have concerns with this approach.

1.  It introduces a new RPC layer with its own RPC API calls instead of using the existing calls.  So it splits the rpc API instead of having a unified API.

2. It brings in large dependencies:  axum and tower.   I have been working to reduce dependencies where possible, so this seems opposite direction.

Some general background:

1. We have previously considered adding a json/rpc layer for example, but did not have bandwidth yet, primarily because other things had higher prioritiy, but also because it would need to be designed correctly to integrate with our existing RPC API.   (Or else we could switch entirely to a new RPC layer and deprecate the old API, but that's a big team decision).  

2. The RPC layer we already have, based on tarpc, uses JSON as the transport.  It can already be connected to by clients.   The issue is that non-rust clients would need to "speak" with tarpc to invoke the rpc calls.   We intend to build some sample code for this, but haven't had time yet.    Pull requests in that area would be most welcome.   A good starting point would be a rust client that invokes any existing neptune-core RPC without using the tarpc crate.  This could serve as an example for other languages.

-------

Stepping back a bit, I would say that nothing is finalized or in stone yet.   Probably the team will need to have a discussion about what we want to do with the RPC layer going forward, and taking the community (needs, feedback, contributions) into account.

dan-da:
So I should first say that my opinions are my own and it will ultimately be a team decision how we move forward with the RPC API.  I will bring this PR up for discussion at the next meeting.

> 1. taRPC api standard was heavily specialized -- like a specific call returning dashboard stats, this one is designed to be more like explorers api with dashboard call separated to multiple calls for better usage scenarios on 3. party applications(Just a boilerplate anyway, still can be standardized).

I think we should make a clear distinction between:
 1. The RPC protocol/mechanism.   (eg tarpc, json-rpc, REST, gRPC, etc)
 2. The API endpoints that are implemented with (1) for performing specific actions.

With this PR, I am primarily concerned that it is changing (1), by adding a new mechanism.

But above you are talking about (2).    We can add new endpoints as needed to our existing RPC mechanism, so long as they make sense.    A PR in that direction would be considered without concerns about (1), though I would suggest first opening an issue for each proposed API, in order to get feedback before spending time implementing.

> 2. Valid point, Axum costs around 169 KiB~, if we are talking about safety(supply chain attack etc.) its developed by tokio-rs which has a big reputation around community, but we can gate it on a build feature if wanted .

yeah, I like Axum and use it in another project actually.  It's just that neptune-core is already huge and as the project progresses,  I hope we can break it out into smaller components, remove more deps, etc, to make compiles faster, reduce supply-chain footprint, etc.

Build feature could be a possibility yes, but then there would be different versions of the binary with different RPC mechanisms.   Seems complex.

> 3. Current taRPC api cannot be exposed to public by default as it includes sensitive calls(like wallet management) and forcing developers to implement taRPC on other languages just doesn't make sense -- bad DX.

You raise a good point that we should consider a separate "no-auth" RPC layer for purely public information.  It's something I've thought about a little, but haven't devoted any effort towards yet.

My initial thought is that even if we separate the API into "private" and "public", it would still be best to have a unified endpoint layer.  By this, I mean that that all endpoints are accessible by all RPC mechanisms.    Forget tarpc for the moment.  In the past I have implemented systems that supported both REST and json/rpc, for example.   I first implemented a transport-agnostic layer that has the core logic of each endpoint.  Then I added REST and json/rpc endpoints as super-thin wrappers around the core logic endpoint.   I think we could do something similar here, to continue with tarpc for rust clients, but also possibly add another mechanism that is more cross-language compat.

Also, we should consider that the present API has some endpoints that can be considered "public", eg "network", "block_height", "latest_tip_digest", "block_info", etc.    Separating into public and private *might* be as simple as removing the auth requirement for the "public" endpoints, though I might be forgetting some attack vector here.

It could also mean listening on a different port for public than private.   But still in that case, all endpoints should be usable via all supported RPC mechanisms (at each port).

> I think it still makes sense to have a new transport method with a better and more general standard could be viable for improving DX.

yeah, I struggle a little because early on I was pushing for this, specifically to switch to json-rpc (like bitcoin-core).  But there was never a team consensus to do that.  So I've made peace with tarpc, and the suggested path towards multi-language support has been that we demonstrate how a client can "speak" to neptune-core using sockets without using the tarpc crate.  And possibly we could make available some libs for other langs.

However... at this point, it may be less work just to impl an mechanism agnostic endpoint layer (as described above) and then add one or more widely used rpc mechanisms.    At least this is the approach I'm leaning towards, though there may be some unforseen complications.

KaffinPX:
I've got a better idea, why don't we replace taRPC with IPC and build "RPC"s outside of core?

PR #499, opened 2025-03-12
refactor: Use mempool to avoid double-spends

Sword-Smith:
Drop the field `selection` from `TxCreationArtifacts` as the task of keeping track of unconfirmed spent UTXOs can be handled by the mempool

### DIFF
--- a/src/models/blockchain/block/mod.rs
+++ b/src/models/blockchain/block/mod.rs
@@ -1017,8 +1017,6 @@ impl Block {
 
 #[cfg(test)]
 pub(crate) mod block_tests {
-    use std::collections::HashSet;
-
     use rand::random;
     use rand::rngs::StdRng;
     use rand::Rng;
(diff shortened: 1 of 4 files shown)


aszepieniec:
This PR does
 - reduce complexity
 - avoid `#[cfg(test)]` flags.

And based on those merits I merged it.

That said, there is accidental design happening here that I think we should synchronize about. Specifically, using the mempool is now the recommended as the way to avoid selecting the same UTXO across multiple smaller transactions. In other words, the mempool is used as a helper for a wallet process.

I thought of the mempool as a queue of broadcasted transactions that one may reasonably expect to be confirmed by the next blocks. In particular, whether a client runs a mempool or not is orthogonal to the question whether they run a wallet or not. In particular, you should be able to run a mempool without running a wallet, and you should be able to run a wallet without running a mempool. This conception seems to be incompatible with requiring the mempool's help when producing a bunch of transactions.

PR #502, opened 2025-03-13
fork!: Increase block size

Sword-Smith:
We discovered that the number of inputs allowed into a block had a much lower practical limit than we intended. The problem is that the `inputs` field of the `TransactionKernel` holds up to 45 authentication paths of the sliding-window Bloom filter MMR. If this MMR has a height of 20, which will be reached when the total number of outputs is ~8M, these 45 authentication paths take up 4.500 `BFieldElement`s. The block proof being ~125.000 `BFieldElement`s and the old block size limit 250.000, this only leaves room for 125.000 / 4.500 ~= 28 inputs! And that's assuming that the mutator set accumulator takes up no space, and that there are no outputs.

With this increase, the practical limit on the number of inputs per block rises to ~195, which is still quite low but should be enough for "some" time.

### DIFF
--- a/src/models/blockchain/block/block_height.rs
+++ b/src/models/blockchain/block/block_height.rs
@@ -35,6 +35,10 @@ pub const BLOCKS_PER_GENERATION: u64 = 160815;
 impl BlockHeight {
     pub const MAX: u64 = BFieldElement::MAX;
 
+    pub const fn new(value: BFieldElement) -> Self {
+        Self(value)
+    }
+
     pub fn get_generation(&self) -> u64 {
         self.0.value() / BLOCKS_PER_GENERATION
     }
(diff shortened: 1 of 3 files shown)


dan-da:
> With this increase

note:   old limit: 2Mb, new limit: 8 Mb.

Sword-Smith:
Questions received through another channel:
> What happens if you put even more, to have for example max 500 inputs? How do you plan to scale this?

We would have to change the representation of transaction inputs to handle this. Specifically, we would want to implement some version of the algorithm outlined here:
https://neptune.cash/learn/authentication-structure-authentication/

This "aggregation" of MMR authentication paths could take place on the block level, leaving it up to the composer to provide the node of the MMR and the proof that these digests are correct.

> what about also decreasing block time?

Current block target interval is 588 seconds. The fastest composer on the network takes about 185 seconds for a block composition. Guessers should be allowed to hash at least 50 % of the block target interval. Anything less reduces the security against reorganizations and will also lead to more orphaned blocks, as two guessers are more likely to find a solution close enough in time that both blocks are broadcast.

Sword-Smith:
Article about this PR: https://neptune.cash/blog/hard-fork-1/

PR #503, opened 2025-03-13
feat: Allow restriction of number of inputs per tx

Sword-Smith:
Allow the user to restrict the number of inputs per transaction to avoid having to deal with transactions that will likely never be mined, as they don't fit into a block.

The current default limit is set quite low, to 28 inputs per transaction due to the big size of transaction inputs. This limit is intended to rise significantly in the future, but for now it should serve to prevent composers and transaction initiators from making transactions that cannot be mined.

Both the wallet state and mempool state gets passed this value such that they, internally, can enforce the correct policies.


Sword-Smith:
Marked as draft because this new parameter is completely untested. Needs tests!

src/config_models/cli_args.rs
```diff
@@ -170,6 +170,21 @@ pub struct Args {
...
+    /// Restrict the number of inputs for transcations that this node deals
+    /// with.
+    ///
+    /// If this value is set, the node will not create transactions with more
+    /// inputs than this limit. It will also not accept such transactions from
+    /// peers into its mempool. And if the node is either upgrading transaction
+    /// proofs on the network or composing, it will never merge two transactions
+    /// that would exceed this limit. Does not affect block validity, i.e.
+    /// consensus. So a block with more than this number of inputs may be
+    /// considered value. Also does not affect which block proposals that a
```
aszepieniec, line 182:
```suggestion
    /// consensus: a block with more than this number of inputs can be
    /// valid. Also does not affect which block proposals that a
```

src/config_models/cli_args.rs
```diff
@@ -170,6 +170,21 @@ pub struct Args {
...
+    /// guessing node will attempt to solve the PoW puzzle for. Set to "none" to
+    /// disable this limit.
```
aszepieniec, line 184: If the default is not none, it should say here what it is.

src/config_models/cli_args.rs
```diff
@@ -170,6 +170,21 @@ pub struct Args {
...
+    #[clap(long, value_parser = parse_optional_usize, default_value = "28")]
```
aszepieniec, line 185: It is not obvious to me that the default string will parse correctly. Please add it to the test case below.

src/main_loop.rs
```diff
@@ -807,14 +807,18 @@ impl MainLoopHandler {
...
-                global_state_mut
+                if let Err(err) = global_state_mut
                     .mempool_insert(
                         pt2m_transaction.transaction.to_owned(),
                         TransactionOrigin::Foreign,
                     )
-                    .await;
+                    .await
+                {
+                    warn!("{err}");
```
aszepieniec, line 817: Add context:
```suggestion
                    warn!("cannot add transaction into mempool: {err}");
```

src/main_loop.rs
```diff
@@ -807,14 +807,18 @@ impl MainLoopHandler {
                 }
 
                 // Insert into mempool
```
aszepieniec, line 809:
```suggestion
                // Insert into mempool or return prematurely
```

src/models/state/mempool.rs
```diff
@@ -226,31 +238,48 @@ impl Mempool {
...
-                own_tx = own_tx || candidate.origin.is_own();
-            }
-
-            let origin = match own_tx {
-                true => TransactionOrigin::Own,
-                false => TransactionOrigin::Foreign,
-            };
+        /// Loop from `i`th most valuable transaction in the mempool to the
+        /// least valuable in order to find a pair whose total number of inputs
+        /// does not exceed the maximum value.
```
aszepieniec, line 251:
```suggestion
        /// Loop from `i`th densest transaction in the mempool to the
        /// least dense in order to find a pair whose total number of inputs
        /// does not exceed the maximum value.
```

src/models/state/mempool.rs
```diff
@@ -226,31 +238,48 @@ impl Mempool {
...
+        fn skip_first_n_transaction(
```
aszepieniec, line 252: Something is missing from this name to describe what the function does.

src/models/state/mempool.rs
```diff
@@ -346,13 +375,18 @@ impl Mempool {
             }
         }
 
-        let mut events = vec![];
+        // Check if tx conforms to mempool policy.
```
aszepieniec, line 378:
```suggestion
        // Check if tx conforms to mempool policy wrt # inputs
```

src/models/state/mempool/mempool_insertion_error.rs
```diff
@@ -0,0 +1,6 @@
+/// The reasons why inserting a transaction into the mempool may fail.
+#[derive(Debug, Clone, Copy, thiserror::Error)]
+pub(crate) enum MempoolInsertionError {
+    #[error("Too many inputs: got {got}, but only {allowed} are allowed.")]
+    TooManyInputs { got: usize, allowed: usize },
+}
```
aszepieniec, line 6: Too small and intricately tied to the mempool module to justify its own file.

aszepieniec (commented):
Basic structure looks good. I have some superficial comments. Before approving I would like to double-check against the list on the whiteboard and verify that the right logic is used in all of those cases.

src/models/state/mempool/mempool_insertion_error.rs
Sword-Smith: It might not be obvious from the code base but I rather like small files :)

src/config_models/cli_args.rs
Sword-Smith: `clap` fills that out automatically when you use e.g. `neptune-core --help`. See this example for how it works with `--max-num-proofs`:
```bash
      --max-num-proofs <MAX_NUM_PROOFS>
          Sets the maximum number of proofs in a `ProofCollection` that can be recursively combined into a `SingleProof` by this machine. I.e. how big STARK proofs this machine can produce
          
          [default: 16]

```

src/models/state/mempool/mempool_insertion_error.rs
jan-ferdinand: It's good to see more use of `thiserror` instead of `anyhow` in the library. I advocate for re-exporting all error types in a top-level `error.rs` to simplify discovery for downstream users.

src/models/state/mempool.rs
```diff
@@ -152,6 +162,7 @@ impl Mempool {
     pub fn new(
         max_total_size: ByteSize,
         max_num_transactions: Option<usize>,
+        max_num_inputs_per_tx: Option<usize>,
```
jan-ferdinand, line 165: This interface is starting to get unwieldy. How about a builder instead? If you don't want to write one yourself, take a look at [bon](https://crates.io/crates/bon).

src/models/state/mempool.rs
```diff
@@ -226,31 +239,55 @@ impl Mempool {
...
+                    };
+                    return Some((ret.try_into().unwrap(), origin));
+                }
             }
+
+            None
         }
 
-        None
+        (0..self.len() - 1).find_map(|i| most_dense_pair_skip_first_n(self, i))
```
Sword-Smith, line 290: I think the expression `self.len() - 1` can overflow. Probably not good.

Sword-Smith: Just shows that these getters need to be tested on an empty mempool. That would have caught this problem by producing a never-ending test case.

PR #506, opened 2025-03-16
Mining-related improvements

KaffinPX:


Sword-Smith (changes requested):
Thank you for the PR. I would like to see a few small changes before merging:
1. Add a test that the `prev_block` field is set correctly. This can be done with a one-liner in the test `pow_puzzle_is_consistent_with_block_hash`.
2. Fix trailing whitespace in docstring of `prev_block` field.
3. Fix max length of docstring for `prev_block` field. Max length for comments is 80 characters.
4. Add a short comment about why you want the struct `ProofOfWorkPuzzle` to have public fields. This doesn't have to be as a docstring, just a short comment somewhere inside or just above the struct `ProofOfWorkPuzzle`.

PR #508, opened 2025-03-17
Use `tokio::interval` over `sleep`

jan-ferdinand:
Note that [`Interval::tick`](https://docs.rs/tokio/latest/tokio/time/struct.Interval.html#method.tick) has [different semantics](https://docs.rs/tokio/latest/tokio/time/index.html#examples) than the previous, manual implementation using [`sleep`](https://docs.rs/tokio/latest/tokio/time/fn.sleep.html). I think the new semantics are the intended ones.

### DIFF
--- a/src/main_loop.rs
+++ b/src/main_loop.rs
@@ -19,6 +19,7 @@ use tokio::sync::broadcast;
 use tokio::sync::mpsc;
 use tokio::task::JoinHandle;
 use tokio::time;
+use tokio::time::MissedTickBehavior;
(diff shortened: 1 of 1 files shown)


Sword-Smith:
Also if, God forbid, any task takes longer than the time interval to complete then that task will, with this commit, choke the main loop. Whereas with the previous behavior, main loop would in that case still be available for other interactions.

Sword-Smith (approved):
In that case, LGTM.

PR #509, opened 2025-03-17
fix(peer_loop): fix race condition on new connection

Sword-Smith:
Two tokio tasks can run simultaneously that establish a connection incoming/outgoing to the same peer. If a write lock is not held over some final checks before a connection is established then two connections to the same peer might be made.

This reintroduces the holding of a lock while this final check is being made. The lock was removed in commit fa89d0414.

### DIFF
--- a/src/peer_loop.rs
+++ b/src/peer_loop.rs
@@ -1784,9 +1784,11 @@ impl PeerLoopHandler {
         const TIME_DIFFERENCE_WARN_THRESHOLD_IN_SECONDS: i128 = 120;
 
         let cli_args = self.global_state_lock.cli().clone();
-        let global_state = self.global_state_lock.lock_guard().await;
 
-        let standing = global_state
+        let standing = self
+            .global_state_lock
+            .lock_guard()
+            .await
             .net
             .peer_databases
             .peer_standings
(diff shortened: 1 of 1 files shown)


src/peer_loop.rs
```diff
@@ -1912,33 +1914,37 @@ impl PeerLoopHandler {
...
-        if global_state
-            .net
-            .peer_map
-            .values()
-            .any(|pi| pi.instance_id() == self.peer_handshake_data.instance_id)
+        // There is potential for a race-condition in the peer_map here, as
+        // we've previously counted the number of entries and checked if
+        // instance ID was already connected. But this check could have been
+        // invalidated by other tasks so we perform it again here under a write
+        // lock.
```
jan-ferdinand, line 1921: This comment doesn't clearly communicate why a write-lock is necessary for the check. How about:

“Up to this point, it is possible for multiple different tasks to perform connection initiation with the same node. To ensure that at most one such connection initiation attempt succeeds,
- acquire an exclusive (i.e. mutable) lock of the global state,
- assert that the node in question is not a peer, and only then
- add the node to the peer map.”

src/peer_loop.rs
```diff
@@ -1912,33 +1914,37 @@ impl PeerLoopHandler {
...
         {
-            bail!("Attempted to connect to already connected peer. Aborting connection.");
-        }
+            let mut global_state = self.global_state_lock.lock_guard_mut().await;
```
jan-ferdinand, line 1923: The global state is only needed for the peer map. I suggest to bind the peer map explicitly. This reduces repeated statements and communicates the fact that only the peer map is being accessed.

```rust
let peer_map = &mut global_state.net.peer_map;
```

src/peer_loop.rs
Sword-Smith: Right. The check needs to be done while holding any lock. But since the check ends with a write to `peer_map` and we cannot release the lock before writing, the lock needs to be a write lock.

src/peer_loop.rs
```diff
@@ -1821,33 +1823,35 @@ impl PeerLoopHandler {
...
-            .peer_map
-            .values()
-            .any(|pi| pi.instance_id() == self.peer_handshake_data.instance_id)
+        // There is potential for a race-condition in the peer_map here, as
+        // we've previously counted the number of entries and checked if
+        // instance ID was already connected. But other tasks could have made
+        // connections to a new node after the completion of that check, so we
+        // need to make the check while holding a lock to ensure that this
+        // check still holds. Since we're modifyin peer_map if this check is
+        // successful, we have to acquire a read-lock.
```
jan-ferdinand, line 1832: At minimum, this should be “write-lock”, not “read-lock”. However, I think the comment can be improved beyond that.

- There's no race condition “in” the `peer_map`, which is a fully sequential data structure.
- TOCTOU is a well-established class of bug and doesn't have to be explained in full detail; mentioning that the lock is maintained to prevent a TOCTOU bug is enough.

```suggestion
        // Multiple tasks might attempt to set up a connection concurrently. To
        // prevent time-of-check to time-of-use bugs, maintain a write-lock on
        // the global state between the checks and updating the peer map.
```

src/peer_loop.rs
jan-ferdinand: It seems that even my suggestion above missed the point. I've made an updated suggestion inline.

PR #511, opened 2025-03-18
Leave bootstrap nodes alone if well connected

jan-ferdinand:
In order to help newcomers to connect to peers on the network, certain nodes act as bootstrap nodes. Those nodes tend to be well known, and users tend to specify them as potential peers via command line arguments.

Previously, connections to potential peers specified as command line arguments were always re-initiated when dropped. This can cause an involuntary denial-of-service of those well-known bootstrap nodes.

With this PR, a well-connected node does not initiate connections with bootstrap nodes.

Second attempt of #488 after its reversal in 10e327b9 and ee379596.

### DIFF
--- a/src/config_models/cli_args.rs
+++ b/src/config_models/cli_args.rs
@@ -82,13 +82,21 @@ pub struct Args {
     #[clap(long)]
     pub(crate) max_connections_per_ip: Option<usize>,
 
-    /// Whether to act as bootstrapper node.
+    /// Whether to act as a bootstrap node.
     ///
-    /// Bootstrapper nodes ensure that the maximum number of peers is never
-    /// reached by disconnecting from existing peers when the maximum is about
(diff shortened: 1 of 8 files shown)


Sword-Smith:
You probably already found the problematic part of the code that causes the old node to hang up when receiving an unknown message type. It's here in `run` in `peer_loop.rs`:

```rust
                peer_message = peer.try_next() => {
                    let peer_address = self.peer_address;
                    let peer_message = match peer_message {
                        Ok(message) => message,
                        Err(err) => {
                            let msg = format!("Error when receiving from peer: {peer_address}");
                            error!("{msg}. Error: {err}");
                            bail!("{msg}. Closing connection.");
                        }
```

Currently this error causes the node to hang up. That behavior could be changed to e.g. log a warning.

PR #512, opened 2025-03-18
`proptest_state_machine` over `PeerMessage`

skaunov:
It's rough yet, and I'd appreciate review of the main approaches in it while I'm dealing with TODO and groom the whole contribution.

I understood @aszepieniec that the interactions are subject to change, so currently it's only outlined how an interaction outcome actually asserted (messages the peer-loop sends to the main, and checking `standing`). I can only repeat current behavior there, and if there's ideas to tune/amend that I guess it would be right place and time for that.

Another decision pending is weighting `PeerMessage` variants. I guess it's important to add random "bad" once a while, but uniform distribution washes away valid interactions to the margin. I can only roughly guess the distribution you might see fit, so it's the other open question.


src/models/blockchain/block/block_header.rs
```diff
@@ -281,6 +283,48 @@ pub(crate) mod block_header_tests {
...
+    proptest::prop_compose! {
+        pub fn block_header_strategy()(
+            version_raw in any::<u64>(),
+            height_raw in any::<u64>(),
+            prev_digest_raw in proptest::collection::vec(any::<u64>(), 5),
+            timestamp_raw in any::<u64>(),
+            nonce_raw in proptest::collection::vec(any::<u64>(), 5),
+            pow_raw in proptest::array::uniform6(any::<u32>()),
+            difficulty_raw in proptest::array::uniform5(any::<u32>()),
+            guesser_digest_raw in proptest::collection::vec(any::<u64>(), 5),
```
jan-ferdinand, line 295: Both `BFieldElement` and `Digest` implement `arbitrary::Arbitrary`. You can directly get a strategy for the correct type by using the `proptest_arbitrary_interop::arb` strategy.

skaunov:
@aszepieniec 
So, the state machine crash the node as was the aim of the task initially. On the other hand, due to the nature of `proptest` it does this in quite a trivial way just hitting some of handful `unwrap` at like <src/models/state/archival_state.rs>. Though `unwrap` in production code is still a bad pattern, and a right design choice for this is beyond my current knowledge of the system: `expect` alone won't save it obviously; should that return `Result` to be able to punish the peer; or should such thing be caught on deserialization and those `Option` should be processed earlier so `PeerMessage` would always have the complete info.

skaunov:
I need to think about "c)"; the first two seems quite extremes of the spectrum yet. :face_with_spiral_eyes:  And I really like the approach with examples.

For a quick peek I hit <https://github.com/Neptune-Crypto/neptune-core/blob/e78ccf5b55bac7948b6de8dae54845646352d4a4/src/models/state/archival_state.rs#L589>. Isn't it better to communicate with an `Err` that the record for `tip_digest` isn't there. Currently I can imagine how `match` will be replaced with `?` at some point, and then if `.unwrap()` will be replaced with another `?` the behavior will be changed to a wrong one.

skaunov:
@aszepieniec , I'd create a new issue from the previous comment since changing to `expect` is needed anyway. Tell me if that's a bad idea, pls.

Let's look at another example. I dug into, but not exhaustively. The test hits <https://github.com/skaunov/neptune-core/blob/8698d30980b1bafafc58622fba915df07e890fb5/src/models/state/archival_state.rs#L503>; it's called in the `pub` function <https://github.com/skaunov/neptune-core/blob/8698d30980b1bafafc58622fba915df07e890fb5/src/models/state/mod.rs#L254> (guarded by `...Lock` but it's already sounds like a borderline in this context) which doesn't check `new_block` height. My point is that it seems to possible to crash a node with this `expect`, and an `Error` for the wrong `height` would communicate to check this when calling the function.

Sword-Smith:
> @aszepieniec , I'd create a new issue from the previous comment since changing to `expect` is needed anyway. Tell me if that's a bad idea, pls.
> 
> Let's look at another example. I dug into, but not exhaustively. The test hits https://github.com/skaunov/neptune-core/blob/8698d30980b1bafafc58622fba915df07e890fb5/src/models/state/archival_state.rs#L503; it's called in the `pub` function https://github.com/skaunov/neptune-core/blob/8698d30980b1bafafc58622fba915df07e890fb5/src/models/state/mod.rs#L254 (guarded by `...Lock` but it's already sounds like a borderline in this context) which doesn't check `new_block` height. My point is that it seems to possible to crash a node with this `expect`, and an `Error` for the wrong `height` would communicate to check this when calling the function.

Wouldn't this `expect` only be hit if someone tried to set a block with height 0 as the new tip? Block height 0 is by definition the genesis block, and the genesis block is considered code, not data, so it can never be changed while the node is running.

Sword-Smith:
> @Sword-Smith , oh I misinterpreted pruning there - thanks! So is it possible to crash a node with the genesis for the tip? :thinking: 

A peer shouldn't be able to do that. If a peer disagrees on the Genesis block, that should trigger an automatic permaban of that peer.

Issue #510, opened 2025-03-18
Propose (and implement?) a refined Tx creation API.


dan-da:
This is a task that arose from #491 and #185.

The goal is to create a set of APIs that are simple to understand and use but flexible enough to enable creation of any Transaction allowed by the neptune consensus rules and to support a broad set of use cases.   And that should be true at both the rust/crate level and the RPC level.

edit: additional design goal:  the APIs should support creating/validating/sending Tx with arbitrary Coin, eg non-native tokens.

jan-ferdinand:
Thanks for the write-up. The interface you sketch seems comfortabe to use. Thanks also for writing up bitcoin's interface, that's a good comparison point.

1. On the RPC `send_native_coin_payment`, you write: “The transaction's proof will need to be upgraded from Witness before it can be included in a block.” Does that imply that `send_native_coin_payment` only ever produces transactions containing `TransactionProof::Witness`?

1. Does it make sense to include RPC endpoints akin to `upgrade_upgradeable_transaction` and a corresponding `poll_for`? (Maybe those endpoints already exist?) This would be especially useful if the answer to the previous question is “yes.”

1. Looking at some [other RPC documentation](https://cloud.google.com/spanner/docs/reference/rpc/google.longrunning), it seems we might want to expand the interface for long-running operations. For example “cancel” seems immediately useful, and “wait” could have applications. We might want to design a unified interface (or steal an existing design) for long-running RPCs.

1. The `generate_tx_details` RPC does not allow to set the mutator set accumulator or the change policy.
    1. Does that mean that the inputs and outputs must be balanced?
    1. I assume the mutator set accumulator can be inferred from the current wallet state. Is that correct?

1. The `poll_for_transaction_proof` RPC returns a `Result<TransactionProof, String>`. Does the `Err` case communicate an RPC failure, or a proving failure, or both? If both, would it make sense to distinguish between the two failure possibilities (other than through the string's content)? For example, would a `Result<Result<TransactionProof, ProvingError>, String>` make sense? (That type feels like a bit of an abomination, to be honest.)

1. Would the (todo) RPC endpoint(s) for selecting transaction inputs return those very inputs, or would they change some internal state in some way? Intuitively, I'd prefer the former.

dan-da:
> Thanks for the write-up. The interface you sketch seems comfortabe to use. Thanks also for writing up bitcoin's interface, that's a good comparison point.

Hey, thanks for reviewing and the feedback!

[...]

PR #513, opened 2025-03-19
tx api refining

dan-da:
addresses #510

This is a (very) draft PR intended to:

1. inform that I'm working on this.
2. share direction/progress and diffs

note that this is building on asz/test-larger-blocks, not master.

so far:

1. implemented builders for TransactionDetails, TransactionProof, Transaction.
2. reimplemented GlobalState::create_transaction() and create_raw_transaction() to use these builders.
3. implemented a method for retrieving all spendable inputs, and an RPC endpoint.
4. made TritonVmJobQueue accessible via a global oncelock.   it's now removed from GlobalStateLock.
5. started a new layer in src/tx_initiation/ that is intended to be callable by rust users of the crate as well as RPC server.
6. separated out recording tx (modifying local state) from broadcasting to peers.
7. added TxInput and TxInputList as newtypes for UnlockedUtxo and Vec<UnlockedUtxo>. 
8. other smaller refactors.

Going forward options would seem to be:

1. we merge this into asz/test-larger-blocks and continue dev there.
2. I continue on my own branch, possibly the branches diverge.
3. these changes (and direction) are not accepted, and we abandon them.


src/models/state/wallet/transaction_builder/transaction_builder.rs
```diff
@@ -0,0 +1,39 @@
+use std::sync::Arc;
+
+use crate::models::state::PrimitiveWitness;
+use crate::models::state::Transaction;
+use crate::models::state::TransactionDetails;
+use crate::models::state::TransactionProof;
+
+#[derive(Debug, Default)]
+pub struct TransactionBuilder {
```
jan-ferdinand, line 9: I know that one of your quests is to reduce the number of dependencies. However, you might want to consider adding a dependency like [`bon`](https://crates.io/crates/bon) for defining the various builders.

Disadvantages:
- one more dependency
- have to learn the API

Advantages:
- less code to write, read, and maintain
- uses the typestate pattern, trading runtime panics for compile errors

For example, this entire file can be replaced by a few additional lines on `struct Transaction`:

```diff
- #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, GetSize)]
+ #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, GetSize, Builder)]
pub struct Transaction {
+   #[builder(with = |x: TransactionDetails| PrimitiveWitness::from_transaction_details(&x).kernel)]
+   #[builder(name = transaction_details)]
    pub kernel: TransactionKernel,

+   #[builder(name = transaction_proof)] // maybe no improvement, but consistent with status quo
    pub proof: TransactionProof,
}
```

I haven't yet figured out how to replace the default `build` method in case the actual build step is doing some heavy lifting.

dan-da: thx for the suggestion.  yeah I played with bon a bit once before.    I think for now I like having total control over the builder, and also I'm trying to consolidate the tx initiation stuff in a central place.  (not quite there yet).    Might be worth revisiting later, esp if you or someone feels more motivated about it than I.

dan-da:
milestone:  the proposed API is basically built:  all builders, middle layer, and rpc endpoints now exist.

there is some cleanup and minor-ish refactoring to do, eg:

- [ ] impl and use thiserror Error types for all APIs.
- [ ] use TransactionProofType instead of ProvingCapability.  (maybe merge them). 
- [ ] split and rename TransactionSender to make layers and public API clearer.

at that point, the original goals will be accomplished.

However, there remains the new goal to make the APIs work with non-native Coin utxos.   There are design challenges around that, eg Utxo presently has no method to return an amount for a non-native Coin, and there is also no type to represent such amounts.  Or utxo without any amount.  So I will welcome any help with working through such issues, so we don't needlessly limit the API, or worse behave incorrectly with non-native Coin.

jan-ferdinand:
Given the PR's size, this is the first of potentially several comments I'll be making. I have a few thousand lines of code left to read. 🙂

I like the APIs suggested in this PR. Allowing downstream crates to use a simple interface if they want to while also providing a way to control all details is powerful.

I like that the low-level builder API can be used seamlessly with the mid-level initiator API. If the downstream crate is only interested in fine-grained control over one aspect of transaction creation, it can use the low-level builder API for that aspect, and use the mid-level initiator API for the rest. It also seems that the mid-level initiator API provides entrypoints to discover the low-level builder API. That is, if a downstream crate wants more control than is provided by the high-level `.send()` API, they find the mid-level initiator API. Each of the relevant methods on the central struct of the mid-level initiator API informs about the relevant builder of the low-level builder API.

Here are some points that I think could be improved:
- Currently, it feels like the documentation's intended audience is PR reviewers, at least some of the time. This is nice while reviewing the PR. Before merging the PR, those parts should probably be reworked.
- I think some of the documentation comments benefit from a slightly more neutral delivery. I understand that the old (current) API might be a point of frustration at times. I caution against letting such sentiment bleed through in the documentation.[^frust]
- I appreciate the desire to create high-quality code, and any attempts to keep code quality high. I am strongly of the belief that this benefits us and the project greatly in both the short and the long run, for example by reducing the number of bugs, increasing developer satisfaction, attracting collaborators, to name a few points. However, the suggested dedicated code owner and the rules associated with this suggestion are new policy for the codebase. In my opinion, this should be discussed with the core team prior to introduction, as there are various pros and cons to consider.
- Personally, I strongly prefer compilation errors over runtime errors. I'd love if the various builders used the typestate pattern and only exposed those methods that make sense in their current state. I'm happy to put in work on that front.

[^frust]: It's both subtle and a real possibility that I'm receiving signals that are not being sent. An example would be “[t]his module attempts to rectify [the] situation [of an unclean API].”

src/job_queue/queue.rs
```diff
@@ -79,8 +80,18 @@ struct AddJobMsg<P> {
...
     tx: mpsc::UnboundedSender<JobQueueMsg<P>>,
-    tracker: TaskTracker,
-    refs: Arc<()>,
+
+    process_jobs_task_handle: JoinHandle<()>, // Store the job processing task handle
+    add_job_task_handle: JoinHandle<()>,      // store the job addition task handle.
+}
+
+// useful for detecting when a receiver gets dropped.
+struct LoggedReceiver<T>(UnboundedReceiver<T>);
```
jan-ferdinand, line 89: `impl Deref`?

```rust
impl<T> core::ops::Deref for LoggedReceiver<T> {
    type Target = UnboundedReceiver<T>;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}

impl<T> core::ops::DerefMut for LoggedReceiver<T> {
    fn deref_mut(&mut self) -> &mut Self::Target {
        &mut self.0
    }
}
```

jan-ferdinand (commented):
~~I notice you sometimes put 2 spaces in between a sentence-ending period and the start of the next sentence. Is that an editing artifact or do they signify something?~~

src/api/tx_initiation/mod.rs
```diff
@@ -0,0 +1,78 @@
+//! provides builders and APIs for initiating transactions.
+//!
+//! The module is designed to provide flexible, low-level manipulation
+//! of transactions as well as an easy to use high-level [send()](send::TransactionSender::send()) method.
+//!
+//! It is composed of three layers:
+//!
+//! 1. [builder] provides a set of transaction builders.
+//! 2. [initiator] provides [TransactionInitiator](initiator) that wraps the builders and broadcasts the tx.
+//! 3. [send] provides [TransactionSender](initiator) with a single, simple `send()` method.
```
jan-ferdinand, line 10:
```suggestion
//! 3. [send] provides [TransactionSender](send::TransactionSender) with a
//!    single, simple `send()` method.
```

dan-da:
closing this.  obsoleted by #544 which was already merged.

PR #516, opened 2025-03-20
Regulate prover fee notifications

### LINKED ISSUE #441: Imported Wallets Do Not Recognize Own Compositions
Wallets come into ownership of the composition rewards of blocks through the `ExpectedUtxo` mechanism. However, when the same wallet secret is imported onto a different machine, it does not recognize the composer UTXOs. It would be nice if it did.

Unfortunately, there is a bit of a complication. The wallet has no way of know how many coins to expect. The best it can do is make an educated guess based on the block subsidy and the transaction fee. If this guess is correct, then it can derive all information necessary to reproduce the addition record and scan incoming blocks for a match. If this guess is incorrect (for instance, because the block included a transaction that bumped the fee) then I don't see how the wallet can recover these funds.

Proposed path forwards:
 - [ ] Add to `neptune-cli` a command `scan-composer-rewards` which scans blocks in an archival node for composer UTXOs that the wallet can unlock. This command should be able to take one or even several guesser fee fractions.
 - [ ] #443
 - [ ] Add an optional CLI argument `--off-chain-composer-rewards` to `neptune-core` that makes announcements for composition rewards off-chain.
 - [ ] Add functionality for exporting and importing `ExpectedUtxo`s.

Unfortunately, it seems that funds are lost in the following scenario:
 - User successfully composed block
 - which included a fee-paying transaction
 - before the default notification was set to on-chain (second bullet point above)
 - on a machine that is no longer accessible
 - but whose secret key was backed up.

### LINKED ISSUE #443: Make composer reward notifications on-chain by default.

aszepieniec:
This PR:
 - Introduces a CLI flag that regulates the notification policy for prover fees (meaning upgrader fees and composition fees, guesser fees are explicitly excluded). The notification policy can be
   - `OnChainSymmetric` (default)
   - `OffChain` (for users who know what they are doing)
   - `OnChainGeneration` (anticipates cold composing/upgrading, where the client mines to an address but does not know the spending key).
 - Streamlines the CLI arguments, and how it is passed down to global state and wallet state.
 - Implements and tests the fee notification policy.
 - Automatically derives and tracks the symmetric (in addition to generation, which was already being done) key with derivation index 0. The default fee notification policy is to use this symmetric key.
 - Adds a step in scan mode when scan mode is active, unrelated to fee notification policy, which tries to reproduce the composer UTXOs and claims them if successful.
 - Modifies the insertion policy into the mempool: new but conflicting transactions with lower fee density are *not* rejected if the inputs and outputs are identical, as this likely indicates that the transaction was updated to a new mutator set.

Fixes #441.
Closes #443.

### DIFF
--- a/src/config_models/cli_args.rs
+++ b/src/config_models/cli_args.rs
@@ -161,19 +161,21 @@ pub struct Args {
     /// transactions they work on, by directing it to an output only they can
     /// spend. By default, a public announcement is added to the transaction to
     /// enable the composer or proof-upgrader to recover such UTXOs after
-    /// importing the key onto a new machine. This public announcement is
+    /// restoring the wallet from seed phrase. This public announcement is
     /// encryped by default under a symmetric key.
     ///
     /// Valid options:
-    ///  - `--fee-notification on-chain-symmetric` (default) On-chain backups
-    ///    using symmetric key ciphertexts.
-    ///  - `--fee-notification on-chain-generation` On-chain backups using
-    ///    generation addresses, which means that public key encryption is used
-    ///    instead. Note that public key ciphertexts are significantly larger
-    ///    (and thus take up more blockchain space) than symmetric ciphertexts.
-    ///  - `--fee-notification off-chain` Avoid on-chain backups. Saves
-    ///    blockchain space, but risks loss of funds. Enable only if you know
-    ///    what you are doing.
+    ///
+    ///  - `on-chain-symmetric` (default) On-chain backups using symmetric key
+    ///    ciphertexts.
+    ///
+    ///  - `on-chain-generation` On-chain backups using generation addresses,
+    ///    which means that public key encryption is used instead. Note that
+    ///    public key ciphertexts are significantly larger (and thus take up
+    ///    more blockchain space) than symmetric ciphertexts.
+    ///
+    ///  - `off-chain` Avoid on-chain backups. Saves blockchain space, but risks
+    ///    loss of funds. Enable only if you know what you are doing.
     ///
     /// This flag does not apply to guesser fees because those UTXOs are
     /// generated automatically.
--- a/src/config_models/cli_args.rs
+++ b/src/config_models/cli_args.rs
@@ -153,6 +153,22 @@ pub struct Args {
     #[clap(long, default_value = "0.01", value_parser = NativeCurrencyAmount::coins_from_str)]
     pub(crate) min_gobbling_fee: NativeCurrencyAmount,
 
+    /// Whether to keep the UTXO notifications for composer fees and
+    /// proof-upgrader fees off chain.
+    ///
+    /// Composers and proof-upgraders can gobble a portion of the fee of
+    /// transactions they work on, by directing it to an output only they can
+    /// spend. By default, a public announcement is added to the transaction
+    /// which enables the composer or proof-upgrader to use the blockchain to
+    /// back up UTXO data and recover such UTXOs after importing the key onto a
+    /// new machine. Enable this flag to avoid wasting blockchain space, if you
+    /// know what you're doing and a robust alternative for backups is in place.
+    ///
+    /// This flag does not apply to guesser fees because those UTXOs are
+    /// generated automatically.
+    #[clap(long)]
+    pub(crate) offchain_fee_notifications: bool,
+
     /// Prune the mempool when it exceeds this size in RAM.
     ///
     /// Units: B (bytes), K (kilobytes), M (megabytes), G (gigabytes)
--- a/src/config_models/cli_args.rs
+++ b/src/config_models/cli_args.rs
@@ -13,6 +13,7 @@ use itertools::Itertools;
 use num_traits::Zero;
 use sysinfo::System;
 
+use super::fee_notification_policy::FeeNotificationPolicy;
 use super::network::Network;
 use crate::job_queue::triton_vm::TritonVmJobPriority;
 use crate::models::blockchain::type_scripts::native_currency_amount::NativeCurrencyAmount;
@@ -158,16 +159,26 @@ pub struct Args {
     ///
     /// Composers and proof-upgraders can gobble a portion of the fee of
     /// transactions they work on, by directing it to an output only they can
-    /// spend. By default, a public announcement is added to the transaction
-    /// which enables the composer or proof-upgrader to use the blockchain to
-    /// back up UTXO data and recover such UTXOs after importing the key onto a
-    /// new machine. Enable this flag to avoid wasting blockchain space, if you
-    /// know what you're doing and a robust alternative for backups is in place.
+    /// spend. By default, a public announcement is added to the transaction to
+    /// enable the composer or proof-upgrader to recover such UTXOs after
+    /// importing the key onto a new machine. This public announcement is
+    /// encryped by default under a symmetric key.
+    ///
+    /// Valid options:
+    ///  - `--fee-notification on-chain-symmetric` (default) On-chain backups
+    ///    using symmetric key ciphertexts.
+    ///  - `--fee-notification on-chain-generation` On-chain backups using
+    ///    generation addresses, which means that public key encryption is used
+    ///    instead. Note that public key ciphertexts are significantly larger
+    ///    (and thus take up more blockchain space) than symmetric ciphertexts.
+    ///  - `--fee-notification off-chain` Avoid on-chain backups. Saves
+    ///    blockchain space, but risks loss of funds. Enable only if you know
+    ///    what you are doing.
     ///
     /// This flag does not apply to guesser fees because those UTXOs are
     /// generated automatically.
-    #[clap(long)]
-    pub(crate) offchain_fee_notifications: bool,
+    #[clap(long, default_value = "on-chain-symmetric", value_parser = FeeNotificationPolicy::parse)]
+    pub(crate) fee_notification: FeeNotificationPolicy,
 
     /// Prune the mempool when it exceeds this size in RAM.
     ///
--- a/src/main_loop/proof_upgrader.rs
+++ b/src/main_loop/proof_upgrader.rs
@@ -158,7 +158,7 @@ impl UpgradeJob {
     /// The gobbling fee charged for an upgrade job
     ///
     /// Gobbling fees are charged when a transaction is upgraded from
-    /// proof-collectino to single-proof, or when two single proofs are merged.
+    /// proof-collection to single-proof, or when two single proofs are merged.
     /// The other cases are either not worth it, as you need to create a single-
     /// proof to gobble, or the proof upgrade relates to own funds.
     ///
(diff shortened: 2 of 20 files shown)


aszepieniec:
image

If CI fails (I think it will) it will be because some proofs are missing from the proof server.

src/main_loop/proof_upgrader.rs
```diff
@@ -422,6 +426,39 @@ impl UpgradeJob {
         }
     }
 
+    fn gobbler_notification_method_with_receiver_preimage(
+        own_wallet_entropy: &WalletEntropy,
+        notification_policy: FeeNotificationPolicy,
+    ) -> (UtxoNotifyMethod, Digest) {
+        let gobble_receiver_key = match notification_policy {
```
Sword-Smith, line 433: The type here is `SpendingKey`. But variable name suggests an address "receiver_key". How about `gobble_beneficiary_key`?

src/main_loop/proof_upgrader.rs
```diff
@@ -422,6 +426,39 @@ impl UpgradeJob {
...
+                SpendingKey::from(own_wallet_entropy.nth_symmetric_key(0))
+            }
+            FeeNotificationPolicy::OnChainSymmetric => {
+                SpendingKey::from(own_wallet_entropy.nth_symmetric_key(0))
+            }
+            FeeNotificationPolicy::OnChainGeneration => {
+                SpendingKey::from(own_wallet_entropy.nth_generation_spending_key(0))
+            }
+        };
+        let receiver_preimage = gobble_receiver_key.privacy_preimage().unwrap();
```
Sword-Smith, line 444: Maybe also use "expect" here?

src/main_loop/proof_upgrader.rs
```diff
@@ -422,6 +426,39 @@ impl UpgradeJob {
         }
     }
 
+    fn gobbler_notification_method_with_receiver_preimage(
+        own_wallet_entropy: &WalletEntropy,
+        notification_policy: FeeNotificationPolicy,
+    ) -> (UtxoNotifyMethod, Digest) {
+        let gobble_receiver_key = match notification_policy {
+            FeeNotificationPolicy::OffChain => {
+                SpendingKey::from(own_wallet_entropy.nth_symmetric_key(0))
```
jan-ferdinand, line 435: Can symmetric or generation keys ever be considered “burned”? Phrased differently, are there scenarios where a user wants to stop using a particular key? For example, can key re-usage lead to transaction linkability?

src/main_loop/proof_upgrader.rs
```diff
@@ -434,53 +471,60 @@ impl UpgradeJob {
...
-                    .await?;
+
+            let expected_utxos =
+                if matches!(fee_notification_policy, FeeNotificationPolicy::OffChain) {
+                    gobbler_transaction_details
+                        .tx_outputs
+                        .iter()
+                        .map(|x| {
+                            ExpectedUtxo::new(
+                                x.utxo(),
```
jan-ferdinand, line 504: That's some serious indentation. Can be tamed by, for example by:

```rust
                let output_to_expected_utxo = |output| {
                    let utxo = output.utxo();
                    let sender_randomness = output.sender_randomness();
                    let notifier = UtxoNotifier::FeeGobbler;
                    ExpectedUtxo::new(utxo, sender_randomness, receiver_preimage, notifier)
                };
                gobbler_transaction_details
                    .tx_outputs
                    .iter()
                    .map(output_to_expected_utxo)
                    .collect_vec()
```

src/main_loop/proof_upgrader.rs
```diff
@@ -434,53 +471,60 @@ impl UpgradeJob {
...
-                    )
-                })
-                .collect_vec();
-            let gobbler = PrimitiveWitness::from_transaction_details(&gobbler);
-            let gobbler_proof =
-                SingleProof::produce(&gobbler, triton_vm_job_queue, proof_job_options.clone())
-                    .await?;
+
+            let expected_utxos =
+                if matches!(fee_notification_policy, FeeNotificationPolicy::OffChain) {
```
jan-ferdinand, line 498: No need for the `matches!` macro.

```suggestion
                if fee_notification_policy == FeeNotificationPolicy::OffChain {
```

src/mine_loop.rs
```diff
@@ -396,23 +398,20 @@ pub(crate) async fn make_coinbase_transaction_stateless(
...
 
-/// Compute `TransactionDetails` and a list of `TxOutput`s for a coinbase
-/// transaction.
-pub(super) fn prepare_coinbase_transaction_stateless(
-    latest_block: &Block,
+/// Produce two outputs spending a given portion of the coinbase.
+///
+/// The coinbase is the block subsidy plus the anticipated transaction fee.
+/// There are two outputs because one is liquid immediately, and the other is
+/// locked for 3 years. The portion is determined by the guesser fee fraction
```
jan-ferdinand, line 405: “The portion […].” The portion that is liquid immediately? Seems unlikely, but I'm not sure what other portion this might refer to.

src/main_loop/proof_upgrader.rs
```diff
@@ -434,53 +471,60 @@ impl UpgradeJob {
...
+                            ExpectedUtxo::new(
+                                x.utxo(),
+                                x.sender_randomness(),
+                                receiver_preimage,
+                                UtxoNotifier::FeeGobbler,
+                            )
+                        })
+                        .collect_vec()
+                } else {
+                    vec![]
```
Sword-Smith, line 512: Please keep the style in the original code. The original code uses shadowing to reduce the number of variables in scope, such that what you now call `gobbler_transaction_details` was previously shadowed once (what you now call) `gobbler_primitive_witness` was declared.

src/main_loop/proof_upgrader.rs
```diff
@@ -422,6 +426,39 @@ impl UpgradeJob {
         }
     }
 
+    fn gobbler_notification_method_with_receiver_preimage(
```
Sword-Smith, line 429: Non-public functions don't belong in the middle of a list of public function declarations. They belong either at the bottom or at the top of the module. Or, if it's not tested, it can also be declared as a helper function inside the function where it's used.

src/mine_loop.rs
```diff
@@ -433,23 +428,53 @@ pub(super) fn prepare_coinbase_transaction_stateless(
...
+    info!("Creating coinbase for block of height {next_block_height}.");
+
+    let coinbase_amount = Block::block_subsidy(next_block_height);
+    let [liquid_coinbase_output, timelocked_coinbase_output] =
+        coinbase_outputs(coinbase_amount, composer_parameters, timestamp)?;
+    let total_composer_fee = liquid_coinbase_output.utxo().get_native_currency_amount()
+        + timelocked_coinbase_output
+            .utxo()
+            .get_native_currency_amount();
+    let total_guesser_fee = coinbase_amount.checked_sub(&total_composer_fee).unwrap();
```
jan-ferdinand, line 471: Is it part of the `coinbase_output()` contract that the total composer fee is less than or equal to the coinbase amount?

src/models/state/wallet/wallet_configuration.rs
```diff
@@ -74,6 +74,12 @@ impl WalletConfiguration {
             }
         };
 
+        if cli_args.compose {
+            if let Some(scan_mode_configuration) = self.scan_mode.as_mut() {
+                scan_mode_configuration.set_guesser_fraction(cli_args.guesser_fraction);
+            };
+        }
+
```
Sword-Smith, line 82: Composers are changing their guesser fraction often. I don't know how meaningful or helpful this parameter definition is here.

src/mine_loop.rs
```diff
@@ -596,17 +615,22 @@ pub(crate) async fn create_block_transaction_from(
...
-        .map(|txo| {
-            ExpectedUtxo::new(
-                txo.utxo(),
-                txo.sender_randomness(),
-                coinbase_recipient_spending_key.privacy_preimage(),
-                UtxoNotifier::OwnMinerComposeBlock,
-            )
-        })
-        .collect();
+    let own_expected_utxos =
```
jan-ferdinand, line 618: 1. A duplicate of this exists at line 1035 in this file. Does it make sense to factor this out into a stand-alone function?
2. There is some serious indentation going on. You can combat this, for example, by:

```rust
impl ComposerParameters {
    // todo: this function's name is probably not on point
    fn expect_utxo(&self, composer_outputs: TxOutputList) -> Vec<ExpectedUtxo> {
        let Some(receiver_preimage) = self.maybe_receiver_preimage() else {
            return Vec::new();
        };

        let output_to_expected_utxo = |output| {
            let utxo = output.utxo();
            let sender_randomness = output.sender_randomness();
            let notifier = UtxoNotifier::OwnMinerComposeBlock;
            ExpectedUtxo::new(utxo, sender_randomness, receiver_preimage, notifier)
        };
        composer_outputs
            .iter()
            .map(output_to_expected_utxo)
            .collect()
    }
}
```

Sword-Smith (approved):
Review done in person. Excellent PR. Most importantly for me, this fixes a bug in the mempool where mutator set updated transactions were (with a 50 % chance) rejected from the mempool.

aszepieniec:
@jan-ferdinand: Thorkil finished his review and the resulting changes have been integrated. You can continue your review now without flux.

src/models/state/wallet/wallet_state.rs
```diff
@@ -165,6 +169,32 @@ impl Debug for WalletState {
 }
 
 impl WalletState {
+    /// Generate [`ComposerParameters`] for composing the next block.
+    ///
+    ///  # Panics
+    ///
+    ///  - If the `guesser_fraction` is not a fraction contained in [0;1].
```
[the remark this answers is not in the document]
jan-ferdinand: Eh, you've resolved that already. :relaxed:

src/config_models/cli_args.rs
```diff
@@ -153,6 +154,32 @@ pub struct Args {
...
+    /// proof-upgrader fees off chain.
+    ///
+    /// Composers and proof-upgraders can gobble a portion of the fee of
+    /// transactions they work on, by directing it to an output only they can
+    /// spend. By default, a public announcement is added to the transaction to
+    /// enable the composer or proof-upgrader to recover such UTXOs after
+    /// importing the key onto a new machine. This public announcement is
+    /// encryped by default under a symmetric key.
+    ///
+    /// Valid options:
```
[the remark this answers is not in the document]
Sword-Smith: Enabling "wrap_help" does not seem to change anything. Nor do ChatGPT's suggestions. Let's stick to newlines between each list item for now; a later fixer is welcome to submit a PR.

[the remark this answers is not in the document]
jan-ferdinand: `wrap_help` doesn't help with this particular issue, but it increases general readibility of the help text in other areas.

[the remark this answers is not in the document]
jan-ferdinand: I agree that that's not the point or focus of this PR. :+1:

src/main_loop/proof_upgrader.rs
Sword-Smith: This PR trades off privacy for proof-upgraders and composers in exchange for the ability to recover all their earnings from the seed phrase. The privacy lost is recipient linkability: observers will be able to tell that the same proof-upgrader earned two different fees, for instance.

Proof-upgraders and composers that do not want to sacrifice their privacy in this way are welcome to use `off-chain`, at the expense of more hassle managing UTXOs across wallet migrations.

About future modifications: we could consider adding variants for
 - cold composing / proof-upgrading
 - same on chain options as currently but with bumped derivation fees
 - new address formats that require the recipient (and in this case, the composer or proof-upgrader) to perform more expensive scanning when recovering UTXOs.

jan-ferdinand: Alright, that makes sense. Thanks for clarifying. I hadn't fully understood the implications with respect to linkability before.

I advocate to document this trade-off in the help text for the relevant command line argument. Furthermore, we might want to consider introducing a section on privacy in the readme, or in other parts of the documentation. Such a section could contain a suggested configuration for maximum privacy, highlighting drawbacks.

PR #519, opened 2025-03-21
refactor(`TritonVmJobQueue`): Use singleton pattern

aszepieniec:
The purpose of the job queue is to prevent different jobs from competing for the same compute resources at the same time. As such, having multiple job queues defeats the purpose of having one. In other words, there should only ever be one job queue.

This commit refactors the interface of `JobQueue` and `TritonVmJobQueue` to enforce that only one instance of the latter type is ever constructed. Specifically, the constructor `start` is downgraded in visibility from public to private. To get the Triton VM proof queue object, the caller calls `global_triton_vm_job_queue()`, which is a public function that internally uses a static `OnceLock` to ensure that no more than one instance is ever created.

Also, this commit removes the reference to the triton vm job queue from all function interfaces.

### DIFF
--- a/src/job_queue/queue.rs
+++ b/src/job_queue/queue.rs
@@ -1,6 +1,7 @@
 use std::collections::VecDeque;
 use std::sync::Arc;
 use std::sync::Mutex;
+use std::sync::OnceLock;
 
 use tokio::sync::mpsc;
 use tokio::sync::oneshot;
@@ -16,6 +17,7 @@ use super::traits::JobCompletion;
 use super::traits::JobResult;
 use super::traits::JobResultReceiver;
 use super::traits::JobResultSender;
+use super::triton_vm::TritonVmJobPriority;
 
 /// A job-handle enables cancelling a job and awaiting results
 #[derive(Debug)]
@@ -77,6 +79,10 @@ struct AddJobMsg<P> {
 }
 
 /// implements a job queue that sends result of each job to a listener.
+///
+/// The generic argument provides type safety and clarity in case we implement
+/// job queues for other job types. For each job type, there should only ever be
+/// one instance.
 pub struct JobQueue<P> {
     tx: mpsc::UnboundedSender<JobQueueMsg<P>>,
     tracker: TaskTracker,
@@ -122,7 +128,7 @@ impl<P> Drop for JobQueue<P> {
 
 impl<P: Ord + Send + Sync + 'static> JobQueue<P> {
     /// creates job queue and starts it processing.  returns immediately.
-    pub fn start() -> Self {
+    fn start() -> Self {
         struct CurrentJob {
             job_num: usize,
             cancel_tx: JobCancelSender,
@@ -272,6 +278,18 @@ impl<P: Ord + Send + Sync + 'static> JobQueue<P> {
             cancel_tx,
         })
     }
+}
+
+/// A job queue for Triton VM Jobs.
+pub type TritonVmJobQueue = JobQueue<TritonVmJobPriority>;
+
+/// Global singleton accessor for the Triton VM Job Queue
+//
+// Ideally we implement a generic function `instance` on JobQueue but it seems
+// as though generic type arguments do not play ball with static pointers.
+pub fn global_triton_vm_job_queue() -> &'static TritonVmJobQueue {
+    static REGISTRY: OnceLock<TritonVmJobQueue> = OnceLock::new();
+    REGISTRY.get_or_init(TritonVmJobQueue::start)
 }
 
 #[cfg(test)]
--- a/src/job_queue/triton_vm/mod.rs
+++ b/src/job_queue/triton_vm/mod.rs
@@ -1,4 +1,5 @@
-mod triton_vm_job_queue;
+mod triton_vm_job_priority;
 
-pub use triton_vm_job_queue::TritonVmJobPriority;
-pub use triton_vm_job_queue::TritonVmJobQueue;
+pub use super::queue::global_triton_vm_job_queue;
+pub use super::queue::TritonVmJobQueue;
+pub use triton_vm_job_priority::TritonVmJobPriority;
--- a/src/job_queue/triton_vm/triton_vm_job_priority.rs
+++ b/src/job_queue/triton_vm/triton_vm_job_priority.rs
@@ -1,5 +1,3 @@
-use super::super::JobQueue;
-
 // todo: maybe we want to have more levels or just make it an integer eg u8.
 // or maybe name the levels by type/usage of job/proof.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Default)]
@@ -11,6 +9,3 @@ pub enum TritonVmJobPriority {
     High = 4,
     Highest = 5,
 }
-
-/// provides type safety and clarity in case we implement multiple job queues.
-pub type TritonVmJobQueue = JobQueue<TritonVmJobPriority>;
--- a/src/main_loop.rs
+++ b/src/main_loop.rs
@@ -30,7 +30,6 @@ use tracing::warn;
 use crate::connect_to_peers::answer_peer;
 use crate::connect_to_peers::call_peer;
 use crate::job_queue::triton_vm::TritonVmJobPriority;
-use crate::job_queue::triton_vm::TritonVmJobQueue;
 use crate::macros::fn_name;
 use crate::macros::log_slow_scope;
 use crate::models::blockchain::block::block_header::BlockHeader;
@@ -432,7 +431,6 @@ impl MainLoopHandler {
     /// Sends the result back through the provided channel.
     async fn update_mempool_jobs(
         update_jobs: Vec<UpdateMutatorSetDataJob>,
-        job_queue: &TritonVmJobQueue,
         transaction_update_sender: mpsc::Sender<Vec<Transaction>>,
         proof_job_options: TritonVmProofJobOptions,
     ) {
@@ -445,10 +443,7 @@ impl MainLoopHandler {
             // Jobs for updating txs in the mempool have highest priority since
             // they block the composer from continuing.
             // TODO: Handle errors better here.
-            let job_result = job
-                .upgrade(job_queue, proof_job_options.clone())
-                .await
-                .unwrap();
+            let job_result = job.upgrade(proof_job_options.clone()).await.unwrap();
             result.push(job_result);
         }
 
@@ -1317,7 +1312,6 @@ impl MainLoopHandler {
         // like mining, or proving our own transaction. Running the prover takes
         // a long time (minutes), so we spawn a task for this such that we do
         // not block the main loop.
-        let vm_job_queue = self.global_state_lock.vm_job_queue().clone();
         let perform_ms_update_if_needed =
             self.global_state_lock.cli().proving_capability() == TxProvingCapability::SingleProof;
 
@@ -1329,7 +1323,6 @@ impl MainLoopHandler {
                 .spawn(async move {
                     upgrade_candidate
                         .handle_upgrade(
-                            &vm_job_queue,
                             tx_origin,
                             perform_ms_update_if_needed,
                             global_state_lock_clone,
@@ -1353,7 +1346,6 @@ impl MainLoopHandler {
     ) {
         // job completion of the spawned task is communicated through the
         // `update_mempool_txs_handle` channel.
-        let vm_job_queue = self.global_state_lock.vm_job_queue().clone();
         if let Some(handle) = main_loop_state.update_mempool_txs_handle.as_ref() {
             handle.abort();
         }
@@ -1371,13 +1363,7 @@ impl MainLoopHandler {
             tokio::task::Builder::new()
                 .name("mempool tx ms-updater")
                 .spawn(async move {
-                    Self::update_mempool_jobs(
-                        update_jobs,
-                        &vm_job_queue,
-                        update_sender,
-                        job_options,
-                    )
-                    .await
+                    Self::update_mempool_jobs(update_jobs, update_sender, job_options).await
                 })
                 .unwrap(),
         );
@@ -1670,8 +1656,6 @@ impl MainLoopHandler {
                         panic!("Expected Primitive witness. Got: {:?}", transaction.proof);
                     };
 
-                    let vm_job_queue = self.global_state_lock.vm_job_queue().clone();
-
                     let proving_capability = self.global_state_lock.cli().proving_capability();
                     let upgrade_job =
                         UpgradeJob::from_primitive_witness(proving_capability, primitive_witness);
@@ -1687,7 +1671,6 @@ impl MainLoopHandler {
                         .spawn(async move {
                         upgrade_job
                             .handle_upgrade(
-                                &vm_job_queue,
                                 TransactionOrigin::Own,
                                 true,
                                 global_state_lock_clone,
@@ -2063,7 +2046,6 @@ mod test {
 
     mod proof_upgrader {
         use super::*;
-        use crate::job_queue::triton_vm::TritonVmJobQueue;
         use crate::models::blockchain::transaction::Transaction;
         use crate::models::blockchain::transaction::TransactionProof;
         use crate::models::blockchain::type_scripts::native_currency_amount::NativeCurrencyAmount;
@@ -2100,7 +2082,6 @@ mod test {
                     fee,
                     in_seven_months,
                     tx_proof_type,
-                    &TritonVmJobQueue::dummy(),
                 )
                 .await
                 .unwrap()
--- a/src/main_loop/proof_upgrader.rs
+++ b/src/main_loop/proof_upgrader.rs
@@ -11,7 +11,6 @@ use tracing::info;
 
 use super::TransactionOrigin;
 use crate::job_queue::triton_vm::TritonVmJobPriority;
-use crate::job_queue::triton_vm::TritonVmJobQueue;
 use crate::models::blockchain::block::block_height::BlockHeight;
 use crate::models::blockchain::block::mutator_set_update::MutatorSetUpdate;
 use crate::models::blockchain::transaction::primitive_witness::PrimitiveWitness;
@@ -97,7 +96,6 @@ impl UpdateMutatorSetDataJob {
 
     pub(crate) async fn upgrade(
         self,
-        triton_vm_job_queue: &TritonVmJobQueue,
         proof_job_options: TritonVmProofJobOptions,
     ) -> anyhow::Result<Transaction> {
         let UpdateMutatorSetDataJob {
@@ -112,7 +110,6 @@ impl UpdateMutatorSetDataJob {
             &old_mutator_set,
             &mutator_set_update,
             old_single_proof,
-            triton_vm_job_queue,
             proof_job_options,
             None,
         )
@@ -255,7 +252,6 @@ impl UpgradeJob {
     /// informs peers of this new transaction.
     pub(crate) async fn handle_upgrade(
         self,
-        triton_vm_job_queue: &TritonVmJobQueue,
         tx_origin: TransactionOrigin,
         perform_ms_update_if_needed: bool,
         mut global_state_lock: GlobalStateLock,
@@ -299,12 +295,7 @@ impl UpgradeJob {
             // No locks may be held here!
             let (upgraded, expected_utxos) = match upgrade_job
                 .clone()
-                .upgrade(
-                    triton_vm_job_queue,
-                    job_options,
-                    &wallet_entropy,
-                    block_height,
-                )
+                .upgrade(job_options, &wallet_entropy, block_height)
                 .await
             {
                 Ok((upgraded_tx, expected_utxos)) => {
@@ -430,7 +421,6 @@ impl UpgradeJob {
     /// prover is busy.
     pub(crate) async fn upgrade(
         self,
-        triton_vm_job_queue: &TritonVmJobQueue,
         proof_job_options: TritonVmProofJobOptions,
         own_wallet_entropy: &WalletEntropy,
         current_block_height: BlockHeight,
@@ -472,9 +462,7 @@ impl UpgradeJob {
                 })
                 .collect_vec();
             let gobbler = PrimitiveWitness::from_transaction_details(&gobbler);
-            let gobbler_proof =
-                SingleProof::produce(&gobbler, triton_vm_job_queue, proof_job_options.clone())
-                    .await?;
+            let gobbler_proof = SingleProof::produce(&gobbler, proof_job_options.clone()).await?;
             info!("Done producing gobbler-transaction for a value of {gobbling_fee}");
             let gobbler = Transaction {
                 kernel: gobbler.kernel,
@@ -496,12 +484,7 @@ impl UpgradeJob {
                 let nondeterminism = single_proof_witness.nondeterminism();
                 info!("Proof-upgrader: Start generate single proof");
                 let single_proof = SingleProof
-                    .prove(
-                        claim,
-                        nondeterminism,
-                        triton_vm_job_queue,
-                        proof_job_options.clone(),
-                    )
+                    .prove(claim, nondeterminism, proof_job_options.clone())
                     .await?;
                 info!("Proof-upgrader, to single proof: Done");
 
@@ -516,12 +499,7 @@ impl UpgradeJob {
 
                     info!("Proof-upgrader: Start merging with gobbler");
                     let ret = lhs
-                        .merge_with(
-                            rhs,
-                            gobble_shuffle_seed,
-                            triton_vm_job_queue,
-                            proof_job_options,
-                        )
+                        .merge_with(rhs, gobble_shuffle_seed, proof_job_options)
                         .await?;
                     info!("Proof-upgrader merging with gobbler: Done");
                     ret
@@ -552,7 +530,6 @@ impl UpgradeJob {
                     left,
                     right,
                     shuffle_seed.to_owned(),
-                    triton_vm_job_queue,
                     proof_job_options.clone(),
                 )
                 .await?;
@@ -560,12 +537,7 @@ impl UpgradeJob {
 
                 if let Some(gobbler) = gobbler {
                     ret = gobbler
-                        .merge_with(
-                            ret,
-                            gobble_shuffle_seed,
-                            triton_vm_job_queue,
-                            proof_job_options,
-                        )
+                        .merge_with(ret, gobble_shuffle_seed, proof_job_options)
                         .await?
                 };
 
@@ -576,8 +548,7 @@ impl UpgradeJob {
             } => {
                 info!("Proof-upgrader: Start producing proof collection");
                 let proof_collection =
-                    ProofCollection::produce(&witness, triton_vm_job_queue, proof_job_options)
-                        .await?;
+                    ProofCollection::produce(&witness, proof_job_options).await?;
                 info!("Proof-upgrader, proof collection: Done");
                 Ok((
                     Transaction {
@@ -591,8 +562,7 @@ impl UpgradeJob {
                 primitive_witness: witness,
             } => {
                 info!("Proof-upgrader: Start producing single proof");
-                let proof =
-                    SingleProof::produce(&witness, triton_vm_job_queue, proof_job_options).await?;
+                let proof = SingleProof::produce(&witness, proof_job_options).await?;
                 info!("Proof-upgrader, single proof: Done");
                 Ok((
                     Transaction {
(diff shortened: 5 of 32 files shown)


aszepieniec:
image

Despite all the above, this PR is not yet ready for merge. I expect CI to fail because I could not get the test suite to pass on my desktop computer. Specifically, the following tests fail:

```
    models::blockchain::transaction::validity::tasm::claims::generate_collect_lock_scripts_claim::tests::unit_test
    models::blockchain::transaction::validity::tasm::claims::generate_collect_type_scripts_claim::tests::unit_test
    models::blockchain::transaction::validity::tasm::claims::generate_k2o_claim::tests::unit_test
    models::blockchain::transaction::validity::tasm::claims::generate_lock_script_claim_template::test::unit_test
    models::blockchain::transaction::validity::tasm::claims::generate_rri_claim::tests::unit_test
    models::blockchain::transaction::validity::tasm::claims::generate_type_script_claim_template::test::unit_test
```

The reason why they fail is a complete mystery to me. Given that they fail after this refactor, I see no reason why they would not have failed beforehand. It seems to be a concurrency issue, so it probably is related somehow. Some pointers to track this down would be appreciated.

aszepieniec:
I just realized that #513 proposes refactors to the Triton VM job queue. Sorry for the unnecessary conflict.

All that said, I prefer this refactor (assuming I can make the failing tests work) over that one because:
 - there is no need to store or pass references to the job queue;
 - there is no need to wrap the references in a smart pointer; and
 - misuse is made impossible through the type system.

I propose to rebase branch asz/test-larger-blocks on top of this.

dan-da (commented):
I have only just looked at the first file so far, and see something that concerns me.

background:  the design of the job-queue is that it can (and should) be in its own crate.  more specifically:

```
job_queue/
├── errors.rs
├── mod.rs
├── queue.rs
├── traits.rs
└── triton_vm
    ├── mod.rs
    └── triton_vm_job_queue.rs
```

everything in the top-level job-queue directory is intended to be non-specific to neptune-core.   Only the items within job_queue/triton_vm are specialized for neptune-core.   Really the triton_vm dir should be elsewhere outside job-queue, but that's what we have for now.

I thought the above was already doc'd in a README, but I don't see one, so that's an oversight I will remedy as soon as I can.

So then, this PR should not be touching queue.rs, unless it is proposing changes to the workings of the job-queue itself, which it is not, iiuc   Only how neptune-core passes it around.

Now and going forward: If changes to the queue are actually needed for some reason, let's have a discussion around it.  Be warned I don't wish to modify that for trivial reasons, and I may push to move it into its own crate, to make it harder to break the abstraction (and to share with others of course).

So then, I suggest to move any related changes into triton_vm/triton_vm_job_queue.rs.  See for example: https://github.com/dan-da/neptune-core/blob/2980e10df2cf3c1a8aa942615847f399d542e9e1/src/job_queue/triton_vm/triton_vm_job_queue.rs#L26

aszepieniec:
> image
> 
> Despite all the above, this PR is not yet ready for merge. I expect CI to fail because I could not get the test suite to pass on my desktop computer. Specifically, the following tests fail:
> 
> ```
>     models::blockchain::transaction::validity::tasm::claims::generate_collect_lock_scripts_claim::tests::unit_test
>     models::blockchain::transaction::validity::tasm::claims::generate_collect_type_scripts_claim::tests::unit_test
>     models::blockchain::transaction::validity::tasm::claims::generate_k2o_claim::tests::unit_test
>     models::blockchain::transaction::validity::tasm::claims::generate_lock_script_claim_template::test::unit_test
>     models::blockchain::transaction::validity::tasm::claims::generate_rri_claim::tests::unit_test
>     models::blockchain::transaction::validity::tasm::claims::generate_type_script_claim_template::test::unit_test
> ```
> 
> The reason why they fail is a complete mystery to me. Given that they fail after this refactor, I see no reason why they would not have failed beforehand. It seems to be a concurrency issue, so it probably is related somehow. Some pointers to track this down would be appreciated.

Might be related to this (thanks, AI):

> Race Conditions: The `get_or_init` method of OnceLock is not guaranteed to call the initialization function only once. If multiple threads call `global_triton_vm_job_queue` simultaneously, they might all attempt to initialize the queue, leading to potential race conditions.

dan-da:
> > I just realized that #513 proposes refactors to the Triton VM job queue. Sorry for the unnecessary conflict.
> 
> unfortunate, but these things happen sometimes.

github bug?   It is saying that @aszepieniec wrote a comment that I authored (and edited).

convo

dan-da:
> Might be related to this (thanks, AI):
> 
> > Race Conditions: The `get_or_init` method of OnceLock is not guaranteed to call the initialization function only once. If multiple threads call `global_triton_vm_job_queue` simultaneously, they might all attempt to initialize the queue, leading to potential race conditions.

I think the AI might be hallucinating in this case.

from https://doc.rust-lang.org/stable/std/sync/struct.OnceLock.html#method.get_or_init

> Many threads may call get_or_init concurrently with different initializing functions, **but it is guaranteed** that only one function will be executed.

I think that is kinda the point of OnceLock.

If this were (somehow) the cause, then my nearly identical solution would likely exhibit the same test failures.  sadly tests are not compiling in my branch yet, so I can't try it now.

aszepieniec:
image

Can confirm bug.

aszepieniec:
> Now and going forward: If changes to the queue are actually needed for some reason, let's have a discussion around it. Be warned I don't wish to modify that for trivial reasons, and I may push to move it into its own crate, to make it harder to break the abstraction (and to share with others of course).

I did not realize it was the intention was to spin it off as its own crate. I think it is a neat piece of code and can benefit lots of projects if it is spun off, so I support this intention. Let's find a way to avoid modifying `queue.rs` -- or restrict modifications to changes that also make sense if it were its own crate.

To explain where I'm coming from, let me explain why modifying `queue.rs` is the most straightforward way to apply the singleton pattern to the job queue for triton vm jobs in the context of `neptune-core`. Take my understanding of rust intricacies with a grain of salt.

 - I moved the singleton-constructor, which I call `global_triton_vm_job_queue` and you called `vm_job_queue` into `queue.rs` because only from `queue.rs` can you call `start()`, which is a private function.
 - IMHO, `start()` must be a private function because otherwise users can mis-use it to accidentally create two instances (like I did).
 - You cannot put a generic `instance()` singleton-constructor function on `JobQueue<P>` because when static memory is allocated, the allocator does not know about generic type arguments. (This seems like a limitation of rust, not of principle, but it also sounds like an insanely complicated job to get right as a compiler developer.)

Here are the two options we've tried:
 1. Drop the requirement that `start()` be private. This option leaves in place the pitfall that I fell into.
 2. Drop the requirement that `queue.rs` make abstraction of `neptune-core` and Triton VM. This makes it impossible to spin off the job queue as its own crate.

This list of options is not exhaustive; there are other things we can try. And I'm sure that others have stumbled across this issue as well and solved it somehow.

dan-da:
> I did not realize it was the intention was to spin it off as its own crate. I think it is a neat piece of code and can benefit lots of projects if it is spun off, so I support this intention. Let's find a way to avoid modifying `queue.rs` -- or restrict modifications to changes that also make sense if it were its own crate.

great, thx.    With your blessing, I will plan on doing that, next moment there is some free time.

> To explain where I'm coming from, let me explain why modifying `queue.rs` is the most straightforward way to apply the singleton pattern to the job queue for triton vm jobs in the context of `neptune-core`. Take my understanding of rust intricacies with a grain of salt.

ok, let's get into the nitty gritty!

> * I moved the singleton-constructor, which I call `global_triton_vm_job_queue` and you called `vm_job_queue` into `queue.rs` because only from `queue.rs` can you call `start()`, which is a private function.

start() is a public fn in master.   evidence:

```
git show master:./job_queue/queue.rs | grep "fn.*start"
    pub fn start() -> Self {
```

I believe you removed the pub in this PR.   (and maybe forgot that?)

> * IMHO, `start()` must be a private function because otherwise users can mis-use it to accidentally create two instances (like I did).

We need to be careful here.  JobQueue itself, as designed, does not have any constraint about one queue per process.   That is a neptune-core requirement, at present.  In master, it is neptune-core specific code that introduces and enforces the "only one queue" rule.  And we could even consider relaxing that constraint in the future, eg if we decide that certain VM job-types do not require a lot of resources, they could be run in a separate JobQueue.

stated differently, placing that requirement inside queue.rs would be breaking the abstraction and would eliminate all multi-queue use cases for the JobQueue.    We can/should think of the files inside top-level job_queue dir as if they are already a separate crate we are consumers of.

Another way of looking at it is that this rule is already enforced by neptune-core specific code in master, via the single TritonVmJobQueue instance in GlobalStateLock that gets passed around.  Our task here then is to maintain the same rule, still in neptune-core specific code, but in a way that doesn't require the variable be passed through many levels of methods.

> * You cannot put a generic `instance()` singleton-constructor function on `JobQueue<P>` because when static memory is allocated, the allocator does not know about generic type arguments. (This seems like a limitation of rust, not of principle, but it also sounds like an insanely complicated job to get right as a compiler developer.)

I think this should be a non-goal, per above.

> Here are the two options we've tried:
> 
> 1. Drop the requirement that `start()` be private. This option leaves in place the pitfall that I fell into.

It's never had that requirement, except in this PR.   Somehow I am missing the pitfall....??

> 2. Drop the requirement that `queue.rs` make abstraction of `neptune-core` and Triton VM. This makes it impossible to spin off the job queue as its own crate.

job-queue, as it exists in master, can be placed into a new crate and be usable for other projects right now, by simply removing the triton_vm sub-directory.   That is the property I'm trying to preserve.

> This list of options is not exhaustive; there are other things we can try. And I'm sure that others have stumbled across this issue as well and solved it somehow.

I don't understand what is wrong with the simple solution I already have in my branch:

in job_queue/triton_vm/triton_vm_job_queue.rs
```
/// returns reference-counted clone of the triton vm job queue.
///
/// callers should execute resource intensive triton-vm tasks in this
/// queue to avoid running simultaneous tasks that could exceed hardware
/// capabilities.
pub fn vm_job_queue() -> Arc<TritonVmJobQueue> {
    static JOB_QUEUE_LOCK: OnceLock<Arc<TritonVmJobQueue>> = OnceLock::new();

    JOB_QUEUE_LOCK
        .get_or_init(|| Arc::new(TritonVmJobQueue::start()))
        .clone()
}
```

This is pretty much identical to your fn, except it exists in a *neptune-specific* directory and file, and thus does not break the JobQueue abstraction.

I'm guessing maybe what has happened is that you've forgotten that JobQueue::start() is pub in master, and so you are devising ways to make it work with that method private, but really that's unnecessary, because it is already public, as per original design.

Or if I'm wrong with that guess, then please let me know if you see something wrong with the vm_job_queue() method above, or its placement, because afaict, it works just fine, and queue.rs is unmodified.  Everything compiles at least, without tests.  I haven't tried running a node and sending tx yet.

aszepieniec:
> With your blessing, I will plan on doing that, next moment there is some free time.

:+1: 

> > Drop the requirement that start() be private. This option leaves in place the pitfall that I fell into.
> 
> It's never had that requirement, except in this PR. Somehow I am missing the pitfall....??

This true anecdote illustrates the pitfall: I made the refactor contained in branch `asz/test-larger-blocks` in good faith and with all my attention devoted to it. I tried my best to follow the intended use patterns of the code that was already there, as best as I could infer it from examples and documentation. I am not stupid, and neither are the two reviewers who reviewed it before you did. And yet, that PR undermines the very point of having a job queue for Triton VM jobs by creating and using two instances of it. In hindsight, after you identified the issue and why it was a problem, it was obvious -- but not beforehand. I posit that the reason why you found the issue was because you had inarticulate familiarity with the job queue. If that's true, *only you* could have found it, and *only you* could upgrade it in any which way. We avoided a catastrophe in this particular instance, but I can't begin to list the reasons why the right domain-specific knowledge might be absent when the next one should be caught.

The pitfall, phrased concisely, is the danger of accidentally modifying the way the job queue is used in such a way that undermines the reason for its existence.

The issue with the design pattern you advocate is that this pitfall persists. We avoided a catastrophe in this particular anecdote, but I can't begin to list the reasons why the right domain-specific knowledge might be absent when the next one should be caught. A robust solution to this problem is a code architecture that makes it difficult or preferably even impossible to use the triton vm job queue differently from how it was intended.

So far very abstract. Concretely:
 - Passing and storing `Arc`s suggests that there are multiple triton vm job queues and the smart pointers are needed to disambiguate. As there is only ever one, a reference to it should be produced when its necessary and not before. There shouldn't be any `Arc`s.
 - Function `start()` is `pub`, meaning that you can call it twice or even more often in different locations and accidentally create two triton vm job queues or even more when the intention is for there to only ever be one. (I realize that that function is also marked `pub` on master but I think this is an error which is why this PR corrects it. I also realize that `start()` is constrained by the right interface of an off-spinnable job queue crate that makes abstraction over the type of jobs it serializes; but I think there is no contradiction between that constraint and my bullet point -- more on that later.)

> We need to be careful here. JobQueue itself, as designed, does not have any constraint about one queue per process, nor should it. That is a neptune-core requirement, at present. In master, it is neptune-core specific code that (a) specializes JobQueue to JobQueue<TritonVmPriority> and (b) introduces and enforces the "only one VM queue" rule. And we could even consider relaxing that rule in the future, eg if we decide that certain VM job-types do not require a lot of resources, they could be run in a separate JobQueue<TritonVmPriority>. Or we could have another JobQueue specialization entirely, such as for the main_loop timer "jobs".

I wonder if these alternative uses for `JobQueue` are hypothetical, or if you've already implemented them or others that you omitted. If hypothetical, then the only available data point about how `JobQueue` is used in practice is as a singleton. You could make an argument that that data justifies revisiting the design -- specifically, maybe `JobQueue` should *always* be a singleton; and if the alternatives where multiple queues make sense are ever realized, they will need some other abstraction at that point. Food for thought.

That said, I'm happy to proceed under the premise that *some* specializations of `JobQueue` *do* need to admit multiple instances of the same type. It follows that `start()` needs to be marked `pub` to accommodate for that.

Bear with me in my train of thought. We have the following constraints.
 - `JobQueue::<P>::start()` must be `pub`.
 - It must be impossible to create multiple instances of `TritonVmJobQueue`.

It follows that
 - `TritonVmJobQueue` does not have a method named `start` *and* marked `pub`;
 - `TritonVmJobQueue` is not a specialization of `JobQueue`.

A possible resolution (that avoids code duplication):
- `TritonVmJobQueue` is a newtype wrapper around `JobQueue<TritonVmPriority>`.

Since turning a `JobQueue<P>` into a singleton type with essentially the same interface except for `start()` seems like a useful generic transformation (even though we only need it for `P = TritonVmPriority`) I propose to extend the interface of `job_queue` with an access point to activate this transformation. I think it means writing and exposing a macro.

I'll implement this proposal next and push it to this branch so we have something concrete to discuss.

aszepieniec:
Some relevant snippets / notes:

In `triton_vm_job_queue.rs`:

```rust
singleton_job_queue! {
    #[doc = "A singleton job queue for Triton VM jobs."]
    TritonVmJobQueue = JobQueue<TritonVmJobPriority>
}
```

In `program.rs`:

```rust
    let job_queue = TritonVmJobQueue::instance();
    let job_handle = job_queue.add_job(Box::new(job), proof_job_options.job_priority)?;
```

File `queue.rs` is now identical to what it is on master. There is a new file `singleton.rs` which contains the singleton generator macro.

dan-da:
preamble:  I think we are coming to some agreement towards a solution, at the bottom.  ;-)

> The pitfall, phrased concisely, is the danger of accidentally modifying the way the job queue is used in such a way that undermines the reason for its existence.

I stated above that I thought this was already clearly documented in a README, and I will correct that asap.

> The issue with the design pattern you advocate is that this pitfall persists. We avoided a catastrophe in this particular anecdote, but I can't begin to list the reasons why the right domain-specific knowledge might be absent when the next one should be caught. A robust solution to this problem is a code architecture that makes it difficult or preferably even impossible to use the triton vm job queue differently from how it was intended.

I finally figured out the pitfall you are referring to.  see below.   I don't think the issue is as dire as you make it sound, especially with proper documentation.  But sure, I'm happy if we can prevent multiple instantiation entirely.

> So far very abstract. Concretely:
> 
> * Passing and storing `Arc`s suggests that there are multiple triton vm job queues and the smart pointers are needed to disambiguate. As there is only ever one, a reference to it should be produced when its necessary and not before. 

Arc is useful for allowing multiple references to a single object.  That is why it exists.  It does not suggest there are multiple instances that need disambiguation, but rather that a given instance can be easily used by multiple callers.

Still, other things being equal, I prefer a solution without Arc.

> There shouldn't be any `Arc`s.

I don't necessarily agree with that as a blanket statement.   Rather, Arc is one of several possible tools in the toolbox.

> * Function `start()` is `pub`, meaning that you can call it twice or even more often in different locations and accidentally create two triton vm job queues or even more when the intention is for there to only ever be one.

Just as one can call HashSet<u32>::new() and get two hash-sets, even if you only want one in your program.

> (I realize that that function is also marked `pub` on master but I think this is an error which is why this PR corrects it.

It is not an error.  That is the intent.  I've been clear about that.

> > We need to be careful here. JobQueue itself, as designed, does not have any constraint about one queue per process, nor should it. That is a neptune-core requirement, at present. In master, it is neptune-core specific code that (a) specializes JobQueue to JobQueue and (b) introduces and enforces the "only one VM queue" rule. And we could even consider relaxing that rule in the future, eg if we decide that certain VM job-types do not require a lot of resources, they could be run in a separate JobQueue. Or we could have another JobQueue specialization entirely, such as for the main_loop timer "jobs".
> 
> I wonder if these alternative uses for `JobQueue` are hypothetical, or if you've already implemented them or others that you omitted.

Like with anything one builds, future uses are hypothetical until they are built.  If I am writing HashMap<T> I don't have to implement 10 different kinds of HashMap specializations to understand that my HashMap can be used to hold different types.  And that's the point of making it generic in the first place.

When designing JobQueue, I knew I could not predict every possible type of job or usage.  As such, I intended from the very start that it would be a generic piece of code, that could accomodate different usage scenarios.   Those include different types of jobs, as well as single or multiple queues.

I enjoy writing code that can be useful beyond my own projects, and so I also intended from the start to make this its own crate.  However, we were pretty hurried, and it was faster just to stick it in a sub-directory for the moment.

> If hypothetical, then the only available data point about how `JobQueue` is used in practice is as a singleton. You could make an argument that that data justifies revisiting the design -- specifically, maybe `JobQueue` should _always_ be a singleton; and if the alternatives where multiple queues make sense are ever realized, they will need some other abstraction at that point. Food for thought.

I could make that argument for any generic type or component if I am only using one in my app.   But the second I want to use a 2nd one, that goes out the window.

> That said, I'm happy to proceed under the premise that _some_ specializations of `JobQueue` _do_ need to admit multiple instances of the same type. It follows that `start()` needs to be marked `pub` to accommodate for that.

thx, glad to hear it!

> Bear with me in my train of thought. We have the following constraints.
> 
> * `JobQueue::<P>::start()` must be `pub`.
> * It must be impossible to create multiple instances of `TritonVmJobQueue`.

We can consider those design constraints for usage within neptune-core, sure.   I don't see the 2nd constraint as being entirely necessary because we could also just document very clearly that "there shall only be one", promise floggings for instantiating a second one, and so on.  But hey, if we can achieve it cleanly, and it prevents some floggings, why not?

I would like to add a constraint, or modify the first one:

* this design should take place entirely in neptune-core code ie (src/job_queue/triton_vm), not touching any of the generic job-queue files ie (src/job_queue/*.rs)

> It follows that
> 
> * `TritonVmJobQueue` does not have a method named `start` _and_ marked `pub`;
> * `TritonVmJobQueue` is not a specialization of `JobQueue`.

Ok, I think I finally see the issue or pitfall you are trying to resolve.  since:

```
pub type TritonVmJobQueue = JobQueue<TritonVmJobPriority>;
```

then one *could* call `TritonVmJobQueue::start()` from anywhere, even though a method like vm_job_queue() might exist, which always returns a single copy.

ok, it's a fun design challenge.

> A possible resolution (that avoids code duplication):
> 
> * `TritonVmJobQueue` is a newtype wrapper around `JobQueue<TritonVmPriority>`.

yes, I was just thinking this as well.

> Since turning a `JobQueue<P>` into a singleton type with essentially the same interface except for `start()` seems like a useful generic transformation (even though we only need it for `P = TritonVmPriority`) I propose to extend the interface of `job_queue` with an access point to activate this transformation. I think it means writing and exposing a macro.

I think this can be done more simply with Deref, and DerefMut, eg:

```
struct TritonVmJobQueue(JobQueue<TritonVmPriority>)

impl Deref for TritonVmJobQueue {
    type Target = JobQueue<TritonVmPriority>;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}

impl DerefMut for TritonVmJobQueue {
    fn deref_mut(&mut self) -> &mut Self::Target {
        &mut self.0
    }
}

impl TritonVmJobQueue {
    pub fn get_instance() -> &'static Self {
        static INSTANCE: OnceLock<Self> = OnceLock::new();
        INSTANCE.get_or_init(|| Self(JobQueue<TritonVmPriority>::start()))
    }
}
```

this solution meets all constraints:

1. can be done entirely in neptune-core code.
2. It's not possible to create a second instance of TritonVmJobQueue  (*)
3. it does not use Arc, but can be used from anywhere to access the single vm-job-queue.

what do you think?

(*) One could still do `JobQueue<TritonVmPriority>::start()` if one really wants to.  It's technically not a TritonVmJobQueue, but it would allow processing the same jobs.   I am perfectly fine with allowing that possibility, as one almost has to be malicious to do it.

dan-da:
> preamble: I think we are coming to some agreement towards a solution, at the bottom. ;-)
> 
> > The pitfall, phrased concisely, is the danger of accidentally modifying the way the job queue is used in such a way that undermines the reason for its existence.
> 
> I stated above that I thought this was already clearly documented in a README, and I will correct that asap.
> 
> > The issue with the design pattern you advocate is that this pitfall persists. We avoided a catastrophe in this particular anecdote, but I can't begin to list the reasons why the right domain-specific knowledge might be absent when the next one should be caught. A robust solution to this problem is a code architecture that makes it difficult or preferably even impossible to use the triton vm job queue differently from how it was intended.
> 
> I finally figured out the pitfall you are referring to. see below. I don't think the issue is as dire as you make it sound, especially with proper documentation. But sure, I'm happy if we can prevent multiple instantiation entirely.
> 
> > So far very abstract. Concretely:
> > 
> > * Passing and storing `Arc`s suggests that there are multiple triton vm job queues and the smart pointers are needed to disambiguate. As there is only ever one, a reference to it should be produced when its necessary and not before.
> 
> Arc is useful for allowing multiple references to a single object. That is why it exists. It does not suggest there are multiple instances that need disambiguation, but rather that a given instance can be easily used by multiple callers.
> 
> Still, other things being equal, I prefer a solution without Arc.
> 
> > There shouldn't be any `Arc`s.
> 
> I don't necessarily agree with that as a blanket statement. Rather, Arc is one of several possible tools in the toolbox.
> 
> > * Function `start()` is `pub`, meaning that you can call it twice or even more often in different locations and accidentally create two triton vm job queues or even more when the intention is for there to only ever be one.
> 
> Just as one can call HashSet::new() and get two hash-sets, even if you only want one in your program.
> 
> > (I realize that that function is also marked `pub` on master but I think this is an error which is why this PR corrects it.
> 
> It is not an error. That is the intent. I've been clear about that.
> 
> > > We need to be careful here. JobQueue itself, as designed, does not have any constraint about one queue per process, nor should it. That is a neptune-core requirement, at present. In master, it is neptune-core specific code that (a) specializes JobQueue to JobQueue and (b) introduces and enforces the "only one VM queue" rule. And we could even consider relaxing that rule in the future, eg if we decide that certain VM job-types do not require a lot of resources, they could be run in a separate JobQueue. Or we could have another JobQueue specialization entirely, such as for the main_loop timer "jobs".
> > 
> > 
> > I wonder if these alternative uses for `JobQueue` are hypothetical, or if you've already implemented them or others that you omitted.
> 
> Like with anything one builds, future uses are hypothetical until they are built. If I am writing HashMap I don't have to implement 10 different kinds of HashMap specializations to understand that my HashMap can be used to hold different types. And that's the point of making it generic in the first place.
> 
> When designing JobQueue, I knew I could not predict every possible type of job or usage. As such, I intended from the very start that it would be a generic piece of code, that could accomodate different usage scenarios. Those include different types of jobs, as well as single or multiple queues.
> 
> I enjoy writing code that can be useful beyond my own projects, and so I also intended from the start to make this its own crate, if allowed. However, we were pretty hurried, and it was faster just to stick it in a sub-directory for the moment.
> 
> > If hypothetical, then the only available data point about how `JobQueue` is used in practice is as a singleton. You could make an argument that that data justifies revisiting the design -- specifically, maybe `JobQueue` should _always_ be a singleton; and if the alternatives where multiple queues make sense are ever realized, they will need some other abstraction at that point. Food for thought.
> 
> I could make that argument for any generic type or component if I am only using one in my app. But the second I want to use a 2nd one, that goes out the window.
> 
> > That said, I'm happy to proceed under the premise that _some_ specializations of `JobQueue` _do_ need to admit multiple instances of the same type. It follows that `start()` needs to be marked `pub` to accommodate for that.
> 
> thx, glad to hear it!
> 
> > Bear with me in my train of thought. We have the following constraints.
> > 
> > * `JobQueue::<P>::start()` must be `pub`.
> > * It must be impossible to create multiple instances of `TritonVmJobQueue`.
> 
> We can consider those design constraints for usage within neptune-core, sure. I don't see the 2nd constraint as being entirely necessary because we could also just document very clearly that "there shall only be one", promise floggings for instantiating a second one, and so on. But hey, if we can achieve it cleanly, and it prevents some floggings, why not?
> 
> I would like to add a constraint, or modify the first one:
> 
> * this design should take place entirely in neptune-core code ie (src/job_queue/triton_vm), not touching any of the generic job-queue files ie (src/job_queue/*.rs)
> 
> > It follows that
> > 
> > * `TritonVmJobQueue` does not have a method named `start` _and_ marked `pub`;
> > * `TritonVmJobQueue` is not a specialization of `JobQueue`.
> 
> Ok, I think I finally see the issue or pitfall you are trying to resolve. since:
> 
> ```
> pub type TritonVmJobQueue = JobQueue<TritonVmJobPriority>;
> ```
> 
> then one _could_ call `TritonVmJobQueue::start()` from anywhere, even though a method like vm_job_queue() might exist, which always returns a single copy.
> 
> ok, it's a fun design challenge.
> 
> > A possible resolution (that avoids code duplication):
> > 
> > * `TritonVmJobQueue` is a newtype wrapper around `JobQueue<TritonVmPriority>`.
> 
> yes, I was just thinking this as well.
> 
> > Since turning a `JobQueue<P>` into a singleton type with essentially the same interface except for `start()` seems like a useful generic transformation (even though we only need it for `P = TritonVmPriority`) I propose to extend the interface of `job_queue` with an access point to activate this transformation. I think it means writing and exposing a macro.
> 
> I think this can be done more simply with Deref, and DerefMut, eg:
> 
> ```
> struct TritonVmJobQueue(JobQueue<TritonVmPriority>)
> 
> impl Deref for TritonVmJobQueue {
>     type Target = JobQueue<TritonVmPriority>;
> 
>     fn deref(&self) -> &Self::Target {
>         &self.0
>     }
> }
> 
> impl DerefMut for TritonVmJobQueue {
>     fn deref_mut(&mut self) -> &mut Self::Target {
>         &mut self.0
>     }
> }
> 
> impl TritonVmJobQueue {
>     pub fn get_instance() -> &'static Self {
>         static INSTANCE: OnceLock<Self> = OnceLock::new();
>         INSTANCE.get_or_init(|| Self(JobQueue<TritonVmPriority>::start()))
>     }
> }
> ```
> 
> this solution meets all constraints:
> 
> 1. can be done entirely in neptune-core code.
> 2. It's not possible to create a second instance of TritonVmJobQueue  (*)
> 3. it does not use Arc, but can be used from anywhere to access the single vm-job-queue.
> 
> what do you think?
> 
> (*) One could still do `JobQueue<TritonVmPriority>::start()` if one really wants to. It's technically not a TritonVmJobQueue, but it would allow processing the same jobs. I am perfectly fine with allowing that possibility, as one almost has to be malicious to do it.

aszepieniec:
Alright, I think we're generally on one page but there is still an epsilon of disagreement.

In particular, while the macro singleton-type generator in `singleton.rs` does live in `job_queue/` and not the subdirectory `triton_vm/`, it *is* fully generic and makes complete abstraction of the job type.

I think it is a nice feature of a job queue crate that it gives the user the option to generate a) a job queue type that admits multiple instances; or b) a job queue type that guarantees there is only one instance.

> > I think it means writing and exposing a macro.
>
> I think this can be done more simply with Deref, and DerefMut

The macro emits code that is identical to what you proposed but also emits an `impl` block defining `instance()`, which returns a reference to the one instance and generates it if it does not exist yet. The reason why you need a macro is to implement this function.

In the end, the macro and its inclusion in `job_queue/` is not a hill I'm willing to die on. I'd say it's your call to make. But I would like to ask you to give it some serious consideration before you decide to reject it. As I see it, including it would be
 - consistent with your argument in favor of genericity in anticipation of hypothetical use cases, and
 - consistent with your counter-argument against my food-for-thought point about having only one data point.
And if you decide to reject it, I wonder what other new arguments you find more compelling than the ones you originally came up with.

PR #526, opened 2025-03-24
Pin version of build-breaking transitive dependency

jan-ferdinand:
- introduce a workflow to check that `neptune-core` can be built in the absence of a `Cargo.lock`, as is the case when using it as a dependency
- fix the version of a transitive dependency that recently started introducing build failures


aszepieniec (approved):
LGTM

PR #527, opened 2025-03-24
Add functionality to build a big transaction

Sword-Smith:
If you want to build a really big transaction, with 50+ inputs and be sure that it gets included in *one* block and not spread over multiple, you will probably have to construct multiple `SingleProof`-backed transactions and have those merged. In order to prevent these partial transactions from ending up in other peer's mempool, and to prevent new blocks from disturbing the proof generation, this PR adds three RPC endpoints: Two to pause and resume state updates, and one to clear the mempool. Combined, these RPC endpoints allow the user to create as big transactions as they want, only limited by the consensus rules.


Sword-Smith:
@aszepieniec thinks the use case for this functionality is marginal. So let's wait and see if there's any demand for it outside of our team. Commit for clearing the mempool was added to master through cherry-picking.

PR #530, opened 2025-03-25
feat(mempool): Filter transactions for mutator set match

Sword-Smith:
When returning a list of transactions for inclusion in a block, only those transactions with a mutator set hash matching the expected value should be returned, as any other transactions cannot directly be merged into a block.


Sword-Smith:
Reviewed and approved by @aszepieniec on Thor's computer.

PR #531, opened 2025-03-25
fix(proof_upgrader): Don't upgrade for less than min gobbling fee

Sword-Smith:
Prior to this commit, proof-upgrading was done pro bono by everyone, for any transactions. With this commit, the `min_gobbling_fee` CLI parameter is respected.


Sword-Smith:
Reviewed and approved by @aszepieniec on Thor's computer.

PR #537, opened 2025-03-27
privacy(mine_loop): Never add zero-valued composer outputs

Sword-Smith:
If a block is empty, i.e. contains no inputs and only the coinbase rewards, *and* the composer chooses to reward the entire block subsidy to the guesser, there is no need for the composer to add any outputs to the coinbase transaction.

Prior to this commit, the composer would in that scenario add two ouputs to the coinbase transaciton. Many such blocks with two zero-valued outputs exist on main net today, so this is not a hypothetical scenario.

Avoiding these zero-valued outputs has two benefits:
 - It leaves more room for other transactions to be merged into the block.
 - It avoids zero-valued leafs in the AOCL whose presence would otherwise reduce the anonymity set as zero valued leafs are less likely to be spent than any other leafs and can, in the case of an empty block, be identified in the AOCL.

### DIFF
--- a/src/mine_loop.rs
+++ b/src/mine_loop.rs
@@ -11,6 +11,7 @@ use block_header::MINIMUM_BLOCK_TIME;
 use composer_parameters::ComposerParameters;
 use futures::channel::oneshot;
 use num_traits::CheckedSub;
+use num_traits::Zero;
 use primitive_witness::PrimitiveWitness;
 use rand::rngs::StdRng;
 use rand::Rng;
@@ -382,7 +383,7 @@ pub(crate) async fn make_coinbase_transaction_stateless(
         composer_parameters,
         timestamp,
         job_options.job_settings.network,
-    )?;
+    );
 
     let witness = PrimitiveWitness::from_transaction_details(&transaction_details);
 
@@ -413,7 +414,9 @@ pub(crate) async fn make_coinbase_transaction_stateless(
     Ok((transaction, composer_outputs))
 }
 
-/// Produce two outputs spending a given portion of the coinbase amount.
+/// Produce outputs spending a given portion of the coinbase amount. Returns
+/// two outputs unless this composition awards the entire coinbase amount to
+/// the guesser, in which case zero outputs are returned.
 ///
 /// The coinbase amount is usually set to the block subsidy for this block
 /// height.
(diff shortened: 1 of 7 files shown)


src/mine_loop.rs
```diff
@@ -427,23 +430,27 @@ pub(crate) async fn make_coinbase_transaction_stateless(
...
-        bail!(
-            "Guesser fee may not exceed coinbase amount. coinbase_amount: {}; guesser_fee: {}; guesser fee fraction: {}.",
-            coinbase_amount.to_nau(),
-            guesser_fee.to_nau(),
-            composer_parameters.guesser_fee_fraction()
-        );
-    };
+    let amount_to_composer = coinbase_amount
+        .checked_sub(&guesser_fee)
+        .expect("total_composer_fee cannot exceed coinbase_amount");
```
dan-da, line 449: 1. previously this was returning an error.  now it panics.  why?
2. the expect message seems wrong.  I think it should be: "guesser_fee cannot exceed coinbase_amount".  which also matches the pre-existing error message.

src/mine_loop.rs
```diff
@@ -427,23 +430,27 @@ pub(crate) async fn make_coinbase_transaction_stateless(
...
+///
+/// # Panics
+///
+/// If the provided guesser fee fraction is not between 0 and 1 (inclusive).
 pub(crate) fn composer_outputs(
     coinbase_amount: NativeCurrencyAmount,
     composer_parameters: ComposerParameters,
     timestamp: Timestamp,
-) -> Result<[TxOutput; 2]> {
+) -> TxOutputList {
```
dan-da, line 443: why go from returning a Result and error(s) to not returning a Result and panicking?

Sword-Smith: No behavior has changed. Only the function signature. The function could not fail, as the `lossy_f64_fraction_mul` function already asserts that `composer_parameters.guesser_fee_fraction()` is a fraction.

src/mine_loop.rs
Sword-Smith: That `bail!` was dead code -- could never be reached because of the above expression ` coinbase_amount.lossy_f64_fraction_mul(composer_parameters.guesser_fee_fraction());`

Sword-Smith: You're right about the wrong error message. Sorry about that.

PR #554, opened 2025-04-17
`BTreeMap` for  `indices_to_hash_map`

skaunov:
This stems from #543 but focuses on the different aspect. It would not be much reasonable to change a function because of a test, but I looked into itself and that seems it only benefits from switching, as this improves `find`. (The `proptest` benefits by the determinism of `BTreeMap`.)

### DIFF
--- a/proptest-regressions/util_types/mutator_set/removal_record.txt
+++ b/proptest-regressions/util_types/mutator_set/removal_record.txt
@@ -1,8 +1,0 @@
-# Seeds for failure cases proptest has generated in the past. It is
-# automatically read and these particular cases re-run before any
-# novel cases are generated.
-#
-# It is recommended to check this file in to source control so that
-# everyone who runs the test benefits from these saved cases.
-cc 1b31f91eeca508bf7a47e528046a353eba79e825fa4ff049bb23bfb5575ae850 # shrinks to input = _RemovalRecordMissingChunkElementIsInvalidPbtArgs { initial_additions: 58, index_to_drop: 1, to_remove_i: 9066402313039900099 }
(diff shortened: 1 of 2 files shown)


aszepieniec:
Can you be more specific about what is improved? In general, accesses in a binary tree map take $O(\log n)$ whereas accesses in a hash map take $O(1)$ time, so I'd like to understand the apparent performance improvement properly before approving this PR.

skaunov:
@aszepieniec 
My intuition was that both `partition` and `find` operations in `split_by_activity` do in essence an `u64` comparison so finding the correct root in a tree or checking the right most leaf should be easier than going over all the `.keys()` arbitrary. But that's just intuition of mine; I didn't dig for an evidence. :speak_no_evil:

aszepieniec:
Let's collect evidence so that we do not have to rely on intuition to determine whether merging this PR is a good idea.

Or we can file this PR under "too low priority for now" and shelve it until we need it.

skaunov:
Sounds good to me. I can redo it with `.last()` instead of `.find` and 
a range set instead of `partition` to make retrieval benefit obvious. 
Indicate if you would want to merge it then, pls. As for inserting 
comparison the whole issue feels too small to actually analyze and 
compare. I even suspect that with such small lengths of the set the 
comparison won't be definite and mostly reflect other things like the 
platform, `rustc` version, hardware, and such.

aszepieniec:
> Sounds good to me. I can redo it with `.last()` instead of `.find` and
> a range set instead of `partition` to make retrieval benefit obvious.
> Indicate if you would want to merge it then, pls.

I don't know, it's not obvious before I can see the code. I can only pass judgment afterwards.

I'm happy shelving this discussion until it is clear we need this particular thing to be faster. I have a hunch that other things are much more deserving of our attention when it comes to optimization.

aszepieniec:
Closed: shelved indefinitely.

Issue #553, opened 2025-04-17
design and implement DB migration strategy.


dan-da:
#552 has highlighted the importance of keeping track of which of our data-types are stored in the various database and then versioning changes to the database schema and types (somehow) so that either the code remains backwards compatible with older DB files or DB files get migrated to newer schemas.

Here are some considerations:

1. We have multiple levelDB databases (files).   Each can be considered to have its own schema that should be versioned.
2. Most DB files use the `storage_schema` types, but one (or more) are just basic key/val.   So there is a mix.
3. Just reading the rust code, it is not readily apparent which types are stored in the database, and that can change over time.  Eg TxOutput was not when first created, but now is due to inclusion in `SentTransaction`.
4. It can sometimes be difficult to make types back-wards compatible when adding new fields due to serde deserialization complexities.  see #552.

Here is a high level design for a DB migration strategy:

1. We could audit the codebase and comment every type that gets stored in the database, so it is clear that changes to the type require special care and additional work and testing.  A cleaner alternative may be to create a db-types layer with "mirror" types that are stored in the DB and impl Into<ReflectedType>.  It then becomes obvious that any changes in this layer require a DB version bump.
2. each DB should have a key "schema-version" that defines the present version.   This can be a simple integer, starting with 1.
3. If "schema-version" key is not present it is treated as version 0, ie before versioning began.
4. neptune-core holds a struct with a field for each DB that identifies the schema version used by neptune-core.
5. Every official release that modifies a data-type used in a DB must bump the appropriate DB version in neptune-core and provide a function for migrating the database from previous version to the new version.   This would likely entail having a copy of the old data-type eg a struct with 5 fields which gets deserialized from the DB, converted into the new data type eg struct with 6 fields, and then saved back into the database.   Exact details are TBD.
6. at neptune-core startup, the databases are opened and the schema-version checked for each.  If any DB has an old version then a migration task is started for that DB.
7. migrations are applied in sequence.  if a given db is at version 5 and neptune-core is at version 8, then migrations 6, 7, 8 get applied.
8. We should have integration test(s) that apply all the database migrations in sequence and serialize/deserialize each type at each version.

WIth this type of strategy in place, the rest of the neptune-core code can simply assume that the databases are up to date and do not need to worry about backwards compatibility.

Note that this design does not provide any facility for downward migration.  That would add a lot more maintenance work and should not be necessary anyway.

dan-da:
[...]

PR #555, opened 2025-04-20
fix: migrate wallet-db to schema version 1

### LINKED ISSUE #552: `neptune-cli send` error
Image

dan-da:
I feel this is quite solid, and thus am going to merge immediately before conflicts occur.   I have made it a PR for easy review and comments by anyone interested.   Since this introduces a novel db versioning and migration system, it seems important for all team members to be aware, at least.  The commit can be reverted if any serious problems are found.

-------------

fixes #552
addresses #553

1. Adds an infrastructure for versioning the wallet DB and migrating older databases to latest version.  Also detects and panics if a newer database schema is used, eg from a newer neptune-core.  #553.

2. Provides documentation for the versioning and migration system.

3. Fixes #552 by creating a DB migration module from v0 to v1. The migration defaults the TxOutput::is_change field to the value of the owned field, which is a best-guess heuristic, since most owned outputs are probably change, in general, and at this point in time.

4. Includes a test that generates a mock v0 database containing a SentTransaction and migrates it to a v1 database.

squashed commits:

* first attempt to migrate wallet db schema.  untested
* add db migration test
* reset schema. take table_count into account
* add fn to dump wallet database, for debugging
* set() upgraded tx, rather than push().  migrate test is working now
* split migration into modules
* improve docs


jan-ferdinand:
Does it make sense to add the following to CI?

1. checkout out tag `release`
1. build
1. create a wallet and all necessary databases
1. checkout `master` (or the latest branch if CI runs for a PR)
1. build
1. perform integration tests (or something else that uses the databases)

I'm not sure whether any of our tests persist anything to the databases.

dan-da:
It's an interesting suggestion.   It could be useful, but also I have other things on my plate and know basically nothing about CI.  So it might depend on if you want to take it on or not.  Or maybe file an issue/suggestion for later.

The test may be a bit redundant because I think it is important that every schema-version upgrade include at least one test which migrates an *actual* database from the previous schema version.   The way I've done this is to copy the DB into `test_data/migrations/wallet_db/v0_to_v1/wallet_db.v0-with-sent-tx`.

I did have to disable that test in windows because it couldn't read the linux-created DB.  So your suggestion would work around that issue and not be redundant for windows platform. 

> I'm not sure whether any of our tests persist anything to the databases.

not really.  technically yes because only on-disk DBs are supported.  But typically each test runs in its own randomly generated test-directory path.  So two invocations of the same test use different base-dir and thus a different database.

However, each test can do whatever it wants.  So tests could be written that use some pre-defined path.  devil might be in the details with CI integration, and also would this test *only* run in CI, or also on dev machines?

jan-ferdinand:
Considering the points you raise, it seems to me that my suggestion is not worth the effort.

- You have already created a test, and arguably, are more targeted one.
- My suggestion would be basically confined to CI; it would not work locally.[^act]
- The engineering and maintenance effort of my suggestion seems high.[^ci]

[^act]: Discounting tools like [act](https://github.com/nektos/act).
[^ci]:  Also, I don't particularly enjoy maintaining CI scripts.

PR #558, opened 2025-04-21
test: Verify SingleProof program doesn't change unintentionally

Sword-Smith:
Add a snapshot of the program digest of SingleProof to ensure that this value does not unintentionally change.


jan-ferdinand (approved):

aszepieniec:
I prefer #561, which has largely the same effect.

That PR:
 - introduces no new dependencies
 - works for all consensus programs

PR #561, opened 2025-04-23
test: Catch changes in `ConsensusProgram`s

aszepieniec:
Add tests for every `ConsensusProgram` that hardcodes the current program hash and fails if the program changes.

Alternative to #558.


src/models/proof_abstractions/tasm/program.rs
```diff
@@ -657,4 +657,18 @@ pub mod test {
             .write_all(&proof_data)
             .expect("cannot write to file");
     }
+
```
dan-da, line 660: please add a doc-comment explaining purpose and intended usage.

In particular, the usage seems a bit novel in that I see it called within test fn, and it also creates and executes `#[test`] fn (internal to caller), whch presumably gets executed by the test-runner.

src/models/proof_abstractions/tasm/program.rs
```diff
@@ -657,4 +657,18 @@ pub mod test {
...
+    #[macro_export]
+    macro_rules! catch_change_in_program {
+        ($consensus_program: expr, $hash_hex: literal) => {
```
jan-ferdinand, line 663: Nits only:

- I usually see no whitespace between the colon and the fragment specifier (“`expr`”, “`literal`”). Minor minor, though. 🤷
- I suggest to add `$(,)?` to allow an optional trailing comma.

```suggestion
        ($consensus_program:expr, $hash_hex:literal $(,)?) => {
```

jan-ferdinand (approved):
Nits & questions only:

- The kind of test this PR adds are snapshot tests. I suggest to name the macro accordingly, for example, `test_program_snapshot` or `snapshot_assert_eq_program_digest` (although that's probably too long).
- I'm always confused about the `#[macro_export]` and `#[macro_use]` annotations. Do you happen to know the difference(s) to doing `pub(crate) use my_macro`? I've filed those annotations as “rust edition 2015 arcana”, but that picture might be incomplete.

aszepieniec:
Comments addressed in 579a214aa3ff812e4c0ef67728fb21ee43dd9b50. Merge pending CI approval.

PR #562, opened 2025-04-24
increase a test timeout so hopefully windows CI passes.

dan-da:
Increasing a test timeout from 5 to 30 secs in the hope it will enable the test to consistently pass CI for windows.

------

will merge if/when all the integration tests in send_and_receive_basic.rs pass on all platforms.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,6 +25,13 @@ jobs:
         with:
           components: rustfmt, clippy
 
+      # Necessary for integration tests where p2p nodes connect to each other.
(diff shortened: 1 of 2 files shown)


dan-da:
nope.   still failing to connect to peer node even with 30 second timeout.  so it seems like something else is at play.   It's weird because (a) it works on the other platforms and (b) it worked on windows also previously.   I will review latest commits for anything that might be related.

dan-da:
fixed in 8608cf6c4cc9a76cc96a1937aa70f15b339aee69.

This PR was a bust, so closing it.

PR #569, opened 2025-04-26
use common tokio Runtime and TritonVmJobQueue for unit tests

### LINKED ISSUE #551: make async unit tests share a single job-queue instance.
We want unit tests to share a single job-queue instance.  This would serialize generation of all proofs, even when running tests in parallel (which is the default).

In order to share a single job-queue instance, the tests must share a single tokio runtime.

This can be done by creating a function like:

```rust
pub fn tokio_runtime() -> Arc<Runtime> {
    use std::sync::OnceLock;
    static INSTANCE: OnceLock<Arc<Runtime>> = OnceLock::new();
    INSTANCE
        .get_or_init(|| Arc::new(Runtime::new().unwrap()))
        .clone()
}
```

We would then convert all the async tests to use it, eg:

```rust
#[tokio::test]
async fn mytest() {..}

// becomes

#[test]
fn mytest() {
    tokio_runtime().block_on(async { 
        // test logic
    })
}
```

But this is kinda gross because:

1. its boilerplate spread across a lot of tests.
2. all the test code has to be indented.

We could do that if we have to, but there's a nicer way.   We can make our own fn attribute, eg:

```rust
#[neptune::test]
fn mytest() {
    // test logic
}
```
nice right?

Unfortunately this magic requires making a procedural macro, which apparently has to be in a separate crate.

That would be easy to do, if we are using a workspace.

#452 creates a workspace for neptune-core.  However I just tried to rebase it, and... huge conflicts.  Also #550 is a pretty big incoming PR.

So I think a plan can be:

1. wait for #550 to be merged (and any other high priority PRs)
2. remake a workspace from scratch, following pattern from #550.
3. get the workspace merged right away.  we can't let it become stale.
4. make a (simple) crate for the #[neptune::test] macro, and any future procedural macros.
5. decorate all the async tests.

dan-da:
I plan to merge this right away to avoid merge conflict hell, but making it a PR in case anyone wishes to review post-hoc.

----------
Selected `cargo test -- --nocapture` log output, demonstrating shared job-queue usage in tests.

```
$ grep "begin job" /tmp/test10.out | head -n 200 | tail -n 1
2025-04-26T18:19:05.498349Z  INFO neptune_cash::job_queue::queue:   *** JobQueue: begin job #113 - bae809233ab5b9ed78369281 - 45 queued job(s) ***

$ grep "begin job" /tmp/test10.out | head -n 600 | tail -n 1
2025-04-26T18:19:49.754200Z  INFO neptune_cash::job_queue::queue:   *** JobQueue: begin job #513 - c26fba5ad85ab5751def2e97 - 47 queued job(s) ***

$ grep "begin job" /tmp/test10.out | head -n 1000 | tail -n 1
2025-04-26T18:20:56.447659Z  INFO neptune_cash::job_queue::queue:   *** JobQueue: begin job #913 - 1982551c7ae04b67d3d16952 - 21 queued job(s) ***

$ grep "begin job" /tmp/test10.out | head -n 1400 | tail -n 1
2025-04-26T18:22:07.064543Z  INFO neptune_cash::job_queue::queue:   *** JobQueue: begin job #1313 - c29014f9d8a50c8ec5db1ee8 - 2 queued job(s) ***

$ grep "begin job" /tmp/test10.out | tail -n 1
2025-04-26T18:23:26.718204Z  INFO neptune_cash::job_queue::queue:   *** JobQueue: begin job #1571 - 41dc5fc021681f06a4f268d4 - 0 queued job(s) ***
```

jobs now have a random job_id that is assigned when added and can be helpful for tracking in the logs.

```
$ grep "c26fba5ad85ab5751def2e97" /tmp/test10.out
2025-04-26T18:19:41.262403Z  INFO neptune_cash::job_queue::queue: JobQueue: job added - c26fba5ad85ab5751def2e97  49 queued job(s).  job running: #465 - 29ae04dca4c2bdefe12b2d1c
2025-04-26T18:19:49.754200Z  INFO neptune_cash::job_queue::queue:   *** JobQueue: begin job #513 - c26fba5ad85ab5751def2e97 - 47 queued job(s) ***
2025-04-26T18:19:49.786660Z  INFO neptune_cash::job_queue::queue:   *** JobQueue: ended job #513 - c26fba5ad85ab5751def2e97 - Completion: Finished - 0.03245471 secs ***
```

_note the 8 second difference between "job added" and "begin job".  Previously these could not be correlated_

-------------

closes #551

purpose: share a common tokio runtime for unit tests so that they can
likewise share a common TritonVmJobQueue.

This enables running tests in parallel even when generating proofs.

Changes:
1. adds test fn tokio_runtime() that returns a singleton Runtime
   reference.
2. add macro shared_tokio_runtime!() which enables modifying an existing
   async fn to use the shared runtime, instead of #[tokio::test]
3. add dev dependency on macro_rules_attr, to apply the macro
4. apply to async test fn with #[apply(shared_tokio_runtime)]
5. tests now acquire singleton TritonVmJobQueue instead of a new one.

Squashed:
* first test compiles with shared_tokio_test attribute
* 1. all async tests now share tokio runtime.
  2. tests now share single TritonVmJobQueue instance
* use shared tokio runtime for test fn that call block_on() manually
 
-------------

the above is the primary commit.   But in the course of this I also implemented some improvements to the job-queue, in this commit.

refactor: simplify job-queue and log unique job-id

These changes were motivated by a need to follow/debug job-queue
behavior in the logs when shared by all async unit tests.

Primary changes:
1. Each job is now assigned a randomly generated 12-byte job-id that is
   logged when adding a job and when it begins and completes
   processing.
2. Simplifies queue implementation by removing the "add_job" spawned
   task.  There is now just a single "process_job" task, and jobs are
   added to the queue, via mutex, in the caller's task.
3. The process_job task listens for a Stop message and will stop
   right away, after signaling running job to cancel, if any.
4. Adds an async stop() method that also drops the queue. It performs
   cleanup better than a simple drop().
5. improve error handling.

Squashed:

* refactor job-queue so there is only one spawned task
* separate add-job and stop into separate channels
* add job_id to jobs, log
* job-queue cleanups
* wait for current job to complete when stopping queue

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -31,4 +31,3 @@ macro_rules! shared_tokio_runtime {
 }
 
 pub(crate) use shared_tokio_runtime;
-
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -1,1 +1,34 @@
 pub mod shared;
+
+use std::sync::OnceLock;
+
+use tokio::runtime::Runtime;
+
+pub fn tokio_runtime() -> &'static Runtime {
+    static RUNTIME: OnceLock<Runtime> = OnceLock::new();
+    RUNTIME.get_or_init(|| Runtime::new().unwrap())
+}
+
+macro_rules! shared_tokio_runtime {
+    (
+        $(#[$fn_meta:meta])*
+        $vis:vis async fn $fn_name:ident() $(-> $ret:ty)? {
+            $($tt:tt)*
+        }
+    ) => {
+        $(#[$fn_meta])*
+        #[test]
+        fn $fn_name() $(-> $ret)? { // Propagate the return type to the #[test] fn
+            let runtime = $crate::tests::tokio_runtime();
+            runtime.block_on(async {
+                async fn __inner() $(-> $ret)? {
+                    $($tt)*
+                }
+                __inner().await // Return the awaited result
+            })
+        }
+    };
+}
+
+pub(crate) use shared_tokio_runtime;
+
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1703,6 +1703,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "macro_rules_attr"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ac8e59369155fdc751b1f687fec504eef885545d7f131f4a19741612fc31fdc7"
+
+[[package]]
 name = "manyhow"
 version = "0.11.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1859,6 +1865,7 @@ dependencies = [
  "humantime",
  "itertools 0.11.0",
  "leveldb-sys",
+ "macro_rules_attr",
  "memmap2",
  "num-bigint",
  "num-rational",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -142,6 +142,7 @@ assert2 = "0.3"
 blake3 = "1.5.4"
 clienter = "0.1.1"
 divan = "0.1.14"
+macro_rules_attr = "0.1.3"
 pin-project-lite = "0.2.14"
 proptest = { version = "1.5" }
 proptest-arbitrary-interop = { version = "0.1" }
--- a/src/connect_to_peers.rs
+++ b/src/connect_to_peers.rs
@@ -582,9 +582,12 @@ mod connect_tests {
     use crate::tests::shared::to_bytes;
     use crate::MAGIC_STRING_REQUEST;
     use crate::MAGIC_STRING_RESPONSE;
+    use macro_rules_attr::apply;
+    use crate::tests::shared_tokio_runtime;
+
 
     #[traced_test]
-    #[tokio::test]
+    #[apply(shared_tokio_runtime)]
     async fn test_outgoing_connection_succeed() -> Result<()> {
         let network = Network::Alpha;
         let other_handshake = get_dummy_handshake_data_for_genesis(network);
@@ -655,7 +658,7 @@ mod connect_tests {
     }
 
     #[traced_test]
-    #[tokio::test]
+    #[apply(shared_tokio_runtime)]
     async fn test_get_connection_status() -> Result<()> {
         let network = Network::Alpha;
         let (_, _, _, _, mut state_lock, own_handshake) =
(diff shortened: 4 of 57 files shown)


dan-da:
CI mac and ubuntu builds are working.   windows fails, as it already did.

There is a new coveralls failure, but I opened #570 about that.  I also plan to add another commit with a test-case for the new stop() method in JobQueue, which isn't presently used anyway.

proceeding with merge.

src/tests.rs
```diff
@@ -1 +1,33 @@
...
+
+pub fn tokio_runtime() -> &'static Runtime {
+    static RUNTIME: OnceLock<Runtime> = OnceLock::new();
+    RUNTIME.get_or_init(|| Runtime::new().unwrap())
+}
+
+macro_rules! shared_tokio_runtime {
+    (
+        $(#[$fn_meta:meta])*
+        $vis:vis async fn $fn_name:ident() $(-> $ret:ty)? {
```
jan-ferdinand, line 15: `vis` is never used. Is that on purpose? If so, I suggest `$_:vis` over `$vis:vis`.

It's [possible](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=8a27f7cabd63a70dda3e26d0936bdcde) to re-use tests in other test code, but it also seems rather arcane.

jan-ferdinand (commented):
Glad to see you could use `macro_rules_attr` so effectively! Left 2 minor questions inline. 
Haven't looked at the job queue in any detail. The job ids look like a good addition.

src/tests.rs
dan-da: good catch.

on purpose... no, it was initially AI generated from your example and then I tweaked it some more until all the tests compiled without warnings and passed.  I don't find macros enjoyable, and only know enough to be dangerous.

Looking at it more closely, yeah I see that $vis:vis is capturing pub visibility, but then is not transferred to the inner async fn definition.   I will correct that.  kinda interesting it didn't cause any error/warning.

dan-da: also addressed in 2d908eedfb696f6bdb5816b3eba22c654ce12b98

dan-da:
@jan-ferdinand thx for the review!

src/tests.rs
jan-ferdinand: The reason there weren't any errors: I don't know of a scenario where using a `#[test] fn foo()` inside some other `#[cfg(test)]` place makes sense, and it'd certainly be a weird pattern to see. Honestly, it'd even be weird to see a `#[test] pub fn foo()`. It's good you put the `$vis:vis` fragment in the macro, but hopefully even that isn't used anywhere.

The reason there weren't any warnings: clippy probably doesn't inspect macro fragment specifiers, only the code generated by the macro.

PR #571, opened 2025-04-27
test: normalize 'mod tests' naming

### LINKED ISSUE #570: cargo llvm-cov appears to be including (some) tests as if code-under-test
I may well have a misunderstanding, but reporting as best I can deduce at the moment.

Some coveralls tests [fail](https://coveralls.io/builds/73398398) for #569 due to decreased coverage.   Upon checking the report I see decreased coverage up to 10% in files that I didn't modify except to change `#[tokio::test]` to `#[apply(shared_tokio_test)]`.   If I'm reading the coveralls report correctly, it seems to be indicating that certain lines *within* the unit test fn themselves are no longer being covered.   Here's an [example](https://coveralls.io/builds/73398398/source?filename=src%2Fmain_loop.rs#L2585), in main_loop.rs.

I was initially a bit baffled about two things:
1. why code in test module is being treated as code-under-test
2. why the coverage appears to have changed for these.

Of these (1) is by far the more important question and (2) I think can be dismissed if (1) is resolved.

Reading the [cargo-llvm-docs](https://github.com/taiki-e/cargo-llvm-cov?tab=readme-ov-file#exclude-code-from-coverage), I notice that they have an attribute for excluding code from coverage, and say this:

> If you want to ignore all #[test]-related code, you can use module-level #[coverage(off)] attribute:
> 
> #![cfg_attr(coverage_nightly, feature(coverage_attribute))]
> 
> #[cfg(test)]
> #[cfg_attr(coverage_nightly, coverage(off))]
> mod tests {
>     // ...
> }
> _cargo-llvm-cov excludes code contained in the directory named **tests** from the report by default, so you can also use it instead of #[coverage(off)] attribute_.

In main_loop.rs, we have:

```
#[cfg(test)]
mod test {
```
_note that it is named `test`, not `tests`_.

And we have a bunch of other places that use module names other than `tests`. 

So I think that to fix this we need to either:
1. rename these modules to `tests` (probably easiest)
2. use coverage(off) attribute as per docs.

-----------

Given these findings, I'm opening this issue rather than delaying merge of #569, which appears mostly an innocent victim here.

dan-da:
closes #570

renames all top-level test modules to `tests`.

This is done because cargo llvm-cov treats `mod test` as tests, but treats everything else as code under test (without a special attribute).


dan-da:
well, it seems my assumptions about how cargo llvm-cov and coveralls work may be wrong somehow.

coveralls reports that this PR actually decreases code coverage by 0.1%, and further it still appears to me that code inside `mod tests` is being interpreted as code under test in the coveralls report.

for example:
https://coveralls.io/builds/73399472/source?filename=src%2Fpeer_loop.rs#L3594

but I haven't dug into coveralls docs, so likely I'm mis-interpreting the report.

jan-ferdinand (approved):
I'm in favor of merging this for the consistency it adds.

jan-ferdinand:
Integrated into and superseded by #574

PR #572, opened 2025-04-28
ci: Use tarpaulin to generate coverage reports

jan-ferdinand:
I'm not sure how much to trust the new data. For example, the file `archival_mutator_set.rs` is now down to <2% test coverage. Prominent functions like [`prove`](https://coveralls.io/builds/73414155/source?filename=src%2Futil_types%2Fmutator_set%2Farchival_mutator_set.rs#L42) are supposedly uncovered. However, I count 11 call sites from within tests. Maybe the tarpaulin doesn't pick up on code executed by the job queue?

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -16,25 +16,24 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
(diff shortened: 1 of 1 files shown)


jan-ferdinand:
Superseded by #574

PR #573, opened 2025-04-29
docs: document release candidate testnet policy

dan-da:
updates the mdbook with:

1. a (new) description of the versioning scheme.

2. a modified git branching visualization and (new) description to reflect the process of iterating release candidates.

3. a policy for testing each release candidate in the public testnet.

-----------

note: all changes are proposals.  If anything is unclear or there is not concensus, then we should consider a meeting to discuss/resolve, as that should hopefully be faster than a lot of comments back and forth.

Also, it's quite possible that the mdbook document I started from was already out of date with respect to current processes, so if that's the case, please consider merging in current practice and/or documenting in a comment, so I can.

### DIFF
--- a/docs/src/contributing/git-workflow.md
+++ b/docs/src/contributing/git-workflow.md
@@ -11,7 +11,7 @@ It can be visualized like this:
 master     / topic   \
 ----------------------------------------->
     * v0.3.0-rc1    * v0.4.0-rc1
-    * v0.3.0        | v0.4.0
+    * v0.3.0        | v0.4.0-rc2
                     ---------
                             * v0.4.0-rc2
                             * v0.4.0
@@ -20,6 +20,8 @@ master     / topic   \
 ```
 
 _note: "*" represents a tag and "|" represents a branch point._
+
+_not shown: changes from branches v0.4.0-rc2 and v0.4.1 should be merged into master, if applicable._
 
 ### visualization, described
 
@@ -105,9 +107,9 @@ It is project policy that each major and minor release be tested as a
 release-candidate in the public community testnet prior to making an official
 release.
 
-A period of public testnet testing provides the community an opportunity to try out
-candidate builds in their own unique environments and report any issues, and to
-prepare for new features in advance.
+A period of public testnet testing provides the community with an opportunity to
+try out candidate builds in their own unique environments and report any issues,
+and to prepare for new features in advance.
 
 It is also an important period of integration testing "in the wild" for the
 release candidate binary.  Operating on a public testnet, it will necessarily be
@@ -116,21 +118,23 @@ that do not occur with automated testing.
 
 #### Some specifics
 
-1. Each release candidate must be publically announced with a dedicated topic in
+1. The release candidate should be published using the same release process as
+   the eventual release; see [releasing](releasing.md).
+2. Each release candidate must be publicly announced with a dedicated topic in
    [talk.neptune.cash](https://talk.neptune.cash).
-2. The release candidate should target 2 weeks of testing after the announcement
+3. The release candidate should target 2 weeks of testing after the announcement
    on average with a minimum of 1 week.
-3. The company behind neptune-cash will provide at least 1 dedicated machine for
+4. The company behind neptune-cash will provide at least 1 dedicated machine for
    the public testnet for the purpose of running the release-candidate binary
    and composing blocks.  Of course the community is also encouraged to run
    nodes and mine (compose or guess) if possible.
-4. A set of automated tests will be created that utilize the RPC API to perform
+5. A set of automated tests will be created that utilize the RPC API to perform
    automated transactions and test basic functionality of the live running
    nodes. These tests should/must be run at least once against each
    release-candidate binary while connected to the public testnet network.
-5. If the release-candidate introduces breaking consensus changes, then a
-reset/restart of the testnet from the genesis block may be performed.
-6. Patch releases may skip testnet testing if the risk seems low and/or the
+6. If the release-candidate introduces breaking consensus changes, then a
+   reset/restart of the testnet from the genesis block may be performed.
+7. Patch releases may skip testnet testing if the risk seems low and/or the
    urgency is sufficiently high.
 
 
@@ -209,36 +213,8 @@ The normal/typical release flow would be:
 
 The release flow for a major version change is essentially the same except that the major version number is bumped and the minor version returns to 0.
 
+Detailed release instructions are in [releasing.md](releasing.md).
 
-## Release Candidate Protocol  (-rc1, -rc2, etc)
-
- - determine source branch and new version.  (see visualization above.)
- - checkout the source branch and ensure working directory is clean.
- - Set the version in `Cargo.toml` (including the -rcX suffix).
- - Ensure that source branch (eg master) builds against crates that live on
-   [crates.io](https://crates.io). In particular, no dependencies on github
-   repositories or revisions.
- - review `README.md` to ensure the installation instructions are up-to-date.
- - Ensure that all tests pass.
- - Create a commit with the subject line `v0.3.0-rc1` (or whatever the new version number is) and in the body list all the changes.
- - Push present branch to github.
- - Add a tag marking the current commit version:
-   - `git tag v0.3.0-rc1` (or whatever the next version is)
-   - `git push --tags`.
- - build binaries and make a "release" on github
- - start a testnet with the release-candidate executables.
- - announce the release and new testnet on [talk.neptune.cash](https://talk.neptune.cash).
-
-## Release Protocol (major.minor.point)
-  Perform steps as above except:
-   - `git tag v0.3.0` (or whatever the next version is)
-   - `git tag latest-release`
-   - `git push --tags`.
-   - no need to make a testnet.
-   - announce the release on [talk.neptune.cash](https://talk.neptune.cash)
-   - Consider updating social media:
-     - telegram, twitter/x, reddit.com/r/cryptocurrencies,
-       reddit.com/r/neptune-cash, etc.
 
 ## latest-release tag
 
--- a/docs/src/contributing/git-workflow.md
+++ b/docs/src/contributing/git-workflow.md
@@ -7,33 +7,81 @@ We follow a standard [GitHub Flow](https://docs.github.com/en/get-started/using-
 It can be visualized like this:
 
 ```
-            --------
-master     / topic  \
-----*----------------------*--------------->
-    \ release              \ release
-    ------------------>    --------->
-    \ hotfix /
-    --------
+            ---------
+master     / topic   \
+----------------------------------------->
+    * v0.3.0-rc1    * v0.4.0-rc1
+    * v0.3.0        | v0.4.0
+                    ---------
+                            * v0.4.0-rc2
+                            * v0.4.0
+                            | v0.4.1
+                            -----------* v0.4.1
 ```
 
-### master branch (aka trunk)
+_note: "*" represents a tag and "|" represents a branch point._
 
-The `master` branch represents the tip of current development. It is an _integration_ branch, in the sense that developer changes from smaller _topic_ branches get merged and integrated into `master` and github's CI performs testing for every pull-request.
+### visualization, described
 
-The master branch of each crate should always build and should always pass all tests.
+1. The `master` branch represents the tip of current development. It is an
+   _integration_ branch, in the sense that developer changes from smaller
+   _topic_ branches get merged and integrated into `master` and github's CI
+   performs testing for every pull-request.
 
-At present, any team member with repo write access may directly commit to the `master` branch. However, as we get closer to a mainnet launch, `master` should/will become locked so that all changes must go through the pull-request process and be peer reviewed.
+2. **Scenario 1 (v0.3.0 - Clean Release):** A release candidate (`v0.3.0-rc1`)
+   is tagged from the `master` branch. Following testing, no significant issues
+   are found, and the **same commit** is directly tagged as the final release
+   (`v0.3.0`). This represents a smooth and efficient release where the first RC
+   is deemed stable enough for production.
 
-### topic branches
+3. **Scenario 2 (v0.4.0 - Iterative Release Candidate Process):** A release
+   candidate (`v0.4.0-rc1`) is tagged from the `master` branch. During testing,
+   problems are identified. This necessitates further development and the
+   creation of a subsequent release candidate (`v0.4.0-rc2`) on a branch derived
+   from `v0.4.0-rc1`. After the issues are resolved in `v0.4.0-rc2`, the final
+   release (`v0.4.0`) is tagged from that stable point.
 
-Even now, team members are encouraged to create a *topic* branch and pull-request for larger changes or anything that might be considered non-obvious or controversial.
+4. **Patch Release (v0.4.1):** Following the stable release of `v0.4.0`, a bug
+   or security issue is discovered. A patch release (`v0.4.1`) is created based
+   on the `v0.4.0` tag to address this specific problem. This allows for a
+   focused update to the stable release without incorporating new features or
+   major changes that would warrant a minor version bump.  Note that a
+   release-candidate is not employed in this example, as such patches should be
+   small and there may be a high urgency.
+
+5. **Topic Branch (`/ topic \`):** This represents a short-lived branch created
+   from the `master` branch for the development of a specific feature or bug
+   fix. Once the work on the topic is complete and tested, it is eventually
+   merged back into the `master` branch, contributing to future release cycles.
+
+
+### Branch policies:
+
+#### master branch (aka trunk)
+
+The master branch of each crate should always build and should always pass all
+tests.
+
+At present, any team member with repo write access may directly commit to the
+`master` branch. However, now that neptune-core has launched Mainnet this policy
+may soon be revised such that `master` will become locked so that all changes
+must go through the pull-request process and be peer reviewed.
+
+#### topic branches
+
+Team members are encouraged to create a *topic* branch and pull-request for
+larger changes or anything that might be considered non-obvious or
+controversial.
 
 tip: *topic* branches are sometimes called *feature* branches.
 
-A *topic* branch typically branches off of `master` or another *topic* branch.  It is intended for an individual
-feature or bug-fix.  We should strive to keep each *topic* branch focused on a single change/feature and as short-lived as possible.
+A *topic* branch typically branches off of `master` or another *topic* branch.
+It is intended for an individual feature or bug-fix.  We should strive to keep
+each *topic* branch focused on a single change/feature and as short-lived as
+possible.
 
-Third party contributors without repo write access must create a *topic* branch and submit a pull request for each change.  This is accomplished by:
+Third party contributors without repo write access must create a *topic* branch
+and submit a pull request for each change.  This is accomplished by:
 1. fork the repo
 2. checkout and build the desired branch (usually master or a release branch)
 3. create a topic branch
@@ -41,7 +89,7 @@ Third party contributors without repo write access must create a *topic* branch 
 5. push your topic branch to your forked repo
 6. submit the pull request.
 
-#### Topic Branch Naming
+##### Topic Branch Naming
 
 When working on an open github issue, it is recommended to prefix the topic branch with the issue identifier.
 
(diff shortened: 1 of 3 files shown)


docs/src/contributing/git-workflow.md
```diff
@@ -51,36 +99,173 @@ If the branch exists in a triton/neptune official repo, (as opposed to a persona
...
+For blockchains it is of primary importance to maintain consensus and network
+compatibility with other nodes on the network.
+
+For these reasons, we define the version identifier as follows.
+
+## version identifier
+
+neptune-core uses sem-ver compatible identifier of the form:
+```
+<major>.<minor>.<patch>[-<candidate>]
```
jan-ferdinand, line 152: Note that “Cargo uses the convention that only changes in the left-most non-zero component are considered incompatible” ([source](https://doc.rust-lang.org/cargo/reference/semver.html#change-categories)). This means that for releases prior to v1.0.0, the semantic version as interpreted by Cargo are

```
0.0.<major>[-<pre-release>][+<metadata>]
0.<major>.<minor>[-<pre-release>][+<metadata>]
```

Notably, this is different from [semver](semver.org) since there, any two releases prior to v1.0.0 are considered mutually incompatible ([source](https://semver.org/#spec-item-4)).

Do we want to incorporate this in some way?

docs/src/contributing/git-workflow.md
```diff
@@ -51,36 +99,173 @@ If the branch exists in a triton/neptune official repo, (as opposed to a persona
...
+*Note that the optional suffix is supported by cargo and related tools.*
+
+## typical release
+
+The normal/typical release flow would be:
+
+1. normal changes (not consensus) are introduced in master.
+2. a release candidate is tagged from master, eg `v0.3.0-rc1` and built.
+3. a testnet is created for `v0.3.0-rc1`
+4. if significant problems are found then a new branch `v0.3.0-rc2` is created at `v0.3.0-rc1`, or possibly from master.  commit(s) are added. When ready, the final commit in the branch is tagged with `v0.3.0-rc2`, another testnet created, and so on.
```
jan-ferdinand, line 203: nit: The branch name `v0.3.0-rc2` deviates from the pattern suggested in the introductory example. Following that pattern, the branch name would be `v0.3.0`, since that's what's currently in the making. I don't have strong opinions here but would prefer consistency. Ultimately, the exact name probably won't matter much since it's unrealistic that we are working on multiple releases at the same time.

docs/src/contributing/git-workflow.md
```diff
@@ -51,36 +99,173 @@ If the branch exists in a triton/neptune official repo, (as opposed to a persona
...
+5. fixes for the release-candidate should also be applied to master, if applicable.
```
jan-ferdinand, line 204: Ah! This answers one of my questions from above. Would it make sense to add a corresponding arrow to the drawing in the intro?

docs/src/contributing/git-workflow.md
```diff
@@ -51,36 +99,173 @@ If the branch exists in a triton/neptune official repo, (as opposed to a persona
...
+6. After two weeks, or when team decides, the latest release-candidate tag is also tagged with `v0.3.0`, signifying an actual release.
+7. The release is performed.
+
+*note that by the time the release actually occurs, master may have diverged significantly.*
+
+The release flow for a major version change is essentially the same except that the major version number is bumped and the minor version returns to 0.
+
+
+## Release Candidate Protocol  (-rc1, -rc2, etc)
```
jan-ferdinand, line 213: I think some of these steps should be replaced by “follow the release protocol” and a link to said release protocol.

docs/src/contributing/git-workflow.md
```diff
@@ -51,36 +99,173 @@ If the branch exists in a triton/neptune official repo, (as opposed to a persona
...
- - Bump the version in `Cargo.toml`
- - Create a commit with the subject line `v0.0.6` (or watever the new version number is) and in the body list all the changes.
- - Push to `master` on github.
- - Add a tag marking the current commit with the version:
-   - `git tag v0.0.6` (or whatever the next version is)
+ - Create a commit with the subject line `v0.3.0-rc1` (or whatever the new version number is) and in the body list all the changes.
+ - Push present branch to github.
+ - Add a tag marking the current commit version:
+   - `git tag v0.3.0-rc1` (or whatever the next version is)
+   - `git push --tags`.
```
jan-ferdinand, line 227: I don't recommend using this command, since it pushes _all_ tags. See also [this](https://stackoverflow.com/questions/5195859/how-do-you-push-a-tag-to-a-remote-repository-using-git#comment40809114_5195913) stackoverflow comment.

Instead, let's go with `git push origin tag v0.3.0-rc1`.

docs/src/contributing/git-workflow.md
```diff
@@ -51,36 +99,173 @@ If the branch exists in a triton/neptune official repo, (as opposed to a persona
...
+ - build binaries and make a "release" on github
+ - start a testnet with the release-candidate executables.
+ - announce the release and new testnet on [talk.neptune.cash](https://talk.neptune.cash).
+
+## Release Protocol (major.minor.point)
+  Perform steps as above except:
+   - `git tag v0.3.0` (or whatever the next version is)
+   - `git tag latest-release`
    - `git push --tags`.
```
jan-ferdinand, line 236: Also here. I think this should work instead: `git push origin tag v0.3.0 tag latest-release`

docs/src/contributing/git-workflow.md
jan-ferdinand: I suppose I never wrote the question down. It was essentially: (why) are fixes enabling a release not added back to the master branch?

docs/src/contributing/git-workflow.md
```diff
@@ -51,36 +99,173 @@ If the branch exists in a triton/neptune official repo, (as opposed to a persona
...
+release-candidate in the public community testnet prior to making an official
+release.
+
+A period of public testnet testing provides the community an opportunity to try out
+candidate builds in their own unique environments and report any issues, and to
+prepare for new features in advance.
+
+It is also an important period of integration testing "in the wild" for the
+release candidate binary.  Operating on a public testnet, it will necessarily be
+exposed to peers running older versions of the software and may shake out issues
```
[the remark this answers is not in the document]
dan-da: short answer:   yes, it's true, as it would be challenging to enforce otherwise on a [long-lived] public testnet.

nuance: we could quibble about the word "necessarily" because it is *possible* all peers could be running same version, esp if participation is very low.  We could replace it with "most likely" to allow that possibility.

long answer, from AI.  asked about bitcoin, but same dynamics should apply to us:

> Yes, the Bitcoin testnet is **heterogeneous** in the sense that nodes running on it often operate using different versions of Bitcoin Core software.
> 
> Here's why:
> 
> * **Testing of New Versions:** The primary purpose of the testnet is to allow developers and testers to experiment with new features, bug fixes, and consensus rule changes *before* they are deployed on the main Bitcoin network. **This inherently involves running pre-release versions (release candidates) alongside older, more stable versions.**
> * **Voluntary Participation:** The Bitcoin testnet is a public, permissionless network where anyone can run a node. There is no central authority mandating which software version participants must use.
> * **Long Lifespan:** The current testnet (testnet3) has been running for many years and has seen numerous Bitcoin Core releases. Nodes may have been started with older versions and might not always be upgraded immediately.
> * **Different Testing Needs:** Individuals and organizations use the testnet for various purposes. Some might be testing specific new features available only in the latest release candidates, while others might be ensuring compatibility of their applications with older, more widely deployed versions.
> 
> **Consequences of Heterogeneity:**
> 
> * **Diverse Feature Set:** The features and consensus rules supported by the network can vary depending on the mix of software versions running.
> * **Potential for Incompatibilities (During Testing):** When new consensus rules or significant protocol changes are being tested in release candidates, there can be temporary incompatibilities between nodes running the new software and those running older versions. This is part of the testing process to identify and resolve such issues.
> * **Real-World Scenario Simulation:** The heterogeneous nature of the testnet, to some extent, mirrors the diversity of software versions that can exist on the mainnet (though the mainnet typically sees a stronger push for upgrades to stable releases).
> 
> **In summary, the Bitcoin testnet is intentionally heterogeneous as it serves as a dynamic environment for testing different versions of Bitcoin Core and their interoperability.** This diversity is crucial for ensuring the robustness and stability of future Bitcoin mainnet releases.

docs/src/contributing/git-workflow.md
```diff
@@ -51,36 +99,173 @@ If the branch exists in a triton/neptune official repo, (as opposed to a persona
...
+
+For these reasons, we define the version identifier as follows.
+
+## version identifier
+
+neptune-core uses sem-ver compatible identifier of the form:
+```
+<major>.<minor>.<patch>[-<candidate>]
+
+where `candidate` is of the form "rc1", "rc2", etc and starts at 1.
```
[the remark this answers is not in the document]
dan-da: yeah, I thought about it a little and concluded those wouldn't really make sense anymore so didn't mention them.

AI indicates that actual btc and xmr usage agrees:

> Based on my knowledge and the search results, here's how Bitcoin Core and Monero typically handle version suffixes:
> 
> **Bitcoin Core:**
> 
> * **Primarily uses `-rc1`, `-rc2`, etc.** for release candidates. This is clearly stated in their "Software Life Cycle" documentation.
> * While the term "beta" might occasionally be used informally within the developer community (as seen in a GitHub issue discussing `-beta` vs. `-rc`), their official releases and tagging scheme predominantly rely on the `-rc` suffix for pre-releases leading up to a stable version.
> * I haven't found evidence of Bitcoin Core using `-alpha` suffixes for public releases. Alpha releases are typically very early, unstable development versions and are usually kept within the development team.
> 
> **Monero:**
> 
> * Based on the available information, Monero also **primarily uses `-rc` suffixes** (like `-rc1`, `-rc2`, etc.) for their release candidates.
> * I haven't found consistent evidence of Monero using `-alpha` or `-beta` suffixes for their publicly announced pre-releases. Like Bitcoin Core, it's possible these terms might be used internally during earlier development stages, but the public pre-releases that users and testers would interact with are generally tagged as release candidates.
> 
> **In summary:**
> 
> Both Bitcoin Core and Monero mainly stick to the `-rc` (release candidate) suffix for their pre-release versions that are made available for public testing before a stable release. While the terms `-alpha` and `-beta` describe earlier, less stable stages of software development, they are not the common suffixes used by these projects for their publicly facing pre-releases.

docs/src/contributing/git-workflow.md
dan-da: thx.  my feeling is that we really should have changed major version to 1 when we launched mainnet.   Mainnet was a very clear signal to end-users saying that this thing is "ready for production use" (whether true or not) which is also an implied dividing line between 0.x.x and 1.x.x.

Since we didn't, we should still correct it in an upcoming release, probaby at the hard-fork by either going to 1.0.0 or 2.0.0.  Once we do that then any 0.x meanings are irrelevant.  So my feeling is we can ignore that distinction/complication for now.

but if you think differently and would like to suggest wording, I'm open to that.

docs/src/contributing/git-workflow.md
dan-da: good catch.  I think the text is correct here, and it's the diagram that needs correction as follows:

```
            ---------
master     / topic   \
----------------------------------------->
    * v0.3.0-rc1    * v0.4.0-rc1
    * v0.3.0        | v0.4.0-rc2
                    ---------
                            * v0.4.0-rc2
                            * v0.4.0
                            | v0.4.1
                            -----------* v0.4.1
```

this simply changes branch "v0.4.0" to "v0.4.0-rc2".

This makes more sense because one would create branch v0.4.0-rc2 in order to achieve that goal, and then tag it once achieved.

docs/src/contributing/git-workflow.md
dan-da: by that, do you mean a link to docs/src/contributing/releasing.md?

Perhaps we need to ensure that document matches up with this one before merging this PR.  I haven't reviewed it yet, other than a 5 second scan just now.

docs/src/contributing/git-workflow.md
dan-da: good point.  yeah, this was just keeping (copy/paste) what was already written so I didn't put thought into it.  will change.

docs/src/contributing/git-workflow.md
```diff
@@ -51,36 +99,173 @@ If the branch exists in a triton/neptune official repo, (as opposed to a persona
...
 So if working on issue `#232` and adding feature *walk_and_chew_gum* one might name the branch `myuser/232_walk_and_chew_gum_pr`.
 
-### release branch
+### Testnet policy (pre-release)
+
+It is project policy that each major and minor release be tested as a
+release-candidate in the public community testnet prior to making an official
+release.
+
+A period of public testnet testing provides the community an opportunity to try out
```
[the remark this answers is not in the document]
dan-da: addressed in 66b171787d18c6ad4aaf45854923545d15666d28

docs/src/contributing/git-workflow.md
dan-da: note added in 66b171787d18c6ad4aaf45854923545d15666d28

docs/src/contributing/git-workflow.md
dan-da: addressed in 66b171787d18c6ad4aaf45854923545d15666d28

docs/src/contributing/git-workflow.md
dan-da: addressed in 66b171787d18c6ad4aaf45854923545d15666d28

docs/src/contributing/git-workflow.md
dan-da: addressed in 66b171787d18c6ad4aaf45854923545d15666d28.

Ok, I made a first effort at syncing up these documents.  So now they link to eachother for some things rather than duplicating.   I think there is still some overlap and room to re-arrange some more, but I'd rather leave that to the author(s) of releasing.md and/or person(s) actually performing the releases.   That may be you @jan-ferdinand, I'm not certain actually.

Anyway there are some changes in releasing.md now that I'd appreciate if you can have a look at.

docs/src/contributing/git-workflow.md
dan-da: addressed in 66b171787d18c6ad4aaf45854923545d15666d28.

I updated the diagram as per above.

docs/src/contributing/git-workflow.md
```diff
@@ -51,36 +99,173 @@ If the branch exists in a triton/neptune official repo, (as opposed to a persona
...
+candidate builds in their own unique environments and report any issues, and to
+prepare for new features in advance.
+
+It is also an important period of integration testing "in the wild" for the
+release candidate binary.  Operating on a public testnet, it will necessarily be
+exposed to peers running older versions of the software and may shake out issues
+that do not occur with automated testing.
+
+#### Some specifics
```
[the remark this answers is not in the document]
dan-da: addressed in 66b171787d18c6ad4aaf45854923545d15666d28.

I added the suggested point and also added near the top of releases.md:

> ## Automation
> 
> This is presently a largely manual process.  It is expected/encouraged that one
> or more custom deployment tools will be developed so this becomes a push-button
> process instead, or even fully automated when certain event(s) happen.  Such a
> tool will remove tedium as well as the human error factor, which is important
> for performing the task regularly and consistently.

docs/src/contributing/git-workflow.md
```diff
@@ -51,36 +99,173 @@ If the branch exists in a triton/neptune official repo, (as opposed to a persona
...
+3. The company behind neptune-cash will provide at least 1 dedicated machine for
+   the public testnet for the purpose of running the release-candidate binary
+   and composing blocks.  Of course the community is also encouraged to run
+   nodes and mine (compose or guess) if possible.
+4. A set of automated tests will be created that utilize the RPC API to perform
+   automated transactions and test basic functionality of the live running
+   nodes. These tests should/must be run at least once against each
+   release-candidate binary while connected to the public testnet network.
+5. If the release-candidate introduces breaking consensus changes, then a
+reset/restart of the testnet from the genesis block may be performed.
```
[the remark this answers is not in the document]
dan-da: addressed in 66b171787d18c6ad4aaf45854923545d15666d28.

dan-da:
@jan-ferdinand thx for the review.   I've addressed all of your suggestions.   I made some changes to releasing.md that I'd appreciate if you can take a look at.  My thought is that this PR could be merged now, and any further changes made in follow-up commits.

docs/src/contributing/releasing.md
```diff
@@ -2,6 +2,23 @@
 
 This section describes the steps to publish a new version of Neptune Core, and to release & distribute its binary artifacts.
 
+## Automation
+
+This is presently a largely manual process.  It is expected/encouraged that one
+or more custom deployment tools will be developed so this becomes a push-button
+process instead, or even fully automated when certain event(s) happen.  Such a
+tool will remove tedium as well as the human error factor, which is important
+for performing the task regularly and consistently.
```
jan-ferdinand, line 11: I don't really agree with this paragraph. At present, releasing is a semi-automated task, and uses a bunch of valuable tools. For example, `dist` does a bunch of heavy lifting: it produces the binaries for various target platforms, provides installers for them, creates a github release, among other things. `cargo release` publishes all inter-dependent workspace crates in the right order[^rel] and can be configured to do the `git tag`ging in the way we want it to.[^tag]

There are also steps that will stay manual for the foreseeable future. For example, creating a changelog is curation work,[^dump] which is difficult to automate.[^llm] I'm happy to have additional steps automated, but I don't believe there's a 1-click solution waiting around the corner.

By and large, I'm happy to have a summary of the document at the top, but I think at present, it undersells the amount of tedium the tools in use remove already. How about moving the summary at the beginning of the “## Release Process Checklist” section here, then writing a new & short sentence for that section?

[^dump]: The alternative of simply dumping the output of `git cliff` into the changelog makes the changelog unreadable and thus superfluous.
[^llm]: Unless we're willing to incorporate an LLM. Which… maybe?
[^tag]: Granted, this configuration is not present at the moment.
[^rel]: This is not really relevant right now, but once we have multiple workspace members, it's a _huge_ help.

docs/src/contributing/releasing.md
```diff
@@ -2,6 +2,23 @@
...
+
+## Release Candidate vs Release.
+
+Review the distinction [here](git-workflow.md), and make sure you know which is being generating and what the correct version should be.
```
jan-ferdinand, line 15:
```suggestion
Review the distinction [here](git-workflow.md), and make sure you know which is being generated and what the correct version should be.
```

docs/src/contributing/git-workflow.md
jan-ferdinand: I agree with the “should have been 1.0.0” sentiment. I personally have a weird irrational apprehension to doing a 1.0.0 release. Let's try to get a 1.0.0 going. I agree that the distinction between Cargo's semver and semver.org are irrelevant then, and that a simpler document is favorable.

docs/src/contributing/git-workflow.md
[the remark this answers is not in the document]
jan-ferdinand: Cool, thanks for the research. I didn't mean to start quibbling about the exact wording; the question was actually inquisitive. 🙂

docs/src/contributing/releasing.md
```diff
@@ -166,15 +186,37 @@ Edit them until you are happy, then push the tag(s) to GitHub.
...
+### If a release-candidate:
 
-By convention, branch `release` should always point to the latest stable commit compatible with the latest release.
+#### Deploy release to testnet machine(s).
+
+(full instructions: TBD)
+
+Announce the release-candidate in a post at talk.neptune.cash.
+
+Ensure that a composer is running using the release-candidate binary.
```
jan-ferdinand, line 197: Maybe flip order, but since this needs an overhaul, it's a bit 🤷.

docs/src/contributing/releasing.md
```diff
@@ -166,15 +186,37 @@ Edit them until you are happy, then push the tag(s) to GitHub.
...
+
+### If an actual release:
+
+#### Set tag `latest-release`
```
jan-ferdinand, line 201: With this PR, what used to be the `release` _branch_ is becoming the `latest-release` _tag_. I think a tag makes more sense than a branch. However, the corresponding references to the branch in the README.md should also be updated.

docs/src/contributing/git-workflow.md
```diff
@@ -51,36 +101,147 @@ If the branch exists in a triton/neptune official repo, (as opposed to a persona
...
+3. The release candidate should target 2 weeks of testing after the announcement
+   on average with a minimum of 1 week.
+4. The company behind neptune-cash will provide at least 1 dedicated machine for
+   the public testnet for the purpose of running the release-candidate binary
+   and composing blocks.  Of course the community is also encouraged to run
+   nodes and mine (compose or guess) if possible.
+5. A set of automated tests will be created that utilize the RPC API to perform
+   automated transactions and test basic functionality of the live running
+   nodes. These tests should/must be run at least once against each
+   release-candidate binary while connected to the public testnet network.
```
aszepieniec, line 134: A different set of tests for every release candidate, or are all release candidates subject to the same set of tests (that can be updated over time to reflect API evolution)?

Let's add this rule to the test policy after the API test suite exists. If we add it before, we risk over-committing and risk having to delay urgent updates, even if they are orthogonal to the things being tested by that test suite.

docs/src/contributing/git-workflow.md
```diff
@@ -51,36 +101,147 @@ If the branch exists in a triton/neptune official repo, (as opposed to a persona
...
+
+resets minor and patch to 0.
+
+**minor**: bumped any time changes from master are included in a release candidate.
+
+resets patch to 0.
+
+**patch**: bumped any time changes from a release branch are included in a release.
+
+resets candidate to "-rc1".
```
aszepieniec, line 194: Based on the picture in the github workflow, a patch release resets the suffix to "" (empty string). Specifically, the version evolves from 0.4.0 to 0.4.1 without suffix. The suffix is only added when a new release candidate is made.

docs/src/contributing/git-workflow.md
```diff
@@ -51,36 +101,147 @@ If the branch exists in a triton/neptune official repo, (as opposed to a persona
...
+
+Detailed release instructions are in [releasing.md](releasing.md).
+
+
+## latest-release tag
 
-The `master` branch can contain changes that are not compatible with whatever network is currently live. Beta-testers looking for the branch that will synchronize with the network that is currently live need branch `release`. This branch may cherry-pick commits that are meant for `master` so long as they are backwards-compatible. However, when this task is too cumbersome, branch `release` will become effectively abandoned -- until the next network version is released.
+The `master` branch may contain changes that are not compatible with previous
+release. Individuals looking for the latest release can simply checkout the
+`latest-release` tag, which is updated as part of the release process.
```
aszepieniec, line 223: What's wrong with the current practice of using *branch* "release" for this purpose?

aszepieniec (commented):
Thanks for the writeup.

I left some comments inline.

In any case, this is something we want @Sword-Smith to be on board with so there is no rush as it will take a while until he is back in the office.

docs/src/contributing/releasing.md
dan-da: We don't have to include the paragraph.  It seems you want it removed, so I will do so.

However, I maintain that the process is inherently a manual process right now or else releases.md could simply be:  "step 1.  run build_and_deploy.sh.    step 2. go do some other work, because you are done".   Or even better "nightly builds are automatically created every day.  release candidates and releases are automatically produced and deployed, when version tags are detected in git".

Instead, it is a pretty lengthy document with a lot of manual steps.

Yes, those manual steps may involve running some complex tools.   What I am suggesting is that custom glue scripts can be created that bind everything together into a single automation.

Now: I am not involved with this release process and am happy to keep it that way.  Those who do the work should set things up how best suits them.   I can say though that I have been involved with release processes in the past, for larger projects than this, and I can tell you that we setup up full automation to have nightly releases, release candidates, and so on.  So if I *were* doing it (which I'm not volunteering for) that's the direction I would pursue.  More work up front, so that there is less work on an ongoing basis.

If this projects succeeds and endures, I have no doubt that will eventually happen.  In the meantime, it's fine to go with whatever is most expedient.

docs/src/contributing/releasing.md
dan-da: yes thx.  and we should probably get buy-off from @aszepieniec regarding the change before merging.  Though to me it is really the only sensible thing... that `latest-release` always points to commit for the latest release.  so hopefully there's consensus on that.

docs/src/contributing/git-workflow.md
```diff
@@ -51,36 +101,147 @@ If the branch exists in a triton/neptune official repo, (as opposed to a persona
...
+that do not occur with automated testing.
+
+#### Some specifics
+
+1. The release candidate should be published using the same release process as
+   the eventual release; see [releasing](releasing.md).
+2. Each release candidate must be publicly announced with a dedicated topic in
+   [talk.neptune.cash](https://talk.neptune.cash).
+3. The release candidate should target 2 weeks of testing after the announcement
+   on average with a minimum of 1 week.
```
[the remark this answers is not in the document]
dan-da: yes that's better, thx.

docs/src/contributing/git-workflow.md
```diff
@@ -51,36 +101,147 @@ If the branch exists in a triton/neptune official repo, (as opposed to a persona
...
+4. The company behind neptune-cash will provide at least 1 dedicated machine for
+   the public testnet for the purpose of running the release-candidate binary
+   and composing blocks.  Of course the community is also encouraged to run
+   nodes and mine (compose or guess) if possible.
```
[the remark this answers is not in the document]
dan-da: In that case I would suggest that the testnet use mock-proofs for now to ensure the blockchain progresses, because otherwise it seems most likely the testnet will simply be dead because no one is donating hardware resources to 24/7 composing. 

One thing I haven't even mentioned yet is that bitcoin's testnet uses a perpetually low difficulty so it is easy for most anyone to mine blocks.  I think even CPU mining still works on the public testnetv3 which has been around since 2014 or so.  So we could do the same, in combination with mock proofs.

Now, mock proofs do not test the system to the same extent that real proofs do, so at such time as an entity steps forward to provide dedicated hardware we might consider switching to a "real proofs" testnet.

In fact, we could facilitate this future switch now in code by creating a new testnet variant, eg testnet-mock, which would set use_mock_proof = true.   So for now we use the `testnet-mock` network for release-candidate testing, and later we switch to `testnet` for testing with real proofs when that is viable.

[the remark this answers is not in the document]
dan-da: today I learned.  here's a brief AI summary of bitcoin's 20 minute rule for testnet difficulty:

> The "20-Minute Exception" Rule: This is the primary reason Testnet difficulty often stays low. If a new block is not found within 20 minutes of the previous block, the difficulty for the very next block is automatically reset to the minimum possible difficulty (1), regardless of the previous difficulty level. This allows even very low-hashrate miners (like CPUs) to eventually find a block and keep the chain moving.

and apparently there is testnetv4 now as of Dec 2024.

Articles:
https://www.bitgo.com/resources/blog/transition-to-bitcoin-testnet4/
https://blog.lopp.net/griefing-bitcoin-testnet/    (problems with ancient testnetv3)

docs/src/contributing/git-workflow.md
dan-da: > A different set of tests for every release candidate, or are all release candidates subject to the same set of tests (that can be updated over time to reflect API evolution)?

the latter.  It's basically an integration test suite, but running against node(s) on a live (test) network. This is the sort of thing that perhaps we could get a google-summer-of-code volunteer/intern to build, with proper guidance.

> Let's add this rule to the test policy after the API test suite exists. If we add it before, we risk over-committing and risk having to delay urgent updates, even if they are orthogonal to the things being tested by that test suite.

I'd suggest keeping it, but mark it as a goal/aspirational rather than policy, rule or promise.

docs/src/contributing/git-workflow.md
dan-da: thx, I'll revisit this.

docs/src/contributing/git-workflow.md
dan-da: 1. A branch that is force-pushed (needed to relocate it) requires anyone that has it checked out to rebase.  Essentially it breaks `git pull`.
2. It is nebulous/confusing.  What does a "release" branch mean?   releases are performed at specific commits.  tags are the mechanism for labeling specific commits.
3. if there are no problems with a particular release that require further commits, then a tag is all that's needed, and a branch need not be created, as per the diagram.
4. I believe it's a fairly non-standard practice.  certainly not something I've seen done before.
5. It feels "wrong" to me, based on decades of experience with version control systems and release management.

In case you don't find my reasons convincing, I also asked gemini AI, who appears to concur.

question:

> Is it a common or recommended git practice to create a branch named "release" that is force pushed to a new location with each official release?

response:

> While it's *possible* to create a "release" branch and force push to it with each official release, it is **generally NOT a common or recommended Git practice**, especially in collaborative environments. Here's why:
> 
> **Problems with Force-Pushing to a Shared "release" Branch:**
> 
> * **Rewriting History:** Force-pushing rewrites the commit history of the remote branch. This can cause significant problems for other collaborators who have already based their work on the previous state of the "release" branch. They would need to rebase or reset their local branches, which can be confusing and error-prone, potentially leading to lost work.
> * **Disrupting Collaboration:** A constantly changing history on a shared branch makes it difficult for team members to track changes, understand the evolution of the release, and collaborate effectively.
> * **Complicating Auditing and Traceability:** Rewriting history makes it harder to audit past releases, track down the origin of specific changes, and understand the lineage of the codebase.
> * **Going Against Git Principles:** Git is designed to track history. Force-pushing undermines this fundamental principle.
> 
> **More Common and Recommended Practices for Handling Releases in Git:**
> 
> 1.  **Release Branches with Merge:**
>     * Create a dedicated release branch (e.g., `release/1.0`, `release/v2.3.0`) from a stable development branch (like `develop` in Gitflow or `main`).
>     * Perform final bug fixes and preparations on this release branch.
>     * **Merge** the release branch into the main branch (`main` or `master`) to mark the official release.
>     * Tag the release commit on the main branch with the release version (e.g., `v1.0`, `2.3.0`).
>     * Optionally, merge the release branch back into the development branch to ensure any last-minute fixes are included in ongoing development.
> 
> 2.  **Tagging Releases on the Main Branch:**
>     * For simpler workflows (like GitHub Flow), you might directly tag the specific commit on the `main` branch that represents the official release.
> 
> 3.  **Immutable Release Branches (Less Common with Force-Push):**
>     * Some teams might create a release branch and treat it as immutable once created. They would never force-push to it after its initial creation. Any necessary changes would involve creating new commits and potentially merging.
> 
> **Why a Constantly Force-Pushed "release" Branch Might Seem Appealing (and why it's problematic):**
> 
> * **"Clean" History:** The idea might be to have a very linear and "clean" history on the "release" branch, showing only the final state of each release. However, this comes at the cost of losing the history of how that release was prepared.
> * **Simplified View:** It might seem simpler to just point the "release" branch to the latest release commit. However, branches are meant to be pointers that move forward, not backward or sideways by force.
> 
> **In summary, while technically feasible, force-pushing to a shared "release" branch is strongly discouraged due to the negative impact on collaboration, history tracking, and the fundamental principles of Git.** The standard practice involves creating release branches, merging them into the main branch, and tagging the release commits. This preserves history and facilitates collaboration.

aszepieniec: I guess the kernel of the argument is that *branches* are for tracking history which is good for collaboration, whereas *tags* are just pointers to commits without context. For releases, you generally don't want to do much collaboration anyway, and the *branch*-machinery that works well for history management is actually working *against* you when just want to get to the current release snapshot.

I agree. Let's use a tag instead.

docs/src/contributing/releasing.md
jan-ferdinand: Those are fair points. I think everything except generating a meaningful changelog can be fully automated.[^night] I also don't want to imply that the release process is final. If someone wants to write and maintain more tools, I won't stand in their way.
Overall, I don't think the sentiment of the paragraph is wrong, but since it is talking about potential features and potential future development plans, it feels like it belongs in the issue tracker more than in the document.

[^night]: For nightly builds, the changelog is irrelevant; do we want or need nightly builds?

Sword-Smith:
Why don't we focus on testing the release candidate on main net instead? That's where the action is. We'd have to create our own action on testnet. On main net, you'd quickly discover if your new version is incompatible with older versions. On testnet, that's less likely to happen. Also: There are sometimes reorganizations on main net, but they never happen on test net because it's not a competitive environment. Reorganizations test important gotchas of blockchain node logic.

dan-da:
> Why don't we focus on testing the release candidate on main net instead?

The keyword here may be "focus".

Ideally, we would be testing each release-candidate in every way possible, right?

Mainnet has these challenges for testing:
1. requires using real funds.  (for devs and community members)
2. composing and upgrading requires serious hardware, and it was previously stated that such is not presently available for release-candidate testing.
3. do we really want to be spamming mainnet with a bunch of test transactions?
4. incoming blocks and Tx are unpredictable/random.  (also true of testnet, testnet-mock).
5. not a "safe" environment for community members to test out changes before release.

I think a sort of ideal scenario for testing release-candidates would be:

1. run a regtest-specific test harness against it.  This is a fully (or mostly) deterministic environment where the only blocks and tx created are the ones we produce.  This test harness could eg test re-org scenarios that are not possible to reliably test on a public network.

2. run a testnet-mock specific test harness against it.   This can perform composing and upgrading cheaply, eg on a VPS.  devs and community members can also perform test transactions at no real cost, run their own test scripts, etc.  In contrast to (1) the node has to deal with unpredictable peer messages, transactions, blocks so issues may be exposed that do not occur in a deterministic environment.

3. run a mainnet specific test harness against it.   This might not perform any composing or upgrading.  It might or might not initiate transactions since that requires real funds and spams the network.  again, devs and community can run their own tests as well.

If asked to choose an order of implementation (focus) with regards to test-harnesses, I would order them as 2, 1, 3.

Sword-Smith:
> If asked to choose an order of implementation (focus) with regards to building test-harnesses, I would order them as 2, 1, 3.

Our disagreements are quite small, I think. I would order them as 2, 3, 1.

See #606 for an alternative PR.

Issue #568, opened 2025-04-25
Node Crashes Upon Startup


aszepieniec:
I recently restarted a node that had been inactive for a while. Prior to restarting it I pulled and compiled from today's master (cf1fc851365297d3afb7e694965a8c074cea1348). It was syncing for a while but syncing was not complete when I decided I wanted to change the CLI arguments. Upon starting the *second* time, and every time after that I get this log output (redacted for privacy):

```
2025-04-25T16:54:01.09975179Z  INFO ThreadId(01) neptune_cash: Starting neptune-core node on main.
2025-04-25T16:54:01.099918161Z  INFO ThreadId(01) neptune_cash: Data directory is [...]/neptune/main
2025-04-25T16:54:01.099944641Z  INFO ThreadId(01) neptune_cash::models::state::wallet::wallet_file: ***** Reading wallet from [...]/neptune/main/wallet/wallet.dat *****

2025-04-25T16:54:01.09998399Z  INFO ThreadId(01) neptune_cash: Now getting wallet state. This may take a while if the database needs pruning.
2025-04-25T16:54:01.100005142Z  INFO ThreadId(01) neptune_cash::models::state::wallet::wallet_state: wallet DB directory path is [...]/neptune/main/databases/wallet. Exists: yes
2025-04-25T16:54:01.106828729Z  INFO ThreadId(01) neptune_cash::models::state::wallet::rusty_wallet_database: Wallet DB schema version 1 is correct.  proceeding
2025-04-25T16:54:01.118612126Z  INFO ThreadId(01) neptune_cash: Got wallet state.
2025-04-25T16:54:01.123421139Z  INFO ThreadId(01) neptune_cash: Got block index database
2025-04-25T16:54:01.127794035Z  INFO ThreadId(01) neptune_cash: Got peer database
2025-04-25T16:54:01.133748933Z  INFO ThreadId(01) neptune_cash: Got archival mutator set
2025-04-25T16:54:01.138134658Z  INFO ThreadId(01) neptune_cash: Got archival block MMR
2025-04-25T16:54:01.149875081Z  INFO ThreadId(01) neptune_cash: Now listening for incoming peer-connections
2025-04-25T16:54:01.150147441Z  INFO ThreadId(01) neptune_cash: Checking if we need to restore UTXOs
2025-04-25T16:54:01.150296625Z  INFO ThreadId(01) neptune_cash::models::state: Checking [...] incoming UTXOs
[...]

PR #580, opened 2025-05-04
Fix MSMP Resync From Blocks

aszepieniec:
Cf. #568.

Function `resync_membership_proofs_from_stored_blocks` contains a logical error (now fixed) by failing to update the `RemovalRecords` in the course of updating a `MsMembershipProof` with a `MutatorSetUpdate`.

Test `resync_ms_membership_proofs_across_stale_fork` missed the opportunity to catch this logic error. It triggered the above function but only with sequences of blocks devoid of `RemovalRecord`s. Now the sequences of blocks have a nontrivial number of `RemovalRecord`s and `AdditionRecord`s. Without the above above fix in `resync_membership_proofs_from_stored_blocks`, this updated test reliably triggers some logic error related to mutator sets in that function.

Also: refactor mock block generators to facilitate the changes to the test.

### DIFF
--- a/src/models/state/mod.rs
+++ b/src/models/state/mod.rs
@@ -1087,6 +1087,24 @@ impl GlobalState {
         Ok(())
     }
 
+    /// Fix mutator set membership proofs that are unsynced.
+    ///
+    /// This method fixes membership proofs that are synced to an old block,
+    /// possibly even a block that lives on an abandoned chain. It does not work
+    /// for corrupted membership proofs. It assumes that the node stores all
+    /// historical blocks; it does not assume that the node stores an archival
+    /// mutator set.
+    ///
+    /// The algorithm works as follows. For each unsynced monitored UTXO, start
+    /// by finding a path connecting the block it is synced to, to the current
+    /// tip. This path may involve some blocks to revert ("backwards"),
+    /// certainly involves a latest universal common ancestor ("LUCA"), and
+    /// probably involves some blocks to apply ("forwards"). As we walk this
+    /// this path, the mutator set membership proof of the monitored UTXO is
+    /// modified in accordance with the mutator set update induced by the block
+    /// in question, which could be a revert (backwards) or a regular apply
+    /// (forwards).
+    ///
     ///  Locking:
     ///   * acquires `monitored_utxos_lock` for write
     pub async fn resync_membership_proofs_from_stored_blocks(
@@ -1224,6 +1242,7 @@ impl GlobalState {
 
                 // apply additions
                 for addition_record in &additions {
+                    // keep removal records up-to-date
                     RemovalRecord::batch_update_from_addition(
                         &mut removals.iter_mut().collect_vec(),
                         &block_msa,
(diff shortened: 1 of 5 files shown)


Sword-Smith:
The logic seems correct, but this solution is really show as you can see from the provided screenshot. This can be solved by reading the membership proofs from the archival mutator set, if such a state is available.

image

dan-da:
> really show

took me a second to realize you mean really slow.

src/models/state/mod.rs
```diff
@@ -1085,6 +1087,24 @@ impl GlobalState {
...
+    /// for corrupted membership proofs. It assumes that the node stores all
+    /// historical blocks; it does not assume that the node stores an archival
+    /// mutator set.
+    ///
+    /// The algorithm works as follows. For each unsynced monitored UTXO, start
+    /// by finding a path connecting the block it is synced to, to the current
+    /// tip. This path may involve some blocks to revert ("backwards"),
+    /// certainly involves a latest universal common ancestor ("LUCA"), and
+    /// probably involves some blocks to apply ("forwards"). As we walk this
+    /// this path, the mutator set membership proof of the monitored UTXO is
```
dan-da, line 1103: typo: "as we walk this this path"

src/models/state/mod.rs
```diff
@@ -1233,11 +1259,19 @@ impl GlobalState {
                 }
 
                 // apply removals
-                for removal_record in &removals {
-                    membership_proof.update_from_remove(removal_record);
-                    block_msa.remove(removal_record);
+                let mut applied_removals = removals.clone();
```
dan-da, line 1262: this clone is unnecessary because removals is not used after this.   So it can be:
```
let mut applied_removals = removals;
```

or just use `removals` in the loop.

jan-ferdinand:
Book-keeping only: If I'm not mistaken, this PR addresses #568.

aszepieniec:
> The logic seems correct, but this solution is really slow as you can see from the provided screenshot. This can probably be solved by reading the membership proofs from the archival mutator set, if such a state is available.
> 
> image

If you store an archival mutator set, then you get the up-to-date mutator set membership proofs in one go without having to walk through the entire sequence of blocks. The downside of that approach is that it assumes that an archival mutator set is present, which is certainly up for debate in the context of succinctness. In particular, having an archival mutator set should not be a requirement to run a node trustlessly.

Nor should it be a requirement to store all historical blocks, for that matter. But the difference is:
 a) The (succinct) protocol gives a stronger guarantee that the blocks required for a MSMP resync are served, than that MSMPs themselves are served. This is the result of the need to store blocks in order to tolerate reorganizations. So every peer that stores an archival mutator set will also store historical blocks but not the other way around.
 b) Querying MSMPs from peers leaks privacy.

The present algorithm is more robust (because weaker assumptions) and more privacy-friendly (interactively), so let's not *delete* it in favor of the faster one that you propose. But if you want to implement it such that the faster algorithm is used when the preconditions are met (and the slower algorithm otherwise) -- please be my guest.

Issue #579, opened 2025-05-01
"job cancelled by caller" error while composing causes panic or shutdown.


dan-da:
This issue is consolidating a few similar reports that mention "job cancelled by caller" and/or "Must be able to merge transactions in mining context".

issues: #489 #545 #575

I *believe* I now have an understanding why this is occurring.  Anyway, this is my working hypothesis.

------------

In prove_concensus_program(), we have:

```rust
    let job_result = match proof_job_options.cancel_job_rx {
        // fix for issue #348.
        // if we have a cancellation channel from caller then we select on
        // both the channel and job.  If we get a cancel request from the
        // caller, or the channel closes, then we cancel the job
        // which removes it from the job-queue.
        Some(mut cancel_job_rx) => {
            tokio::select! {
                // case: sender cancelled, or sender dropped.
                _ = cancel_job_rx.changed() => {
                    debug!("forwarding job cancellation request to job");
                    cancel_tx.send(())?;
                    anyhow::bail!("job cancelled by caller");
                }
                // case: job completion.
                result = job_handle.result() => result,
            }
        }
        None => job_handle.result().await,
    };
```

So the "job cancelled by caller" comes from here, but only if:
1. TritonVmProofJobOptions::cancel_job_rx is not None *and*
2. we receive a message over the cancellation channel *or*
3. the cancellation channel sender gets dropped.

In the case of composing, it appears that conditions (1) and (3) are met.

in compose_block():
```
job_options.cancel_job_rx = Some(cancel_compose_rx);
```

in mine():
```rust
select! {
    _ = &mut guess_restart_timer => {..}
    Ok(Err(e)) = &mut composer_task => {
        // handle composer_task errors
        // presently this stops composing and/or shuts down, depending on the error
    },
    Some(main_message) = from_main.recv() => {
        match main_message {
            MainToMiner::NewBlock => {..}
            MainToMiner::NewBlockProposal => {..}
            // other messages
        }
    }
    new_composition = composer_rx => {..}
    new_block = guesser_rx => {..}
}
```

So if any of those events occur, the tokio task represented by composer_task will be aborted. Since it holds the channel sender for the proving job, the sender gets dropped, satisfying condition (3).  In particular NewBlock and NewBlockProposal are pretty much guaranteed to occur while composing.

### So, how to fix?

The fix appears pretty simple. 

[...]

PR #584, opened 2025-05-11
feat: improve job_queue ergonomics

dan-da:
note: this PR does not change neptune-core logic.  Only JobQueue API.  I am intending to merge it immediately once CI is happy.  I am making it a PR for a) CI checks, b) to document, and c) in case anyone wishes to review/comment post-hoc.

-------------------

This is a follow-up to #579 that implements some ergonomic improvements to the JobQueue especially with regards to JobResult conversions.

A key features of the JobQueue is that a single JobQueue can accept Jobs of arbitrary (dyn) type and those jobs can return results of arbitrary (dyn) type.  This flexibility comes at a cost:  The type returned by each Job must impl the JobResult trait.  And the caller awaiting the job must downcast the result into the expected type, which is verbose/ugly and kind of uncommon.

The key innovation in this PR is to introduce a new type JobResultWrapper<T> that implements JobResult and provides TryFrom conversions for JobCompletion and `Box<dyn JobResult>`.  This enables:
1. The Job can return results without manually implementing JobResult
2. The Job initiator can easily/cleanly obtain results from the JobCompletion as the expected type T.

Another nice cleanup: I figured out how to make JobHandleError impl Sync.  So we get rid of the ugly JobHandleErrorSync type and all the `map_err(|e| e.into_sync())` when obtaining job results.

Finally, this adds a "hello world" doctest example as well as a more detailed example that implements both a blocking job and an async job that each finds prime numbers and runs them in the same JobQueue.

A few other ergonomic niceties are included, listed below.

Changes:

+ add JobResultWrapper to simplify working with job-specific result types so Job's don't need to manually impl JobResult and callers avoid downcasting results.
+ add simple hello_world JobQueue usage example at job_queue module root.
+ add detailed JobQueue usage example at job_queue module root.
+ make JobHandleError Sync
+ remove JobHandleErrorSync
+ add JobHandleError::is_panic(), into_panic(), try_into_panic() panic_message(), each with doctest example.
+ remove Debug bound from JobResult trait
+ custom impl Debug for JobCompletion
+ add JobQueue::add_job_mut() for explicit mutability usage.
+ add_job and add_job_mut() accept `Into<Box<dyn Job>>` so usage can be add_job(job) instead of add_job(Box::new(job).
+ new tests and new/improved doc-comments

### DIFF
--- a/src/job_queue/errors.rs
+++ b/src/job_queue/errors.rs
@@ -1,73 +1,195 @@
+use std::sync::Arc;
+use std::sync::Mutex;
+
 /// a job-handle error.
-///
-/// This error is not `Sync` due to the `JobPanicked` variant
-/// which holds panic information as returned from tokio::spawn().
-///
-/// If a Sync type is needed, use the `into_sync()` method.
 #[derive(Debug, thiserror::Error)]
 pub enum JobHandleError {
     #[error("the job was cancelled")]
     JobCancelled,
 
+    // note: the Arc<Mutex<_>> is used to make the inner Box `Sync`.
+    // The mutex is never acquired and can thus never be poisoned.
     #[error("the job panicked during processing")]
-    JobPanicked(Box<dyn std::any::Any + Send + 'static>),
+    JobPanicked(Arc<Mutex<Box<dyn std::any::Any + Send + 'static>>>),
 
     #[error("channel send error cancelling job")]
     CancelJobError(#[from] tokio::sync::watch::error::SendError<()>),
 
     #[error("channel recv error waiting for job results: {0}")]
     JobResultError(#[from] tokio::sync::oneshot::error::RecvError),
+
+    #[error("downcast failed converting '{from}' to '{to}'")]
+    JobResultWrapperError {
+        from: &'static str,
+        to: &'static str,
+    },
 }
 
-impl JobHandleError {
-    /// convert into a type that implements `Sync`
-    ///
-    /// note that the JobPanicked variant becomes a `String`.
-    pub fn into_sync(self) -> JobHandleErrorSync {
-        self.into()
+impl From<Box<dyn std::any::Any + Send + 'static>> for JobHandleError {
+    fn from(panic_info: Box<dyn std::any::Any + Send + 'static>) -> Self {
+        Self::JobPanicked(Arc::new(Mutex::new(panic_info)))
     }
 }
 
-#[derive(Debug, thiserror::Error)]
-pub enum JobHandleErrorSync {
-    #[error("the job was cancelled")]
-    JobCancelled,
+impl JobHandleError {
+    /// Returns true if the error was caused by the task panicking.
+    ///
+    /// ```
+    /// use neptune_cash::job_queue::JobQueue;
+    /// use neptune_cash::job_queue::traits::*;
+    ///
+    /// struct PanicJob;
+    ///
+    /// #[async_trait::async_trait]
+    /// impl Job for PanicJob {
+    ///     fn is_async(&self) -> bool {
+    ///         true
+    ///     }
+    ///     async fn run_async(&self) -> Box<dyn JobResult> {
+    ///        panic!("{}", "boom");
+    ///     }
+    /// }
+    ///
+    /// #[tokio::main]
+    /// async fn main() -> anyhow::Result<()> {
+    ///     let job_queue = JobQueue::start();
+    ///     let job_handle = job_queue.add_job(PanicJob, 1)?;
+    ///
+    ///     let err = job_handle.await?.unwrap_err();
+    ///     assert!(err.is_panic());
+    ///
+    ///     Ok(())
+    /// }
+    /// ```
+    pub fn is_panic(&self) -> bool {
+        matches!(self, Self::JobPanicked(_))
+    }
 
-    #[error("the job panicked: {0}")]
-    JobPanicked(String),
+    /// into_panic() panics if the Error does not represent the underlying task terminating with a panic. Use is_panic to check the error reason or try_into_panic for a variant that does not panic.
+    ///
+    /// ```should_panic(expected = "boom")
+    /// use neptune_cash::job_queue::JobQueue;
+    /// use neptune_cash::job_queue::traits::*;
+    ///
+    /// struct PanicJob;
+    ///
+    /// #[async_trait::async_trait]
+    /// impl Job for PanicJob {
+    ///     fn is_async(&self) -> bool {
+    ///         true
+    ///     }
+    ///     async fn run_async(&self) -> Box<dyn JobResult> {
+    ///        panic!("{}", "boom");
+    ///     }
+    /// }
+    ///
+    /// #[tokio::main]
+    /// async fn main() -> anyhow::Result<()> {
+    ///     let job_queue = JobQueue::start();
+    ///     let job_handle = job_queue.add_job(PanicJob, 1)?;
+    ///
+    ///     let err = job_handle.await?.unwrap_err();
+    ///
+    ///     // Resume the panic on the main task
+    ///     std::panic::resume_unwind(err.into_panic());
+    ///
+    ///     Ok(())
+    /// }
+    /// ```
+    pub fn into_panic(self) -> Box<dyn std::any::Any + Send + 'static> {
+        self.try_into_panic()
+            .expect("should be JobPanicked variant")
+    }
 
-    #[error("channel send error cancelling job")]
-    CancelJobError(#[from] tokio::sync::watch::error::SendError<()>),
-
-    #[error("channel recv error waiting for job results")]
-    JobResultError(#[from] tokio::sync::oneshot::error::RecvError),
-}
-
-impl From<JobHandleError> for JobHandleErrorSync {
-    fn from(e: JobHandleError) -> Self {
-        match e {
-            JobHandleError::JobPanicked(panic_info) => {
-                // we have to convert panic payload to a string because panics can
-                // contain any type, and they are Send but not Sync.
-                let panic_message =
-                    panic_info
-                        .downcast_ref::<String>()
-                        .cloned()
-                        .unwrap_or_else(|| {
-                            if let Some(s) = panic_info.downcast_ref::<&'static str>() {
-                                (*s).to_string()
-                            } else {
-                                format!(
-                                    "Panic occurred with an unsupported payload type: {}",
-                                    std::any::type_name_of_val(&*panic_info)
-                                )
-                            }
-                        });
-                Self::JobPanicked(panic_message)
+    /// Consumes the `JobHandleError`, returning the object with which the task panicked if the task terminated due to a panic. Otherwise, self is returned.
+    ///
+    /// ```should_panic(expected = "boom")
+    /// use neptune_cash::job_queue::JobQueue;
+    /// use neptune_cash::job_queue::traits::*;
+    ///
+    /// struct PanicJob;
+    ///
+    /// #[async_trait::async_trait]
+    /// impl Job for PanicJob {
+    ///     fn is_async(&self) -> bool {
+    ///         true
+    ///     }
+    ///     async fn run_async(&self) -> Box<dyn JobResult> {
+    ///        panic!("{}", "boom");
+    ///     }
+    /// }
+    ///
+    /// #[tokio::main]
+    /// async fn main() -> anyhow::Result<()> {
+    ///     let job_queue = JobQueue::start();
+    ///     let job_handle = job_queue.add_job(PanicJob, 1)?;
+    ///
+    ///     let err = job_handle.await?.unwrap_err();
+    ///
+    ///     // Resume the panic on the main task
+    ///     let panic = err.try_into_panic()?;
+    ///     std::panic::resume_unwind(panic);
+    ///
+    ///     Ok(())
+    /// }
+    /// ```
+    pub fn try_into_panic(self) -> Result<Box<dyn std::any::Any + Send + 'static>, Self> {
+        match self {
+            // Consume self here
+            Self::JobPanicked(arc_mutex) => {
+                // 1. The 1st unwrap() cannot fail because the Arc is never cloned.
+                //    Thus Arc::into_inner() is guaranteed to return Some().
+                // 2. The 2nd unwrap() cannot fail because the Mutex is guaranteed unpoisoned
+                //    because lock() is never called on it.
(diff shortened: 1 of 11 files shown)


src/job_queue/job_result_wrapper.rs
```diff
@@ -0,0 +1,267 @@
+//! This module provides type `JobResultWrapper` to enhance the ergonomics of
+//! working with job-specific result types which must implement the `JobResult`
+//! trait.
+//!
+//! It is useful for:
+//!
+//! 1. returning job results of type T as `Box<dyn JobResults>` when implementing
```
jan-ferdinand, line 7: Once here, once below:

```suggestion
//! 1. returning job results of type T as `Box<dyn JobResult>` when implementing
```

src/job_queue/job_result_wrapper.rs
```diff
@@ -0,0 +1,267 @@
...
+    fn from(wrapper: JobResultWrapper<T>) -> Self {
+        Box::new(wrapper) as Box<dyn JobResult>
+    }
+}
+
+impl<T: Debug> Debug for JobResultWrapper<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("JobResultWrapper").field(&self.0).finish()
+    }
+}
```
jan-ferdinand, line 108: Isn't this the same as a `#[derive(Debug)]` annotation?

src/job_queue/job_result_wrapper.rs
```diff
@@ -0,0 +1,267 @@
...
+//
+// Ideally, this could be expressed more clearly using negative trait bounds
+// (#[feature(negative_impls)]), like this:
+//
+// impl<T: Clone + !Copy> Clone for JobResultWrapper<T> {..}
+// impl<T: Clone + Copy> Clone for JobResultWrapper<T> {..}
+//
+// However, negative trait bounds are not yet stable in Rust as of rustc
+// v1.86.0.
+#[allow(clippy::expl_impl_clone_on_copy)]
```
jan-ferdinand, line 127: I strongly suggest to use `expect` over `allow` wherever possible.

src/job_queue/job_result_wrapper.rs
```diff
@@ -0,0 +1,267 @@
...
+}
+
+impl<T: Display> Display for JobResultWrapper<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+// Clippy warns about a manual Clone impl on a Copy type, but this is necessary
+// to handle the case where T is Clone but not Copy.
```
jan-ferdinand, line 117: This is not true: the trait bound on the generic(s) dictate(s) what is (and is not) derived. [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=445affd5d593c24643fbff89be9f6daa)

Also applies to various other manual trait implementations below.

src/job_queue/mod.rs
```diff
@@ -23,8 +24,354 @@
...
+//! use neptune_cash::job_queue::JobQueue;
+//! use neptune_cash::job_queue::channels::JobCancelReceiver;
+//! use neptune_cash::job_queue::traits::*;
+//! use rand::Rng;
+//!
+//! // ### First lets define some common types ###
+//! // -------------------------------------------
+//!
+//! // define job priority levels for this job-queue.
+//! // note that the queue could process only one type of
```
jan-ferdinand, line 148: Maybe “is able to” instead of “could”?

src/job_queue/queue.rs
```diff
@@ -102,11 +102,17 @@ impl<P: Ord + Send + Sync + 'static> JobQueue<P> {
         Ok(())
     }
 
-    /// adds job to job-queue and returns immediately.
+    /// adds job to job-queue (with interior mutability)
     ///
-    /// job-results can be obtained by via JobHandle::results().await
-    /// The job can be cancelled by JobHandle::cancel()
-    pub fn add_job(&self, job: Box<dyn Job>, priority: P) -> Result<JobHandle, AddJobError> {
+    /// returns immediately (does not wait for job).
```
jan-ferdinand, line 107: “does not wait for job [completion]”?

src/job_queue/queue.rs
```diff
@@ -102,11 +102,17 @@ impl<P: Ord + Send + Sync + 'static> JobQueue<P> {
...
+    ///
+    /// note that this method utilizes interior mutability. Consider calling
+    /// [`Self::add_job_mut()`] instead to make the mutation explicit.
```
jan-ferdinand, line 110: Why have this method at all? Add a (dev-)comment? `#[deprecated]`?

src/models/proof_abstractions/tasm/program.rs
```diff
@@ -142,14 +141,7 @@ pub(crate) async fn prove_consensus_program(
...
     // obtain resulting proof.
-    let result: Result<Proof, ProverJobError> = completion
-        .result()?
-        .into_any()
-        .downcast::<ProverJobResult>()
-        .expect("downcast should succeed, else bug")
-        .into();
-
-    Ok(result?)
+    Ok(ProverJobResult::try_from(completion)?.into_inner()?)
```
jan-ferdinand, line 144: 🙌

jan-ferdinand (commented):
Cool! Great docs, great ergonomics bump.

I've identified one incorrect assumption that can lead to panics and one (type of) boilerplate reduction; see inline comments.

src/job_queue/errors.rs
```diff
@@ -1,73 +1,195 @@
...
+    /// }
+    /// ```
+    pub fn try_into_panic(self) -> Result<Box<dyn std::any::Any + Send + 'static>, Self> {
+        match self {
+            // Consume self here
+            Self::JobPanicked(arc_mutex) => {
+                // 1. The 1st unwrap() cannot fail because the Arc is never cloned.
+                //    Thus Arc::into_inner() is guaranteed to return Some().
+                // 2. The 2nd unwrap() cannot fail because the Mutex is guaranteed unpoisoned
+                //    because lock() is never called on it.
```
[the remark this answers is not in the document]
dan-da: good point.

This issue highlights for me the distinction/asymmetry between `pub struct Bar(u32)` and `enum Foo { Bar(u32) }`.   In the former the wrapped u32 is private; in the latter it is public.  So it easy to confuse the two usages.

The newtype wrapper seems good here.  I will do that.

src/job_queue/errors.rs
```diff
@@ -1,73 +1,195 @@
...
 #[derive(Debug, thiserror::Error)]
 pub enum JobHandleError {
     #[error("the job was cancelled")]
     JobCancelled,
 
+    // note: the Arc<Mutex<_>> is used to make the inner Box `Sync`.
+    // The mutex is never acquired and can thus never be poisoned.
     #[error("the job panicked during processing")]
-    JobPanicked(Box<dyn std::any::Any + Send + 'static>),
+    JobPanicked(Arc<Mutex<Box<dyn std::any::Any + Send + 'static>>>),
```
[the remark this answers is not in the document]
dan-da: tl;dr I'm not opposed to some kind of change if you can prove it out.  But it seems fine as-is to me.

---------

I don't believe the suggested change works here and the compiler and LLM seem to agree.

high level:
1. tokio's JoinError::into_panic() returns the boxed type.
2. We store it (unchanged) in JobHandleError::JobPanicked inside Arc<Mutex<_>>.  This is done in order to satisfy compiler that the type is `Sync`.
3. JobHandleError::try_into_panic() returns the same, unchanged type.

iiuc, you are suggesting to change (2) and (3).  But... even if we could store it as `dyn ...` inside the `Arc<Mutex<_>>` we would still need to box again in (3). 

And further, making the change results in this compiler error:

```
error[E0277]: the size for values of type `(dyn std::any::Any + std::marker::Send + 'static)` cannot be known at compilation time
   --> src/job_queue/errors.rs:144:45
    |
144 |                 Ok(Box::new(Arc::into_inner(arc_mutex).unwrap().into_inner().unwrap()))
    |                             --------------- ^^^^^^^^^ doesn't have a size known at compile-time
```

An LLM has this to say about it:

> Regarding the suggestion to use Arc<Mutex<dyn std::any::Any + Send + 'static>>> for JobPanicked: This would prevent try_into_panic() from working. Arc::into_inner() requires the Arc to have a single owner and returns the inner value by value. Similarly, Mutex::into_inner() returns the inner value by value if the mutex is not poisoned. Since dyn Any is unsized, these into_inner() calls cannot return it directly. The current Box<dyn Any> within JobPanicked ensures a Sized type is managed, allowing try_into_panic() to potentially extract the owned Box when the Arc has a single owner. Directly storing an unsized dyn Any makes this ownership transfer impossible.

src/job_queue/job_result_wrapper.rs
dan-da: no because this is implementing Debug for JobResultWrapper only *if* it is implemented for T.  That permits JobResultWrapper to wrap T (job results) that do not impl Debug.

src/job_queue/queue.rs
dan-da: because neptune-core makes use of it, and it would be a little tricky to change.

In particular:

```
pub struct TritonVmJobQueue(JobQueue<TritonVmJobPriority>)`
pub fn `vm_job_queue() -> Arc<TritonVmJobQueue>` { .. }
```

vm_job_queue is returning a singleton instance in an Arc.  It gets called in a bunch of locations.  Worse though, the resulting value gets passed through a bunch of functions related to composing and proving and maybe upgrading Tx.

To change it, the Arc would need to be changed to an Arc<tokio::sync::Mutex<_>>, or better crate::lock::tokio::AtomicMutex.  And then it needs to be acquired at each usage site, which I think is actually just one, in prove_consensus_program().

It's seems the right thing to do so that all mutability is made explicit, but seems a bit low priority at the moment, given that JobQueue already uses its own Mutex internally.

aside/history: this situation has arisen because the very first JobQueue design used only message passing and the JobQueue struct itself only held a channel-sender for sending to a spawned task that actually proceses jobs.   So that might be considered kind of an obscure form of (logical) interior mutability because the only mutations are occurring inside a different task, on mutable variables.  But anyway the add_job() method didn't appear to need to take &mut self since it only sent a message.   Later, JobQueue was modified to hold a shared mutex over the data, which simplified some things and also so that add_job() could log the number of presently running jobs.  The signature should have changed then, but did not until now.  Also that change was sort of for expediency.... long term I'd like to get back to a pure channel based approach.  In which case, it's debatable if there should be an add_job_mut() or not, since no fields inside JobQueue struct need to change.   thoughts?

src/job_queue/job_result_wrapper.rs
jan-ferdinand: Yeah, it's definitely an oddity of rust. For example, another case where this trips people up is:

```rust
#[derive(Clone)]
struct Foo<T>(Arc<T>); // could be Clone no matter the T, but isn't
```

There are crates like [`derive-where`](https://crates.io/crates/derive-where) that help out in such cases.

src/job_queue/errors.rs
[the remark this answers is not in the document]
jan-ferdinand: Now that I understand the interplay better: the `Box` you need, but the `Arc`, you don't. For `Arc<T>` is to be `Send`, the `T` must _already_ be `Send`, and the same is true for `Sync`, so you're not gaining any thread safety with the `Arc`. Removing the `Arc` is trivial since it implements `Deref` and you're only constructing it in 2 places:

```
> git diff --stat
 src/job_queue/errors.rs         | 7 +++----
 src/job_queue/job_completion.rs | 3 +--
 2 files changed, 4 insertions(+), 6 deletions(-)
```

The `Mutex<T>` is critical because it is `Sync` (which you need) if `T` is `Send` (which it is). Other candidates that have the same property would be [`AtomicPtr`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicPtr.html), [`ReentrantLock`](https://doc.rust-lang.org/std/sync/struct.ReentrantLock.html) (nightly only), and [`Exclusive`](https://doc.rust-lang.org/std/sync/struct.Exclusive.html) (nightly only). Out of all these, `Exclusive` sounds like it's expressing the desired properties the best. Until it is stabilized, `Mutex` seems like a fine stand-in.

src/job_queue/queue.rs
jan-ferdinand: > thoughts?

Not many helpful ones. Overall, it sounds like the design and the API are still somewhat in flux. I think it's fine to have both these methods around while that's true. And the note regarding priorities also feels apt. Thanks for the context. :+1:

src/job_queue/errors.rs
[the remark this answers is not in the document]
dan-da: addressed in 8b189964d2c0fd51b8e9c4f7899b2f713bc95b99 with `PanicInfo` newtype.

src/job_queue/errors.rs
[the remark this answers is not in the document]
dan-da: removed Arc in 8b189964d2c0fd51b8e9c4f7899b2f713bc95b99.  thx!

src/job_queue/queue.rs
dan-da: addressed in 8b189964d2c0fd51b8e9c4f7899b2f713bc95b99

src/job_queue/mod.rs
dan-da: addressed in 8b189964d2c0fd51b8e9c4f7899b2f713bc95b99.  removed superfluous line.

src/job_queue/job_result_wrapper.rs
dan-da: no longer relevant as of 8b189964d2c0fd51b8e9c4f7899b2f713bc95b99.  trait is derived instead.

src/job_queue/job_result_wrapper.rs
dan-da: addressed in 8b189964d2c0fd51b8e9c4f7899b2f713bc95b99.  Debug is derived.

src/job_queue/job_result_wrapper.rs
dan-da: addressed in 8b189964d2c0fd51b8e9c4f7899b2f713bc95b99

dan-da:
all suggestions have been addressed in 8b189964d2c0fd51b8e9c4f7899b2f713bc95b99.   thx @jan-ferdinand for the review.

PR #587, opened 2025-05-12
Thv/benchmark wallet state update

Sword-Smith:
Huge performance boost (2x) from the `assert` to `debug_assert` change. But I'm mostly interested in feedback on the benchmark, and how to handle state there.


dan-da:
seems like a good speedup.  :-)

I don't think the benchmarks.rs belongs in src/api as the code seems quite specific to that benchmark.

Also I think the benchmark can leverage the new api stuff to be more concise.

It's easier just to demonstrate though, so I've modified the wallet_state bench, and also made a few more things pub in neptune_core.

I will push a branch shortly with the suggested changes.

dan-da:
here is a branch with suggested changes.  At minimum I don't think benchmark.rs belongs in src/api.

The rest of the changes demonstrate an alternate way to write the benchmark, based on 45e888d, and a few tweaks to neptune-core code to facilitate.

danda/benchmark-wallet-state-update  (in this repo)

dan-da (changes requested):
Since I see more commits that do not yet address my previous comments, I will re-iterate the request to remove benchmarks.rs from src/api with some additional context/rationale, and make this a more official "request changes".   rationale:

1. I'm not convinced that benchmark related code belongs in src/api at all, even general purpose types.  A case could be made for that, but so far has not been advanced.
2. the contents seem tailored to the needs of wallet_state.rs benchmark, not general purpose.
3. the content and location of benchmarks.rs is inconsistent with other types in src/api, ie not following convention.  If we were to support a benchmarks api, it should be in its own sub-directory and probably should define a Benchmark type, and crate::api::Api should have a method for instantiating it.
4. it violates the rules of src/api module, as specified in src/api/README-dev.md.  In particular rules 3, 4, 6, and 8 which can be summarized as "no panics", "return real error types",  "doc-comment all pub items", and "make integration test".

I can suggest a few alternative (and easier) approaches:

1. move contents of benchmarks.rs into the wallet_state.rs benchmark.  Then make the methods it calls from neptune_cash pub as necessary.  Probably the benchmark is just highlighting things that should be pub anyway, eg for 3rd party wallet usage.
2. move the contents of benchmarks.rs elsewhere under src.
3. use the approach I demonstrated in branch:  danda/benchmark-wallet-state-update

Sword-Smith:
> Since I see more commits that do not yet address my previous comments ...

Just rebasing, no new commits. The code from `api/benchmarks` is being moved.

dan-da:
> There's _a lot_ of code being made public with this PR. I'm not a fan of that since it means that we will not be warned about unused code.

yeah...  there are quite a few things that are presently pub(crate) in the codebase that will likely need to be made pub for the wallet competition.  And this benchmark is probably finding some of those.   But then it may be exposing others that really needn't be pub, except for the benchmark.  So that's a tricky line to walk.

A few thoughts:

1. @jan-ferdinand had suggested a way to mark a pub fn as not visible in docs somehow, so it is not very discoverable and kind of a hint not to use it.

2. consider the other approach I suggested:  "move the contents of benchmarks.rs elsewhere under src".  Then we need not make anything else pub.  My concern was mainly with putting in src/api, a carefully curated module.  But perhaps it could live under src/util_types/benchmarks or something.  It could also be marked with the attribute referenced in (1).

3. The branch I created was an attempt to make the benchmark without need to expose many new APIs.  It definitely has some warts, but shows it is mostly possible.

Personally I would lean towards (2) at this point, but I don't feel strongly about it.

@jan-ferdinand and @aszepieniec might have opinions to offer.

benches/wallet_state.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    /// Maintain 100 membership proofs, while receiving an additional 100 UTXOs.
+    mod maintain_100_100 {
+
+        use super::*;
+        use crate::helper::*;
+
+        fn update_wallet_with_block2(bencher: Bencher) {
+            let rt = tokio::runtime::Runtime::new().unwrap();
+            let mut wallet_state = rt.block_on(devops_wallet_state_genesis(Network::Main));
```
dan-da, line 48: style: consider a single block_on() with a single async block.

(except for the bench closure which needs its own)

jan-ferdinand:
- The annotation to hide an item from the public API is [`#[doc(hidden]`](https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#hidden). It is more than a hint: items marked this way are not part of the public API. Tools like `cargo-semver-check` will [correctly pick this up](https://predr.ag/blog/checking-semver-for-doc-hidden-items/).
- It might be possible to use a crate like [`visibility`](https://crates.io/crates/visibility) to modify an item's visibility based on the compilation profile. I haven't tested it, but the annotation `#[cfg_attr(bench, visibility::make(pub))]` on a private item might achieve the desired effect.

aszepieniec:
> It might be possible to use a crate like [visibility](https://crates.io/crates/visibility) to modify an item's visibility based on the compilation profile. I haven't tested it, but the annotation #[cfg_attr(bench, visibility::make(pub))] on a private item might achieve the desired effect.

OOooh, I like it!

jan-ferdinand:
> I haven't tested it, but the annotation `#[cfg_attr(bench, visibility::make(pub))]` on a private item might achieve the desired effect.

Unfortunately, this doesn't work _as_ easily. The `bench` attribute does not exist in this form. What you can do instead (and this time, I have tried it):
- declare a new feature in the Cargo.toml, for example `benchmarking = []`
- annotate the item in question with `#[cfg_attr(feature = "benchmarking", visibility::make(pub))]`
- use the annotated item in a benchmark
- run `cargo bench --features benchmarking`

The downside is that compilation will fail with a simple `cargo bench`, since the items in question will have the wrong visibility.

Sword-Smith:
I managed to better encapsulate the stuff needed for this benchmark. So now only two functions go from `pub(crate)` to `pub`. I'm very happy with where this has landed.

Also: This 2x speedup in wallet state update is something we need to roll out ASAP. State updates to the wallet are extremely slow, and by an order of magnitude the bottleneck for state updates, as evidenced by the below log showing a node with ~400 UTXOs applying a block state update.

```
2025-05-22T11:11:55.831666305Z DEBUG ThreadId(01) neptune_cash::models::state::wallet::wallet_state: Found valid mp for UTXO with leaf index: 64276
2025-05-22T11:11:55.831678729Z DEBUG ThreadId(01) neptune_cash::models::state::wallet::wallet_state: doing maintenance on 407/407 monitored UTXOs
2025-05-22T11:11:56.548856821Z DEBUG ThreadId(01) neptune_cash::models::state::wallet::wallet_state: Block has 0 removal records
2025-05-22T11:11:56.567658622Z DEBUG ThreadId(01) neptune_cash::models::state::wallet::wallet_state: Number of mutated membership proofs: 1
2025-05-22T11:11:57.429364122Z DEBUG ThreadId(01) neptune_cash::models::state: Flushed all databases
2025-05-22T11:11:57.429395871Z DEBUG ThreadId(01) neptune_cash: executed set_new_tip_internal() in 1.6940187899999999 secs.  location: /home/thv/repos/neptune-core/src/models/state/mod.rs:1409:9
2025-05-22T11:11:57.429418544Z DEBUG ThreadId(01) neptune_cash::main_loop: Storing block 08338635382424590278,11913466815894687866,17487415327791684952,09982352714201000328,00000000000000110312 in database. Height: 14865, Mined: 2025-05-22T11:21:54.067+02:0
```

dan-da:
> Also: This 2x speedup in wallet state update is something we need to roll out ASAP.

If urgent, then I would suggest backporting this specific feature onto a branch at v0.2.2, and releasing an update of that.

I think we are still some weeks away from being able to properly release something from master:

1. there are still some open issues tagged as release-blocker that need resolving.
2. once a release-candidate is tagged, we are targetting a two week testing period on testnet-mock (and anywhere else)
3. we also need to write a basis automated test harness for the release-candidate running on testnet-mock.

PR #588, opened 2025-05-12
fix: avoid panic in TransactionDetails::new()

### LINKED ISSUE #586: Bug in `TransactionDetails::new`
The regular subtraction here should be a `checked_sub`.

```rust
if total_spend < total_spendable {
    let diff = total_spend - total_spendable;
    bail!("Missing change output in the amount of {}", diff);
}
```

Error:
```
thread 'main' panicked at src/models/blockchain/type_scripts/native_currency_amount.rs:368:9:
Cannot subtract `NativeCurrencyAmount`s; use `checked_sub` instead.
stack backtrace:
```

dan-da:
This PR addresses #586 by removing the offending line completely.  It was only used for creating an error message and that is better done by returning a thiserror derived error.

---------

closes #586.

modifies `TransactionDetails::new()` to return a new error type `TransactionDetailsError` instead of using anyhow::bail!().

This then enables us to map that error into CreateTxError and get rid of the ugly CreateTxError::Failed variant that was a catch-all for anyhow errors.

------

edit: I have a small concern about adding a new variant CreateTxError::InvalidMutatorSetMembershipProof because this is kind of an internal error state, not caused by the caller initiating the tx.  But I'm not sure what else to do with it.

edit: I should also note that CreateTxError is not yet present in a release, so removal of Failed variant is a non-breaking change in that sense.

### DIFF
--- a/src/api/tx_initiation/error.rs
+++ b/src/api/tx_initiation/error.rs
@@ -9,6 +9,7 @@ use crate::job_queue::errors::AddJobError;
 use crate::job_queue::errors::JobHandleError;
 use crate::models::blockchain::transaction::transaction_proof::TransactionProofType;
 use crate::models::proof_abstractions::tasm::prover_job::ProverJobError;
+use crate::models::state::transaction_details::TransactionDetailsError;
 use crate::models::state::tx_proving_capability::TxProvingCapability;
 
 /// enumerates possible transaction send errors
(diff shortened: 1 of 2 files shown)


dan-da:
I'm also sort of debating with myself whether TransactionDetails should perform any validation at all.

1. TransactionDetailsBuilder already does some basic validation.
2. the `Transaction` itself can be truly validated by calling verify_proof().

So I think I'm convincing myself it's best that TransactionDetails skip the validations and just return `Self`.

what do reviewers think?

dan-da:
superceded by #590.  closing.

Issue #517, opened 2025-03-20
Goal: Make all of state restorable from blocks in block-index database


Sword-Smith:
Only these two parts of the state, the blocks stored in the `*.blk` files and the block index database, the rest of the state should be recoverable. If I delete everything but these two directories, the node currently panics in the wallet-database recovery because there is no archival mutator set that matches the state of the block index database. So, as a minimum, the archival mutator set must be rebuilt from the block data, if the archival mutator set has been deleted.

Sword-Smith:
Closed by #592

Issue #585, opened 2025-05-12
difficulty syncing blocks


dan-da:
note: I'm going from memory, so this report is a little vague on details:

-------

I have a node behind NAT that is having a lot of difficulty connecting to peers and syncing blocks.   I've been attempting to get it synced for a couple weeks now.   Well most of that time was just letting it sit and hoping ti would sync but it never did.

This node was at around block 6438, from late february when I started the attempt, so it was missing a couple of months worth of blocks, but not starting from genesis.

The first issue I encountered is difficulty connecting to peers.   Using the bootstrap peers listed in the neptune-core README.md, I was either not able to establish connection, or refused, or perhaps a connection was established, but nothing beyond that.  The node never did any successful peer discovery and didn't seem to be attempting anything much in that area from the logs.     Adding some additional known peers I was finally able to get 100 blocks synced, but encountered many many "sync timeout" errors and punished peers over and over for that.  After that I let it run for week(s) and no further blocks were synced.

Today I tried again, armed with an updated list of potential peers.   I was able to establish 4 peer connections (Still not a lot) out of perhaps 20 tried.   Once again I am getting a lot of sync timeout errors, and have so far been able to sync only 100 blocks.  Also the "syncing" flag in dashboard is set to false although the block height is only 6638.

Since I assume others are able to sync, I'm guessing my issues may be related to NAT somehow and/or my relatively slow connection.  Anyway it doesn't seem to be working "well".

![Image](https://github.com/user-attachments/assets/c8df4df6-f704-45f9-87b4-01c24ef66ec5)

This screenshot is from today after a restart, new peers, and attempting for about 3 hours, with just 100 blocks synced so far.

dan-da:
about 13 hours later, the node has 6 peers (gained 2) and has synced to block 7638 (from 6638), so that is (slow) progress at least.  There are many occurrences of timed-out sync challenge response.

![Image](https://github.com/user-attachments/assets/6d22f5f8-e6fc-4acb-927b-327de57abab1)

dan-da:
about 26 hours later, the node has synced to block 9388 (apr 15) from 7638.  That's only about 67 blocks per hour.   There are presently 6 peers.

![Image](https://github.com/user-attachments/assets/a55f9397-2324-4f51-81c9-c8e2e0719ac4)

Also most of the time dashboard says "synchronizing: false".

![Image](https://github.com/user-attachments/assets/26204823-0de2-4bf6-a791-0a089c664ae6)

[...]

PR #592, opened 2025-05-14
Bootstrap with downloaded blocks

Sword-Smith:


Sword-Smith:
There is a bug here: The block validation does not take into account that the previously stored block might not be the tip. In case of a reorganization on the machine that produces the `blocks` directory (where the "blk" files are stored), multiple blocks of a specific height might be stored. So if the next processed block is not a successor of the previous, the correct previous block must be found, in case -- in the case where block validation is happening. If there is not block validation happening, this bug is never triggered.

dan-da:
> No description provided.

Please provide a description of:
1. What issue(s) this is addressing.
2. How it addresses it.

It might seem obvious, but the context helps before digging into details of a PR.

Sword-Smith:
Closes #517. Addresses #585.

dan-da:
I'm unclear why this was merged without an approving review.    I asked for a description of the PR before I begin reviewing and was waiting for one, but instead it was merged.

I see a commit title that worries me:

> feat: add option to ignore block-validation on block-bootstrapping

Without further context, it is unclear whether block validation is ignored completely, or just deferred to a later stage.   If the former, I'd be curious to hear the justification, as that seems unsound on the face of it.

Sword-Smith:
1. Yes, your guess is correct. We can't yet recover from block data alone. That kind of recovery only works through the bootstrap mode added in this PR.
2. Yes bootstrapping works in this scenario. Program will print a warning message for each block already in state that's also present in the bootstrapping source data.
3. Everything works hunky-dorey.

Issue #567, opened 2025-04-25
Modularize Mempool


jan-ferdinand:
Currently, the mempool is rather tightly integrated with various other aspects of the client. In order to tackle issues like #489, #542, or #560 fundamentally rather than ad-hoc, the mempool should be re-designed. Loose coupling and a clean interface should be aspired to. The following are my notes from trying to understand the current and future requirements and uses of the mempool. Since the current mempool is integrated tightly, it's easy to overlook important requirements and uses. If you identify any such misses, please notify me about them.

Once all requirements are identified, a new mempool design is in order. Help and support during this and all other phases is appreciated.

## Requirements

- Wallet: Query for or get informed about unconfirmed transactions impacting the balance.
- Upgrader: Track transactions that send money to us, including fees and other sources.
- Miner: Find transactions with most profitable fees for given space. (Knapsack problem.)
- Favor transactions with higher proof quality, especially when they relate to the same transaction kernel. (Single Proof > Proof Collection > Primitive Witness)
- No “inclusion jitter”. A transaction that's being worked on (e.g., a proof upgrade, transaction update, …) remains in the pool until the work is done, then gets swapped atomically.
- Handle chain re-organizations gracefully: allow out-of-date transactions in pool, anticipating a re-org that means they are up-to-date again.
- Handle conflicting transactions gracefully. (Currently: The same input must never be spent by any 2 transactions in the mempool. I don't understand the implications of this policy yet.)
- Admit out-of-date transactions, especially if they benefit us. They should be updated (by us or the network) for inclusion in a future block; they should not be discarded by the entire network.

## Current use

- Transaction updating when receiving a new block is done in the main loop's `handle_peer_task_message()` via `spawn_mempool_txs_update_job()`. The main loop then directly modifies the mempool's internals by overwriting the out-of-date transaction with the updated transaction.
- Transaction upgrading is done in the main loop via scheduled task `proof_upgrader()`. The corresponding file `proof_upgrader.rs` has its own function `get_upgrade_task_from_mempool()`. The proof upgrader's `handle_upgrade()` then removes or overwrites (?) the transaction as appropriate via some helper functions.
- Helper functions `mempool_{insert, remove, clear, …}` on GlobalState are a shim of and duplicate some basic mempool API and notify the wallet about the corresponding events. This looks like an ad-hoc instance of the subscriber pattern to me.
- …

dan-da:
#489 may be worth consideration.

[...]

PR #593, opened 2025-05-14
Re-design Mempool

jan-ferdinand:
Replace the existing mempool (also “transaction pool”) with a new design, as discussed in #567, and reflect 3-step mining (which was always de-facto present) explicitly in the code.

This draft PR is intended as an early review opportunity for all interested parties. In particular, factoring out the now-explicit first step of 3-step mining, “upgrading”, is a non-trivial task: because the step wasn't clearly identified before, it has snuck into various different parts of the codebase.

### DIFF
--- a/src/config_models/cli_args.rs
+++ b/src/config_models/cli_args.rs
@@ -102,17 +102,30 @@ pub struct Args {
     #[clap(long, alias = "notx")]
     pub(crate) no_transaction_initiation: bool,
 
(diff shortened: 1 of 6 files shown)


aszepieniec:
image

Just :heart: already :-)

jan-ferdinand:
> Just ❤️ already :-)

Can't promise it's gonna stay like this. A _lot_ of implementation is currently missing. 🤐

Sword-Smith:
Superseded by #610.

PR #594, opened 2025-05-14
test(`Chunk`): Add test computing statistics of `Chunk` lengths

aszepieniec:
Add a test that computes statistics about the number of elements in `Chunk`s. Based on these statistics, it is justifiable to use 12 bits to store the length.

### DIFF
--- a/src/util_types/mutator_set/chunk.rs
+++ b/src/util_types/mutator_set/chunk.rs
@@ -132,11 +132,14 @@ impl<'a> Arbitrary<'a> for Chunk {
 #[cfg(test)]
 #[cfg_attr(coverage_nightly, coverage(off))]
 mod tests {
-    use std::collections::HashSet;
+    use std::collections::{HashMap, HashSet};
 
     use num_traits::Zero;
-    use rand::RngCore;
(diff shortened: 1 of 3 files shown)


src/util_types/mutator_set/chunk.rs
```diff
@@ -311,4 +314,69 @@ mod tests {
 
         assert_eq!(chunk, decoded);
     }
+
+    #[ignore = "informative statistics"]
```
dan-da, line 318: Please add a comment describing the purpose of the test and any additional context in your head.

dan-da (changes requested):
I'd like to see a comment above the test describing it.  otherwise, lgtm.

dan-da (approved):
Thanks for the comment, that's good context.

One observation:  this "test" appears to be more about gathering statistics, possibly a one-time kind of activity, than about verifying correct functionality.   Or possibly an "over-time" activity to measure improvement.  Either way, I wonder if we should establish some kind of convention for these types of stats-gathering "tests", such as:

```
#[cfg(test)]
mod stats {
   ...
}
```

PR #603, opened 2025-05-24
Correct the proving-capability abstraction for usage with triton-vm Proof (as opposed to TransactionProof)

### LINKED ISSUE #576: Rename and refactor TxProvingCapability to use correct abstraction
There has been a discussion about TxProvingCapability and the fact that as currently implemented it is a bad fit for ProofBuilder and ProverJob which deal solely with Proofs.

This "bad fit" is because TxProvingCapability is presently just an enum with variants for PrimitiveWitness, ProofCollection, and SIngleProof.   These variants are tied to transactions proofs, but `Proof` is a more general and abstract type than that.

So the general idea then is rename TxProvingCapability to just `ProvingCapability` and make it reflect a machine's actual ability to generate proofs.   Concretely by including a field for max_log2_padded_height.   Although there may be other factor(s).

Here is where this change is first proposed:
https://github.com/Neptune-Crypto/neptune-core/pull/550#discussion_r2058759243

The new type might look something like:

```
pub VmProvingCapability {
  // from CLI, or auto determined.
   max_log2_padded_height: u8,
}

impl VmProvingCapability {
    pub async fn ensure_can_prove_claim_triple(
        &self,
        program: Program,
        claim: Claim,
        nondeterminism: NonDeterminism,
    ) -> bool {
        let mut vmstate = VMState::new(program, claim.input.into(), nondeterminism);
        self.can_prove_claim(&mut vmstate).await
    }

    pub async fn ensure_can_prove_claim(&self, vmstate: &mut VMState) -> bool {
        // note: this would run inside spawn_blocking().
        if vmstate.run().is_ok() {
            let big_max = 2u32.pow(max.into());
            return vmstate.cycle_count.next_power_of_two() < big_max;
        }
        false
    }

    // automatically detect the max_log2_padded_height for this device.
    pub async fn auto_detect() -> Self {
        // how to impl this?
    }

    // the above methods may be all we will need.
    // below methods demonstrate how we could accommodate arbitrary classes of claims/proofs, old or new.
    // and support fast (estimated) checks based on pre-determined log2_padded_height values.
    // note: all integer constants are basically just made up and would need to be determined.

    fn can_prove(&self, log2_padded_height: u8) -> bool {
        self.max_log2_padded_height >= log2_padded_height
    }

    pub fn can_prove_tx_proof(&self, txp: TransactionProofType) -> bool {
        match txp {
            TransactionProofType::SingleProof => self.can_prove_tx_single_proof(),
            TransactionProofType::ProofCollection => self.can_prove_tx_proof_collection(),
            TransactionProofType::PrimitiveWitness => true,
        }
    }

    pub fn can_prove_tx_single_proof(&self) -> bool {
        self.can_prove(22);
    }

    pub fn can_prove_tx_proof_collection(&self) -> bool {
        self.can_prove(11);
    }

    pub fn can_prove_block_proof(&self) -> bool {
        self.can_prove_single_proof();   // not sure if this is right
    }

    pub fn can_prove_kernel_to_outputs(&self) -> bool {
        self.can_prove(10);
    }

    pub fn can_prove_lockscript(&self) -> bool {
        self.can_prove(6);
    }

    pub fn can_prove_typescript(&self) -> bool {
        self.can_prove(5);
    }

    /// and so on with any type.   

    // note that we could get much fancier, eg by using a trait:

    pub fn can_prove_provable_type(&self, provable: &impl ProvableType) -> bool {
        self.can_prove(provable_type.log2_padded_height_estimate());
    }
```

`note: improvements have been suggested in the original thread.  the above is just a starting point`.

An unresolved (not previously asked) question is:  how can we auto-detect a machine's max log2_padded_height capability?   ie, what does the implementation of the `auto_detect()` method look like?   I think that worst case, we can do something similar to cli_args::Args::proving_capability() which check RAM and CPU capacities and maps those to specific values.  But maybe there is a better and more precise approach, eg running some kind of (quick) stress test?    @jan-ferdinand I'm hoping you will have some ideas/suggestions here.

dan-da:
note to reviewers:  the primary changes are in files: vm_proving_capability.rs, cli_args.rs, transaction_proof_builder.rs, prover_job.rs, and triton_vm_job_queue/mod.rs.   Other changes are mainly dictated by the modifications in these files.

---------
edit: reviewers, please review and double check the two logging modes for vm_state (proof inputs).  These are described fully in src/util_types/log_vm_state.rs.   They are used in prover_job.rs (logs all proofs) and transaction_proof_builder.rs (logs only single-proofs without primitive-witness claims).  In particular, it is important that we not leak secrets in the NoWitness mode that could result in loss-of-funds if the resulting json file is shared.  Files of both types can be generated via:
```
mkdir /tmp/vm_state
mkdir /tmp/vm_state_no_witness

NEPTUNE_VM_STATE_DIR=/tmp/vm_state cargo test
NEPTUNE_VM_STATE_NO_WITNESS_DIR=/tmp/vm_state_no_witness cargo test
```
(note that each mode can be independently enabled)

The /tmp/vm_state directory will contain many files and is around 5.8Gb for me, while the other dir is much smaller.

------------

also, in testing I found that some ProofCollection proofs were requiring log2-padded-height of 15, so I changed TransactionProofType::ProofCollection variant to specify 15 instead of 11. 

commit text follows and describes the changes well.

-------

refactor: make proving-capability correct for proving

closes #576.

TxProvingCapability was being used by ProofBuilder and by ProvingJob but its
variants were logically incorrect for these usages because these types only care
about NeptuneProof, not about TransactionProof variants, which
TxProvingCapability mirrored.

This commit fixes that by renaming enum TxProvingCapability to struct
VmProvingCapability which is a newtype for log2(padded_height).

Some impl<From> are added to TransactionProofType that make it easy to convert
to a VmProvingCapability and vice-versa.

capability detection is moved from cli_args::Args into
VmProvingCapability::auto_detect().  Proof-collection is now log2-padded-height
of 15, rather than 11.  This is because some tests were failing and required 15
so it is experimentally determined.

changes --tx-proving-capability to --vm-proving-capability <int>.

removes --max_log2_padded_height_for_proofs as --vm-proving-capability has the
same meaning.

removes fields proof_type and max_log2_padded_height from ProverJobSettings, as
they are no longer relevant.  proof_type makes no sense there and
max_log2_padded_height is present inside vm_proving_capability.

adds method transaction_proof_type() to TransactionProofBuilder, where it
actually does make sense.

updates tests, adds docs and examples.

Squashed:

+ removed proof_type and max_log2_padded_height from ProverJobSettings.
  Added transaction_proof_type to TransactionProofBuilder.
+ re-implemented TxProvingCapability as struct, not enum.
+ rename TxProvingCapability VmProvingCapability
+ default tx_proof_type to best proof capable of.  other misc
+ move capability auto-detection from Args into VmProvingCapability
+ ProofCollection now 15 instead of 11

----------

feat: add env var VM_JOB_QUEUE_SHARED for tests

This adds a kind of turbo-mode for running unit tests.

Normally the job-queue is a shared singleton instance.

When running unit tests, a shared instance has these characteristics:

1. tests must wait for each other's jobs to complete. CPU usage tends to
    be low for much of the testing duration.
2. it is possible to run all tests in parallel and generate proofs when
    proofs do not exist in local cache or on proof-server.
3. total time for running all tests increases substantially compared to
    scenario where each test using its own job-queue instance.

A non-shared instance has these characteristics:

1. tests run independently and use up all CPU cores.
2. it is not possible to run all tests in parallel and generate proofs
    as it would exhaust device's resources, especially RAM.  This mode
    only works well when proofs are already cached.  A workaround is to
    run with --test-threads 1 to generate proofs.
3. total time for running all tests decrease substantially (assuming
    proofs are cached) vs the shared-instance scenario.

When running unit tests, shared mode is the default. The mode can now
be selected at runtime via:
```
VM_JOB_QUEUE_SHARED=false cargo test <args>

VM_JOB_QUEUE_SHARED=true cargo test <args>
```

----------

feat: env LOG2_PADDED_HEIGHT_METHOD for proving capability

A previous commit made calculation of log2(padded-height) about 4x slower for
ProofBuilder and ProverJob because it used a different calculation method.  This
commit returns to the baseline method and speed, but enables runtime selection
via environment variable.

By default the log2(padded-height) is calculated using VmState::run().

A more accurate but slower way is to use VM::trace_execution_of_state().
This is typically about 4x slower.

And the fastest method is to skip running the program entirely.  But
that option is only available when running unit tests.

These methods can be selected at runtime:

```
LOG2_PADDED_HEIGHT_METHOD=trace neptune-core <args>
LOG2_PADDED_HEIGHT_METHOD=run neptune-core <args>

LOG2_PADDED_HEIGHT_METHOD=skip cargo test <args>
```

----------------

    chore: improvements to VmProvingCapability
    
    addresses #603 review comments.  In particular:
    
    + suggest padded-height value in --vm-proving-capability help text for each
      TransactionProofType variant.
    + add dep on const_format crate, for formatting const strings
    + handle error for VMState::run() instead of unwrap()
    + writes vm state to disk if NEPTUNE_WRITE_VM_STATE_DIR is set
    + warn instead of panic if total memory is reported as zero.
    + adds back some asserts
    + make some enums non_exhaustive
    + improve some doc-comments
    
    Also:
    + renames LOG2_PADDED_HEIGHT_METHOD to NEPTUNE_LOG2_PADDED_HEIGHT_METHOD
    + removes unwrap() when calling spawn_blocking()

--------------------
    chore: make vm_state logging more flexible
    
    Logging proof inputs can be a security risk if the claims contain
    primitive-witness data (secrets).
    
    This commit creates two logging modes: Proof and NoWitness.
    
    Proof mode logs all proofs processed by the ProverJob.
    It is enabled by setting NEPTUNE_VM_STATE_DIR=<dir>
    
    NoWitness mode logs only SingleProof that are produced from non
    primitive-witness claims.  These are logged from TransactionProofBuilder.
    It is enabled by setting NEPTUNE_VM_STATE_NO_WITNESS_DIR=<dir>
    
    A new module, log_vm_state, is introduced that holds the logging code.

### DIFF
--- a/src/api/tx_initiation/builder/transaction_proof_builder.rs
+++ b/src/api/tx_initiation/builder/transaction_proof_builder.rs
@@ -351,7 +351,8 @@ where
     // log proof inputs if matching env var is set. (does not expose witness secrets)
     // maybe_write() logs warning if error occurs; we ignore any error.
     let _ = log_vm_state::maybe_write(
-        LogProofInputsType::NoWitness,
+        LogProofInputsType::DoesNotContainWalletSecrets,
+        SingleProof.program(),
         &claim,
         nondeterminism.clone(),
     );
--- a/src/config_models/cli_args.rs
+++ b/src/config_models/cli_args.rs
@@ -383,7 +383,7 @@ pub struct Args {
     pub(crate) scan_keys: Option<usize>,
 }
 
-pub const VM_PROVING_CAPABILITY_HELP: &str = const_format::formatcp!(
+const VM_PROVING_CAPABILITY_HELP: &str = const_format::formatcp!(
     "\
 Specifies device's capability to generate TritonVM proofs
 
--- a/src/models/proof_abstractions/tasm/prover_job.rs
+++ b/src/models/proof_abstractions/tasm/prover_job.rs
@@ -178,9 +178,12 @@ impl ProverJob {
     async fn prove(&self, rx: JobCancelReceiver) -> JobCompletion {
         // log proof inputs if matching env var is set (may expose witness secrets)
         // maybe_write() logs warning if error occurs; we ignore any error.
-        let _ = log_vm_state::maybe_write(LogProofInputsType::Proof, &self.claim, || {
-            self.nondeterminism.clone()
-        });
+        let _ = log_vm_state::maybe_write(
+            LogProofInputsType::MayContainWalletSecrets,
+            self.program.clone(),
+            &self.claim,
+            || self.nondeterminism.clone(),
+        );
 
         // produce mock proofs if network so requires. (ie RegTest)
         if self.job_settings.network.use_mock_proof() {
--- a/src/util_types/log_vm_state.rs
+++ b/src/util_types/log_vm_state.rs
@@ -2,8 +2,7 @@ use std::io::Write;
 use std::path::PathBuf;
 
 use crate::models::blockchain::shared::Hash as NeptuneHash;
-use crate::models::blockchain::transaction::validity::single_proof::SingleProof;
-use crate::models::proof_abstractions::tasm::program::ConsensusProgram;
+use crate::triton_vm::prelude::Program;
 use crate::triton_vm::proof::Claim;
 use crate::triton_vm::vm::NonDeterminism;
 use crate::triton_vm::vm::VMState;
@@ -12,24 +11,25 @@ use crate::triton_vm::vm::VMState;
 ///
 /// this type facilitates offering distinct logging modes for:
 ///
-/// 1. `Proof` (any kind of Proof)
-/// 2. `NoWitness`
+/// 1. `MayContainWalletSecrets` (any kind of Proof)
+/// 2. `DoesNotContainWalletSecrets`
 ///
-/// The `Proof` mode is useful for logging inputs to every single Proof that is
-/// generated.  These include proofs generated from `PrimitiveWitness` meaning
-/// that the claims expose secrets and should not be shared. This mode of
-/// logging is considered a security risk, but can be useful for investigating
-/// or researching alone, or on testnet(s), etc.
+/// The `MayContainWalletSecrets` mode is useful for logging inputs to every
+/// single Proof that is generated.  These include proofs generated from
+/// `PrimitiveWitness` meaning that the claims expose secrets and should not be
+/// shared. This mode of logging is considered a security risk, but can be
+/// useful for investigating or researching alone, or on testnet(s), etc.
 ///
-/// The `NoWitness` mode is useful when logging proof inputs for purposes of
-/// sharing with others, eg neptune-core developers for debugging.  However it
-/// does not log any Proofs generated from a `PrimitiveWitness`.
+/// The `DoesNotContainWalletSecrets` mode is useful when logging proof inputs for
+/// purposes of sharing with others, eg neptune-core developers for debugging.
+/// However it does not log any Proofs generated from a `PrimitiveWitness`
+/// and thus does not leak wallet secrets.
 pub enum LogProofInputsType {
-    /// log any inputs of any kind of proof, possibly leaking witness secrets
-    Proof,
+    /// log proof inputs that may contain wallet secrets
+    MayContainWalletSecrets,
 
-    /// log inputs of proof, without leaking witness secrets
-    NoWitness,
+    /// log proof inputs that do not contain wallet secrets
+    DoesNotContainWalletSecrets,
 }
 
 impl LogProofInputsType {
@@ -39,16 +39,16 @@ impl LogProofInputsType {
     /// directory in which to write proof files.
     pub const fn env_var_name(&self) -> &str {
         match *self {
-            Self::Proof => "NEPTUNE_VM_STATE_DIR",
-            Self::NoWitness => "NEPTUNE_VM_STATE_NO_WITNESS_DIR",
+            Self::MayContainWalletSecrets => "NEPTUNE_VM_STATE_WITH_SECRETS_DIR",
+            Self::DoesNotContainWalletSecrets => "NEPTUNE_VM_STATE_NO_SECRETS_DIR",
         }
     }
 
     /// returns file name prefix
     pub const fn file_prefix(&self) -> &str {
         match *self {
-            Self::Proof => "vm_state.claim",
-            Self::NoWitness => "vm_state.no_witness_claim",
+            Self::MayContainWalletSecrets => "vm_state.unsafe_to_share",
+            Self::DoesNotContainWalletSecrets => "vm_state.safe_to_share",
         }
     }
 }
@@ -71,16 +71,18 @@ impl LogProofInputsType {
 ///
 /// Security:
 ///
-/// Files of type [LogProofInputsType::Proof] should only be used for debugging
-/// by the wallet owner as they may contain wallet secrets.
+/// Files of type [LogProofInputsType::MayContainWalletSecrets] should only be
+/// used for debugging by the wallet owner as they may contain wallet secrets.
 ///
-/// Files of type [LogProofInputsType::NoWitness] can be shared with others
-/// eg, neptune-core developers, for purposes of debugging/assistance.
+/// Files of type [LogProofInputsType::DoesNotContainWalletSecrets] can be shared
+/// with others eg, neptune-core developers, for purposes of
+/// debugging/assistance.
 ///
 /// It is the *callers* responsibility to ensure that the provided claim matches
 /// the `log_proof_inputs_type`
 pub fn maybe_write<'a, F>(
     log_proof_inputs_type: LogProofInputsType,
+    program: Program,
     claim: &Claim,
     nondeterminism: F,
 ) -> Result<Option<PathBuf>, LogVmStateError>
@@ -92,20 +94,17 @@ where
     };
     let prefix = log_proof_inputs_type.file_prefix();
 
-    write(&dir, prefix, claim, nondeterminism()).inspect_err(|e| tracing::warn!("{}", e))
+    write(&dir, prefix, program, claim, nondeterminism()).inspect_err(|e| tracing::warn!("{}", e))
 }
 
 fn write(
     dir: &str,
     file_prefix: &str,
+    program: Program,
     claim: &Claim,
     nondeterminism: NonDeterminism,
 ) -> Result<Option<PathBuf>, LogVmStateError> {
-    let vm_state = VMState::new(
-        SingleProof.program(),
-        claim.input.clone().into(),
-        nondeterminism,
-    );
+    let vm_state = VMState::new(program, claim.input.clone().into(), nondeterminism);
 
     let filename = format!(
         "{}.{}.{}.json",


src/models/state/vm_proving_capability.rs
```diff
@@ -0,0 +1,265 @@
...
+    /// The program is executed in blocking fashion so it will block concurrent
+    /// async tasks on the same thread.  async callers should use
+    /// `check_if_capable_async()` instead.
+    ///
+    /// #### environment variable: LOG2_PADDED_HEIGHT_METHOD
+    ///
+    /// By default the log2(padded-height) is calculated using VmState::run().
+    ///
+    /// A more accurate but slower way is to use VM::trace_execution_of_state().
+    /// This is typically about 4x slower.
```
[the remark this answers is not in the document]
dan-da: the run() method seems "good enough" for now as the default and the env var provides a simple way to use either so I don't see any urgency.  Anyway I would defer to you on such matters and prioritization as I don't  understand the difference between the two methodologies anyway.

src/models/state/vm_proving_capability.rs
```diff
@@ -0,0 +1,265 @@
...
+        nondeterminism: NonDeterminism,
+    ) -> Result<(), VmProvingCapabilityError> {
+        let log2_padded_height = Self::obtain_log2_padded_height(program, claim, nondeterminism)?;
+        self.can_prove(log2_padded_height)
+    }
+
+    /// executes the supplied program triple to obtain the log2(padded_height)
+    ///
+    /// perf: this is an expensive operation; it may be under a second up to
+    /// 5+ seconds depending on the program's complexity.
```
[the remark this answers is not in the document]
dan-da: yeah, when running unit tests I've typically seen run() take somewhere around a second and trace_execution_of_state() was taking 4-5 seconds.  Of course this is going to be highly variable depending on the program, cpu, env-var setting, etc.  So I was being kind of intentionally vague and hand-wavy while trying to give at least *some* estimate, since 3 seconds is quite different from 30 minutes, for example.   Since you say you've seen a minute-plus, I will use that instead.

src/models/state/vm_proving_capability.rs
```diff
@@ -0,0 +1,265 @@
...
+            }
+
+            // this is fastest, as it avoids running program at all.
+            // but only supported for unit tests, as a "turbo" mode.
+            #[cfg(test)]
+            "skip" => Ok(0),
+
+            "run" | &_ => {
+                // this is baseline
+                vmstate.run().unwrap();
```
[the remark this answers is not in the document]
dan-da: yeah, I don't want to ignore the error, just for correctness.  In the context of VmProvingCapability only (ignoring future proving) it seems it would give a wrong answer since the program could crash early on, and thus the cycle_count would be artificially low.

Unfortunately, the suggestion doesn't work simply:

```
error[E0277]: `?` couldn't convert the error to `VmProvingCapabilityError`
   --> src/models/state/vm_proving_capability.rs:181:30
    |
181 |                 vmstate.run()?;
    |                         -----^ the trait `From<InstructionError>` is not implemented for `VmProvingCapabilityError`
```

So I will probably need to add a #[from] InstructionError variant.

src/models/proof_abstractions/tasm/prover_job.rs
```diff
@@ -157,89 +143,22 @@ impl ProverJob {
...
-        tracing::debug!("executing VM program to determine complexity (padded-height)");
-
-        assert_eq!(self.program.hash(), self.claim.program_digest);
-
-        let mut vm_state = VMState::new(
-            self.program.clone(),
-            self.claim.input.clone().into(),
-            self.nondeterminism.clone(),
-        );
-        maybe_write_debuggable_vm_state_to_disk(&vm_state);
```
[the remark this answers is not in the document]
dan-da: will do.  I need to review exactly what it is doing.  maybe it belongs in vm_proving_capability.rs.  Also I should note that it is now planned to make this fn's logging behavior configurable via an env variable.

src/models/state/vm_proving_capability.rs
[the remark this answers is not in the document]
jan-ferdinand: Ah, it's because you're doing `state.run()`, not `VM::run()`. I see why you do (or inherited) this: `VM::run()` does not return the `VMState`, and I think that's the correct API.

A `VMError` is a

```rust
pub struct VMError {
    /// The reason Triton VM crashed.
    pub source: InstructionError,

    /// The state of Triton VM at the time of the crash.
    pub vm_state: Box<VMState>,
}
```

so you have all the ingredients to build the `VMError` right in the commented line, if you want to do that instead:

```rust
if let Err(err) = vmstate.run() {
    return Err(tasm_lib::triton_vm::error::VMError::new(err, vmstate).into());
}
```

Alternatively, you can downgrade the debug-capabilities that come with `VmProvingCapabilityError` and modify the `VmExecutionFailed` to be `#[from] InstructionError` instead of `#[from] VMError`. Then, the “trace” match-arm could contain:

```rust
    let (aet, _) = VM::trace_execution_of_state(vmstate).map_err(|e| e.source)?;
}
```

Which path to take depends on the answer to the question: how useful is it to have the state of Triton VM at the time of execution failure (for debugging purposes)?

src/models/state/vm_proving_capability.rs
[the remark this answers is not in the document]
jan-ferdinand: I don't think I've _seen_ minute-plus on Neptune core, but I wouldn't be surprised if a proof-collection-to-single-proof execution with many input utxos reaches that.

src/config_models/cli_args.rs
```diff
@@ -520,57 +506,25 @@ impl Args {
...
     /// Cache the result so we don't estimate more than once.
-    pub fn proving_capability(&self) -> TxProvingCapability {
-        *self.tx_proving_capability_cache.get_or_init(|| {
-            if let Some(proving_capability) = self.tx_proving_capability {
+    pub fn proving_capability(&self) -> VmProvingCapability {
+        *self.vm_proving_capability_cache.get_or_init(|| {
+            if let Some(proving_capability) = self.vm_proving_capability {
                 proving_capability
-            } else if self.compose {
-                TxProvingCapability::SingleProof
```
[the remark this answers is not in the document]
jan-ferdinand: Makes perfect sense. Now I also know where I stand on this. 😃

src/models/proof_abstractions/tasm/prover_job.rs
```diff
@@ -157,89 +143,22 @@ impl ProverJob {
...
-        if !capability.can_prove(proof_type) {
-            return Err(ProverJobError::TooWeak {
-                capability,
-                proof_type,
-            });
-        }
-
-        tracing::debug!("executing VM program to determine complexity (padded-height)");
-
-        assert_eq!(self.program.hash(), self.claim.program_digest);
```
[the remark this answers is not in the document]
jan-ferdinand: > what does triton_vm do if these don't match?

There are multiple readings of that question:

1. In this function, in the subsequent call to `run()` (or `trace_execution()`), Triton VM doesn't get to see the full `claim`, only `claim.input`. Ergo, this check can only be done here. A similar statement is true for the similar checks below.
2. In the one convenience function[^fn] where Triton VM gets access to both a claim and a (program, non-determinism) tuple, it checks that the claim is consistent with the program and the program's output given the input (taken from the claim) and the non-determinism. (It doesn't panic on mismatch, of course.)
3. If there is a mismatch during proof verification, verification will fail. (Here, there is no proof that could be verified.)

[^fn]: `triton_vm::prove`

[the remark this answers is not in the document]
jan-ferdinand: >  But in the future I imagine there may be more complex scenarios with user generated programs (smart contracts) running

Maybe? I'm not sure if these asserts are sanity checks and the corresponding claims were generated locally. (@Sword-Smith) If they are only sanity checks, I'd expect a `debug_assert` to communicate that, so I'm assuming they're doing something “real”.

src/models/state/vm_proving_capability.rs
```diff
@@ -0,0 +1,265 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.log2_padded_height)
+    }
+}
+
+impl FromStr for VmProvingCapability {
+    type Err = clap::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.parse::<u8>() {
```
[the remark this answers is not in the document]
dan-da: > Would it be an interesting idea to introduce one additional level of indirection for the CLI parsing, using a type with a FromStr that first tries to parse as TransactionProofType and then as VmProvingCapability, ultimately returning a VmProvingCapability

It's certainly an interesting idea.   Ultimately though I think I don't want to go down that route because then we are just re-introducing the conflation between TransactionProofType and VmProvingCapability at the user-interface level, ie `--vm_proving_capability=single_proof` conflates the two concepts.    I think it will be simpler and cleaner to just document in the --vm_proving_capability help text the "score" for each TransactionProofType variant as a reference, eg:

```
Recommended values by transaction-proof type:
  primitive-witness: 0
  proof-collection: 15
  single-proof: 22
```
The scores would be obtained via TransactionProofType::log2_padded_height() in case they ever change.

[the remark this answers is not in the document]
jan-ferdinand: That's a fair point. :+1:

src/models/state/vm_proving_capability.rs
```diff
@@ -0,0 +1,265 @@
...
+    ///
+    /// ```
+    /// use neptune_cash::api::export::VmProvingCapability;
+    /// use neptune_cash::api::export::TransactionProofType;
+    ///
+    /// let capability: VmProvingCapability = 16.into();
+    ///
+    /// assert!(capability.can_prove(15u32).is_ok());
+    /// assert!(capability.can_prove(16u32).is_ok());
+    /// assert!(!capability.can_prove(17u32).is_ok());
```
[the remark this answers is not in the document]
dan-da: used suggestion in 0717e8abc5d49a294bfc57469646c650bab9a5dd

src/models/state/vm_proving_capability.rs
```diff
@@ -0,0 +1,265 @@
...
+        crate::macros::log_scope_duration!(crate::macros::fn_name!());
+
+        let mut vmstate = VMState::new(program, claim.input.into(), nondeterminism);
+
+        let method =
+            std::env::var("LOG2_PADDED_HEIGHT_METHOD").unwrap_or_else(|_| "run".to_string());
+
+        match method.as_str() {
+            "trace" => {
+                // this is about 4x slower.
```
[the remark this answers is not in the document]
dan-da: used suggestion in 0717e8abc5d49a294bfc57469646c650bab9a5dd

src/models/state/vm_proving_capability.rs
```diff
@@ -0,0 +1,265 @@
...
+
+/// represents proving capability of a device.
+///
+/// The proving capability is represented as log2(padded_height) where
+/// padded_height is an indicator of the execution complexity of the
+/// (program, claim, nondeterminism) triple necessary for generating a
+/// TritonVm `Proof`.
+///
+/// A rough indicator of a device's capability can be obtained via
+/// the `auto_detect()` method.
```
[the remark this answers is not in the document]
dan-da: changed this portion of comment to "//" in 0717e8abc5d49a294bfc57469646c650bab9a5dd.   making auto_detect() pub would be problematic because it should probably be async for future-proofing, which poses problems for cli_args::Arg.  Much simpler just to keep it private for now.

src/models/state/vm_proving_capability.rs
```diff
@@ -0,0 +1,265 @@
...
+            Err(e) => Err(clap::Error::raw(
+                clap::error::ErrorKind::ValueValidation,
+                format!("invalid log2_padded_height '{}': {}", s, e),
+            )),
+        }
+    }
+}
+
+#[derive(Debug, Clone, thiserror::Error)]
+pub enum VmProvingCapabilityError {
```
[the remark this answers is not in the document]
dan-da: changed to non_exhaustive in 0717e8abc5d49a294bfc57469646c650bab9a5dd

src/models/proof_abstractions/tasm/prover_job.rs
```diff
@@ -21,32 +20,23 @@ use crate::job_queue::channels::JobCancelReceiver;
...
 use crate::models::proof_abstractions::NonDeterminism;
 use crate::models::proof_abstractions::Program;
-use crate::models::state::tx_proving_capability::TxProvingCapability;
-use crate::triton_vm::vm::VMState;
+use crate::models::state::vm_proving_capability::VmProvingCapability;
+use crate::models::state::vm_proving_capability::VmProvingCapabilityError;
 
 /// represents an error running a [ProverJob]
 #[derive(Debug, thiserror::Error)]
 pub enum ProverJobError {
```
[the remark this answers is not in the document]
dan-da: done in 0717e8abc5d49a294bfc57469646c650bab9a5dd

src/config_models/cli_args.rs
```diff
@@ -394,6 +383,32 @@ pub struct Args {
     pub(crate) scan_keys: Option<usize>,
 }
 
+pub const VM_PROVING_CAPABILITY_HELP: &str = const_format::formatcp!(
```
jan-ferdinand: Is `pub` necessary for clap?

src/util_types/log_vm_state.rs
```diff
@@ -0,0 +1,145 @@
...
+/// that the claims expose secrets and should not be shared. This mode of
+/// logging is considered a security risk, but can be useful for investigating
+/// or researching alone, or on testnet(s), etc.
+///
+/// The `NoWitness` mode is useful when logging proof inputs for purposes of
+/// sharing with others, eg neptune-core developers for debugging.  However it
+/// does not log any Proofs generated from a `PrimitiveWitness`.
+pub enum LogProofInputsType {
+    /// log any inputs of any kind of proof, possibly leaking witness secrets
+    Proof,
```
jan-ferdinand: `All` or `Any` instead of `Proof`?

src/util_types/log_vm_state.rs
```diff
@@ -0,0 +1,145 @@
...
+            Self::Proof => "NEPTUNE_VM_STATE_DIR",
+            Self::NoWitness => "NEPTUNE_VM_STATE_NO_WITNESS_DIR",
+        }
+    }
+
+    /// returns file name prefix
+    pub const fn file_prefix(&self) -> &str {
+        match *self {
+            Self::Proof => "vm_state.claim",
+            Self::NoWitness => "vm_state.no_witness_claim",
```
jan-ferdinand: I like the `no_witness_claim` to indicate also in the filename that the claim does not contain a primitive witness. I'm less sure about `vm_state.claim`: the state file does not only contain the claim but also the non-determinism and the proof (neither of which is part of the claim). Maybe just drop the `.claim` suffix of the prefix?

src/util_types/log_vm_state.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/// If the environment variable specified by [LogProofInputsType::env_var_name()] is set,
+/// write the initial VM state to file `<DIR>/<prefix>.<pid>.<claim>.json`.
+///
+/// where:
+///  DIR = value of environment variable.
+///  prefix = LogProofInputsType::file_prefix()
+///  pid = process-id
+///  claim = hex-encoded hash of input Claim
```
jan-ferdinand, line 63: Could it make sense to use a timestamp here? That way, it's easier to identify which of potentially many programs was the last one before, say, a crash.

src/util_types/log_vm_state.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+fn write(
+    dir: &str,
+    file_prefix: &str,
+    claim: &Claim,
+    nondeterminism: NonDeterminism,
+) -> Result<Option<PathBuf>, LogVmStateError> {
+    let vm_state = VMState::new(
+        SingleProof.program(),
```
jan-ferdinand: This surprises me. As far as I know, there are multiple programs that could be of interest, like the `Merge` Program or the (set of?) programs that are used to go from `PrimitiveWitness` to `ProofCollection`. Is that generalization left as a future task?

dan-da: thx for catching this!  it surprises me too, heh.

I started out with a single method called from the ProverJob, which handles any type of proof.   But that didn't satisfy the constraint of "avoid logging witness secrets" as the ProverJob is unable to make such a distinction.   So then I changed it to be called by TransactionProofBuilder which does have that information, and that only applies to SingleProofs so I eliminated the program arg.   But I noticed the majority of proofs do not get logged and decided it would be useful for ourselves and "power users" to be able to log from both places independently, even if secrets are logged from ProverJob.  In refactoring for that, it seems I left it as always writing the SingleProof program.  will fix!

src/config_models/cli_args.rs
dan-da: I don't think so. will remove if possible.

src/util_types/log_vm_state.rs
dan-da: could be.  I definitely considered those.  will review.

src/util_types/log_vm_state.rs
dan-da: > the state file does not only contain the claim but also the non-determinism **and the proof**

I'm confused about how the file can contain the proof.

My understanding has been that VMState::run() just runs the program and does not generate a proof.  As such it can run in seconds while the real proof generation with triton_vm::prove() takes sometimes minutes for the same program, claim/input, nondeterminism triple.

Can you clarify about that?

dan-da: also, I noticed that the proof caching mechanism used for neptune-core testing identifies the cached data by the hash of the claim.   Here is the proof_filename() fn from program.rs.

```
    fn proof_filename(claim: &Claim) -> String {
        let base_name = Hash::hash(claim).to_hex();

        format!("{base_name}.proof")
    }
```

So I was essentially working from this precedent (which I did not author).

It does seem logical to me that hash(program, claim, nondeterminism) would be a better identifier, but from above the claim seems to have been deemed sufficient.

src/util_types/log_vm_state.rs
dan-da: umm... maybe?    The filesystem will of course have a timestamp, and `ls -lt` will already sort by it and display it.

Another idea I considered was to create a sub-directory for each PID.  still on the fence if that's a good idea or not, so haven't done it.

dan-da:
> In case you want to land the tx-proving-capability PR without spending time discussing logging things, I think the logging could be a separate PR. If you don't feel like such a discussion is holding things back, I'm happy to progress as is.

I think it makes sense to continue in this PR since this PR removed the original logging (which you flagged in review).   So if we were to merge the PR without it, then we have a kind of regression in master until a subsequent PR is merged.  Also too, I think it will not take much effort to address your latest comments.

src/util_types/log_vm_state.rs
dan-da: I'm leaning towards this now.  what do you think?

```rust
pub enum LogProofInputsType {
    /// log proof inputs that may leak witness secrets
    MaybeLeaksWitnessSecrets,

    /// log proof inputs that do not leak witness secrets
    DoesNotLeakWitnessSecrets,
}
```

src/util_types/log_vm_state.rs
dan-da:
```
Self::Proof => "vm_state.claim",
Self::NoWitness => "vm_state.no_witness_claim",
```

I also was considering that perhaps the `Self::Proof` case should somehow warn not to share, eg: "vm_state.do_not_share" or "vm_state.unsafe_to_share".   what do you think?

jan-ferdinand: > I'm confused about how the file can contain the proof.

And rightfully so. I was trying to say that the VM state also contains the non-determinism and the program (not proof).

> My understanding has been […]. Can you clarify about that?

What you describe here is accurate.

> also, I noticed that the proof caching mechanism used for neptune-core testing identifies the cached data by the hash of the claim.

I concur that a hash of the claim is sufficient for identifying a proof.[^why] I think I left the comment more because I think that the filename `vm_state.claim` suggests that the file contains _only_ the claim, while in truth, it's the entire VM state.

> I also was considering that perhaps the Self::Proof case should somehow warn not to share

I like that idea and your suggestion. How about commenting on the share-safety in both cases?

```rust
Self::Proof => "vm_state.do_not_share",
Self::NoWitness => "vm_state.safe_to_share",
```


[^why]: In case you're interested in the “why”: A claim contains the hash of the program, so the program is pinned down that way. The exact non-determinism is irrelevant for the proofs validity. In general, there can be many different non-determinisms with which it's possible to generate a proof. The point is that the non-determinism doesn't have to be pinned down.

src/util_types/log_vm_state.rs
jan-ferdinand: Hm. I feel like the bikes could do with a bigger shed. Here we go.[^tongue]

It requires a somewhat intimate understanding of Neptune's inner workings to know that (a) “witness” is a “primitive witness” and (b) a primitive witness contains secret key material of the wallet. How about `IncludeSecrets` & `ExcludeSecrets`? `MayLeakSecrets` & `SecureSecrets`? Still doesn't click completely for me…

[^tongue]: That's tongue-in-check. 🙂 I do think that good names are important and worth spending time on.

src/util_types/log_vm_state.rs
dan-da: > vm_state.claim suggests that the file contains only the claim, while in truth, it's the entire VM state.

The problem with abbreviating things is it can have different readings.  I was intending it to mean "vm_state for claim X" where X is included in the filename.

In terms of symmetry, I like this better/best:

```
Self::Proof => "vm_state.unsafe_to_share",
Self::NoWitness => "vm_state.safe_to_share"
```

I'm hesitant to label it "safe_to_share" though.  Seems like we are then making some kind of guarantee.  If there's consensus amongst team members it is "safe" and we should name it that I'm ok with it, I just don't want to make the claim on my own.

src/util_types/log_vm_state.rs
dan-da: I think the part that is still a little fuzzy or "odd" for me is that in all cases we are dealing with a Claim, or more specifically claim inputs.  So when a claim is generated from a primitive-witness we consider these inputs "secrets" but when generated from another proof we consider them "not secrets".  Presumably they are "not secrets" because (a) they cannot be reversed, and (b) they cannot be used as keys to authorize spending of  other UTXOs.

So I "get" that, but I think my mind does not like that it is not reflected in the type system at all, not even as a method on Claim.   So then we need this extra type to differentiate somehow.  An alternative would be to store the information inside Claim somehow when it is created.   Though that would be a much bigger change than deciding on a name for these variants and surely outside scope of this PR.

ok, just focusing on names, how about:

```
pub enum LogProofInputsType {
    /// log proof inputs that may contain wallet secrets
    MayContainWalletSecrets,

    /// log proof inputs that do not contain wallet secrets
    DoesNotContainWalletSecrets,
}
```

also, is there a better name than `LogProofInputsType`?   I used the term "ProofInputs" to encapsulate Claim, Program, Nondeterminism triple, but am open to suggestions.

src/util_types/log_vm_state.rs
jan-ferdinand: Let's hear from people with more knowledge of the architecture whether these VM states are actually safe to share. :+1: (@aszepieniec, @Sword-Smith)

src/util_types/log_vm_state.rs
jan-ferdinand: The problem is not the claim but the non-determinism. Another name for the non-determinism could be “secret input”[^secret] which means that Triton VM takes that thing as input but reveals no information about it.[^zk]
As far as I understand things, some part of the primitive witness gets used as (public) input, which is part of the claim, and other parts are used as (secret input aka) non-determinism. In general, the non-determinism must not be shared, and the public input _must_ be shared for proof validation to stand a chance. We have domain-specific knowledge that in the case of `SingleProof` and `Merge` (maybe others), the non-determinism _can_ be shared without compromising secrets. So the distinction is encoded in the type system, but for convenience and by using additional knowledge, this encoding is ignored in some cases. Maybe this could be encoded harder using additional types. I concur with that being outside the scope of the PR. :relieved:

[^secret]: The reason it's not called “secret input” (anymore) is that non-determinism contains more stuff than “just” secret input. For example, the non-determinism can initialize Triton VM's RAM to whatever it chooses. As a mental concept, secret input is a good approximation.
[^zk]: That's part of the zero-knowledge guarantee that the proof system is all about.

`MayContainWalletSecrets` & `DoesNotContainWalletSecrets` sounds good to me.

> is there a better name than `LogProofInputsType`?

I've struggled with naming that triple in the past. Concretely, in the various test modules of Triton VM, I call it `ProgramAndInput` which, having an “and” in the type name, raises a red flag. `ProofInputs` is not wrong, since those are the inputs to `.prove()`. A noun I'm more used to seeing after “log” than “type” is “level”, but I'm not sure that goes well with the suggested variant names. `VmStateLogLevel`? `ProverInputLogType`? `LogLevelOfProvingInputs`? `LoggingTypeOfTritonVMProverInputTriple`?

src/config_models/cli_args.rs
dan-da: removed `pub` in 20705ddf9963fad63c5a3719469b131195fa733b

src/util_types/log_vm_state.rs
dan-da: renamed variants to `MayContainWalletSecrets` and `DoesNotContainWalletSecrets`  in 20705ddf9963fad63c5a3719469b131195fa733b

src/util_types/log_vm_state.rs
dan-da: 20705ddf9963fad63c5a3719469b131195fa733b changes it to:

```
            Self::MayContainWalletSecrets => "vm_state.unsafe_to_share",
            Self::DoesNotContainWalletSecrets => "vm_state.safe_to_share",
```

src/util_types/log_vm_state.rs
dan-da: I decided against adding another field as the filename is already quite long.

src/util_types/log_vm_state.rs
dan-da: fixed in 20705ddf9963fad63c5a3719469b131195fa733b

dan-da:
ok, we have one approving review.   @aszepieniec or @Sword-Smith, can one of you please review and merge if you agree.   thanks!    (In particular, @aszepieniec you might want to check it since this PR came out of our lengthy discussions.)

aszepieniec:
ACK. Will do.

src/models/state/vm_proving_capability.rs
```diff
@@ -0,0 +1,281 @@
...
+    /// assert!(capability.can_prove(15u32).is_ok());
+    /// assert!(capability.can_prove(16u32).is_ok());
+    /// assert!(capability.can_prove(17u32).is_err());
+    ///
+    /// assert!(capability.can_prove(TransactionProofType::PrimitiveWitness).is_ok());
+    /// assert!(capability.can_prove(TransactionProofType::ProofCollection).is_ok());
+    /// assert!(capability.can_prove(TransactionProofType::SingleProof).is_err());
+    ///
+    /// let single_proof_capability: VmProvingCapability = TransactionProofType::SingleProof.into();
+    /// assert!(single_proof_capability.can_prove(TransactionProofType::SingleProof).is_ok());
```
aszepieniec, line 70:
```suggestion
    /// assert!(single_proof_capability.can_prove(TransactionProofType::SingleProof).is_ok());
    /// ```
```

src/models/state/vm_proving_capability.rs
```diff
@@ -0,0 +1,281 @@
...
+    pub fn can_prove(&self, other: impl Into<u32>) -> Result<(), VmProvingCapabilityError> {
```
aszepieniec, line 71: This function signature is confusing. Syntactically, the type of `other` is, if I read correctly, anything that can be cast via `.into()` into a `u32`, which includes `i8`, `u16`, `i16`, `u32`. But semantically, I am not sure if any of those options make sense. And then the docstring suggests that you can pass `TransactionProofType` variants, which is unexpected because it is not obvious how they map to `u32`s -- or more pertinently, to whatever the benchmark for comparing proving task complexity is. Lastly, why would `u32` even be the right primitive type to capture that benchmark? I am reading this with the context that `VmProvingCapability` is a wrapper around `u8` which certainly suffices to capture all $\log_2$-padded-heights that we are ever going to need.

I would suggest:
 - Drop the genericity in favor of explicit function names, such as `can_prove_log2_padded_height(log2_padded_height)` and `can_prove_transaction(transaction_proof_type)`.
 - Drop the `Into` or `From` conversions from `TransactionProofType` to `u32`s.

src/models/state/vm_proving_capability.rs
```diff
@@ -0,0 +1,281 @@
...
+    fn from(capability: VmProvingCapability) -> Self {
+        capability.log2_padded_height
+    }
+}
+
+impl From<VmProvingCapability> for u32 {
+    fn from(capability: VmProvingCapability) -> Self {
+        capability.log2_padded_height.into()
+    }
+}
```
aszepieniec, line 47: I do not like these `From` implementations. There is no natural mapping from `u8` to VM proving capability or vice versa. The fact that `VmProvingCapability` is a wrapper around `u8` is an implementation detail which these `From` fail to make abstraction of. They only make sense if you already know that `VmProvingCapability` is a wrapper around `u8` (thus raising the bar for newcomers), and even then only as long as that is true. In the future we might expand `VmProvingCapability` in ways that do not change the semantic captured by the struct, which is the information necessary to determine whether you can produce a particular proof. For instance, we might add available parallelism or available AVX instructions or GPU acceleration, etc. In such an event the `From` conversions certainly make no sense, but will leave the person applying the refactor scratching their head about what to replace them with.

I favor more explicit conversion functions:
 - `from_log2_padded_height`
 - `log2_padded_height` (accessor).

src/models/state/vm_proving_capability.rs
```diff
@@ -0,0 +1,281 @@
...
+            TransactionProofType::SingleProof.into()
+        } else if total_memory > PROOF_COLLECTION_MEMORY_USAGE
+            && physical_core_count > PROOF_COLLECTION_CORE_REQ
+        {
+            TransactionProofType::ProofCollection.into()
+        } else {
+            TransactionProofType::PrimitiveWitness.into()
+        }
+    }
+}
```
aszepieniec, line 249: The logic in here seems rather roundabout to me. We have a number capturing the total available RAM. We compare it against memory required for {ProofCollection, SingleProof}, and map the correct one of these into a $\log_2$-padded-height ... which represents the biggest table that, accounting for arithmetic, can fit in RAM.

Why not go straight from RAM to RAM? Or more specifically, why not go straight from total RAM available to max $\log_2$-padded-height and bypass {ProofCollection, SingleProof}?

src/api/tx_initiation/builder/transaction_proof_builder.rs
```diff
@@ -241,16 +252,19 @@ impl<'a> TransactionProofBuilder<'a> {
             job_queue,
             proof_job_options,
             valid_mock,
+            transaction_proof_type,
         } = self;
 
         let proof_job_options = proof_job_options.ok_or(ProofRequirement::ProofJobOptions)?;
+        let transaction_proof_type = transaction_proof_type
+            .unwrap_or_else(|| proof_job_options.job_settings.vm_proving_capability.into());
```
aszepieniec, line 260: This `.into()` hides *a lot* of logic. If I understand correctly, it is mapping a `VmProvingCapability` to a `TransactionProofType` by selecting the `TransactionProofType` whose corresponding `VmProvingCapability` is the largest subject to being smaller than the given `vm_proving_capability`. A mapping as complex and non-trivial as that warrants an explicit, descriptive name. For instance: `most_complex_transaction_proof_possible`.

In contrast, `.into()` should be reserved for cases where target type and concrete mapping are so obvious that the extra explicit indication of a dedicated function can be omitted.

scripts/php/find_slow_tests.php
```diff
@@ -0,0 +1,99 @@
...
+    echo "--- unit test stats ---\n";
+    echo "Count: " . $count . "\n";
+    echo "Total: " . $sum . "\n";
+    echo "Min:   " . $min . "\n";
+    echo "Max:   " . $max . "\n";
+    echo "Mean:  " . $mean . "\n";
+    echo "Median: " . $median . "\n";
+    echo "----------------------\n";
+}
+
```
aszepieniec, line 99: The last time I tried to install PHP to run one of the scripts in this repo, I ended up corrupting my operating system and I had to reinstall ubuntu from scratch. (It was certainly not the fault of the script I was trying to run because I had not even gotten to that point.) I am not tempted to try again to see what went wrong the first time.

This is not a vendetta against PHP. I actually think it is a fantastic language and great at what it does. But I do want to raise the question whether it is a good idea to include PHP scripts in this repository. They obviously come with a huge dependency even if the fallout of installing that dependency does not include corrupting the operating system.

Possible alternatives:
 - bash
 - rust
 - python
 - nothing

If the issue is, "who is going to write the bash/rust/python equivalent?" then I wonder how far we get with asking AI to transcribe it.

If the issue is, "who is going to maintain it?" then PHP is the most ill-suited option for random new developers. (Okay, scratch that. Bash is. But PHP is second. The point remains that rust and python are both more accessible and more popular.)

src/main_loop/proof_upgrader.rs
```diff
@@ -142,23 +140,22 @@ impl UpgradeJob {
...
+            TransactionProofType::PrimitiveWitness if network.use_mock_proof() => {
                 UpgradeJob::PrimitiveWitnessToSingleProof { primitive_witness }
             }
-            TxProvingCapability::PrimitiveWitness => {
+            TransactionProofType::PrimitiveWitness => {
                 panic!("Client cannot have primitive witness capability only")
             }
-            TxProvingCapability::LockScript => todo!("TODO: Add support for this"),
         }
     }
```
aszepieniec, line 160: Why not take a target `TransactionProofType` directly? Allowing anything that can be cast `into` one seems rather indirect and error-prone.

src/models/blockchain/transaction/transaction_proof.rs
```diff
@@ -37,21 +49,64 @@ impl From<&TransactionProof> for TransactionProofType {
...
+
+        if max >= TransactionProofType::SingleProof.log2_padded_height() {
+            TransactionProofType::SingleProof
+        } else if max >= TransactionProofType::ProofCollection.log2_padded_height() {
+            TransactionProofType::ProofCollection
+        } else {
+            TransactionProofType::PrimitiveWitness
+        }
+    }
+}
```
aszepieniec, line 64: This mapping is complex and non-trivial, and thus poorly suited for a `From`-implementation. I would suggest using a dedicated function instead, with a descriptive name, perhaps something like `most_complex_possible_transaction_proof`.

src/models/blockchain/transaction/transaction_proof.rs
```diff
@@ -37,21 +49,64 @@ impl From<&TransactionProof> for TransactionProofType {
...
+    fn from(proof_type: TransactionProofType) -> Self {
+        proof_type.log2_padded_height()
+    }
+}
+
+impl From<TransactionProofType> for u32 {
+    fn from(proof_type: TransactionProofType) -> Self {
+        proof_type.log2_padded_height().into()
+    }
+}
```
aszepieniec, line 82: Mapping `TransactionProofType`s to `u8`s or `u32`s seems like it is exposing the implementation detail that "proving capability" is represented as a $\log_2$-padded-height. These conversion methods only make sense if you take that context into account. To readers unfamiliar with that context, they are confusing.

I would suggest dropping these `From`-implementations and relying instead on `From<TransactionProofType> for VmProvingCapability`, which I have no objection to.

src/models/blockchain/transaction/transaction_proof.rs
```diff
@@ -37,21 +49,64 @@ impl From<&TransactionProof> for TransactionProofType {
...
 
-    pub fn is_single_proof(&self) -> bool {
-        *self == TransactionProofType::SingleProof
+    /// provides an estimate of padded-height complexity for each variant.
+    ///
+    /// these values were determined by running unit tests
+    /// and logging padded-height values in the ProverJob.
+    ///
+    /// They might need to be adjusted in the future.
+    pub(crate) const fn log2_padded_height(&self) -> u8 {
```
aszepieniec, line 96: Is it possible to change the function name to reflect the fact that it returns an estimate or heuristic? Right now it reads like an accessor returning either a field verbatim or a straightforward function of the encapsulated data.

(The feature that disqualifies this function from being straightforward is that its accuracy is up in the air.)

src/models/state/mempool.rs
```diff
@@ -644,9 +645,11 @@ impl Mempool {
...
         new_block: &Block,
         predecessor_block: &Block,
-        tx_proving_capability: TxProvingCapability,
+        vm_proving_capability: impl Into<VmProvingCapability>,
         composing: bool,
     ) -> (Vec<MempoolEvent>, Vec<UpdateMutatorSetDataJob>) {
+        let vm_proving_capability = vm_proving_capability.into();
+
         // If the mempool is empty, there is nothing to do.
         if self.is_empty() {
```
aszepieniec, line 654: Allowing anything that can be cast `into` a `VmProvingCapability` as opposed to a `VmProvingCapability` directly, is more permissive and also more error-prone. (Why) do we need the permissiveness?

src/models/state/mempool.rs
```diff
@@ -1637,9 +1642,9 @@ mod tests {
         let network = Network::Main;
         let alice_wallet = WalletEntropy::devnet_wallet();
         let alice_key = alice_wallet.nth_generation_spending_key_for_tests(0);
-        let proving_capability = TxProvingCapability::SingleProof;
+        let proving_capability = TransactionProofType::SingleProof;
```
aszepieniec, line 1645: The variable is called `proving_capability` but its type is `TransactionProofType`. Discrepant.

Please modify the variable name to reflect the data or move the `.into()` up so that the data matches with the variable name.

src/models/state/mempool.rs
```diff
@@ -1999,7 +2004,7 @@ mod tests {
 
         /// Return a valid, deterministic transaction with a specified proof type.
         async fn tx_with_proof_type(
-            proof_type: TxProvingCapability,
+            proof_type: impl Into<TransactionProofType>,
             network: Network,
```
aszepieniec, line 2008: What's the benefit of this genericity? If none, drop.

aszepieniec (changes requested):
Conceptually, I approve of these changes. However, I would like to push back against the improper use (in my eyes) of automatic conversion traits (`From`/`into` specifically), which I think are confusing and error-prone. I left more arguments and suggestions in specific in-line comments. In summary:

 - `TransactionProofType` --> `VmProvingCapability`. This can and should be `From` because it's quite obvious what's going on and quite simple too. So no changes requested for this one.
 - `VmProvingCapability` --> `TransactionProofType`. This mapping does have to be computed in some locations and I think you chose the right concrete map. However, the logic is non-trivial. Please use a dedicated function with a descriptive name.
 - {`TransactionProofType`, `VmProvingCapability`} --> {`u8`, `u32`} These maps are confusing and they only make sense if you rely on the contextual knowledge that in the end the thing being compared is $\log_2$-padded-heights. This reliance raises the bar for newcomers who do not have that context. These maps resist change because they effectively pin down internal implementation details. And lastly, I do not think they are necessary. Please drop them.

Issue #600, opened 2025-05-20
CI coverage check did not complete.  failed to generate report: process didn't exit successfully


dan-da:
I'm uncertain if this is just a transient error or a persistent one.  I will try re-rerunning the job.

-------------

from: https://github.com/Neptune-Crypto/neptune-core/actions/runs/15131978306/job/42534868095?pr=599

     Summary [2839.814s] 678 tests run: 678 passed (27 slow), 6 skipped
[...]

PR #605, opened 2025-05-26
ci: Fix coverage workflow

jan-ferdinand:
Due to a regression (?) in the rustc compiler, the code coverage workflow has been failing since 2025-05-20. Temporarily use a version of nightly that is known to work, until rust-lang/rust#141577 (or whatever replaces it) is resolved.

See also: #600


Sword-Smith:
Cool. It'd be nice to get the green check mark back. No reservations from my side.

PR #606, opened 2025-05-26
docs: document release candidate testnet policy

Sword-Smith:
Alternative PR to #573. Main change relative to #573 is that this describes using main net for testing release candidates, as main net has two benefits over test net:
- guarantees presence of older versions of the node software among peers
- is a competitive environment which test net is likely never to become

None of these changes to the documentation are intended to affect testing on networks using mocked proofs, like `TestnetMock` and `RegTest`. The changes relative to #573 are only intended to favor `Network::Main` over `Network::Testnet` for running release candidate instances.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ then the flag `--port <port>` is your friend.
 
 ## Release Candidates and Public Testnet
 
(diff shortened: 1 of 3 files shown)


Sword-Smith:
Not going forward with this. We'll have to rely on integration tests and unit tests and ad hoc manual testing.

PR #608, opened 2025-06-10
feat: bootstrap flushing period

KaffinPX:


aszepieniec:
Excellent feature.

I'm a little confused. If I read correctly, the current behavior is to flush once at the end of the synchronization. This feature enables nodes to flush the database at regular intervals during the sync. That behavior saves memory. Is that the motivation? I originally read it the other way around, as though the motivation was a *faster* sync due to *fewer* flushes.

Could you add a test that verifies that the CLI argument has the intended effect? Also, I am concerned that the database ends up in a corrupt state if syncing from stored blocks is terminated at some point roughly in the middle between two flushes. Could you also add a test that simulates this behavior?

I realize that "bootstrap-flush-period" is in line with existing CLI arguments "bootstrap-from-directory" and "disable-bootstrap-block-validation" but all three are at odds with existing CLI argument "bootstrap" which means something completely different. Let's start a conversation about what to name the new commands to (or rename the old command to) to avoid confusion.

A suggestion: rename "bootstrap" in the sense of "initial state synchronization" to "initial-sync".

KaffinPX:
> I'm a little confused. If I read correctly, the current behavior is to flush once at the end of the synchronization. This feature enables nodes to flush the database at regular intervals during the sync. That behavior saves memory. Is that the motivation? I originally read it the other way around, as though the motivation was a _faster_ sync due to _fewer_ flushes.

Yes, this will enable users to flush in an interval allowing bootstrapping on devices with lower memory, to get the old behavior this can be set to zero.

> Could you add a test that verifies that the CLI argument has the intended effect? Also, I am concerned that the database ends up in a corrupt state if syncing from stored blocks is terminated at some point roughly in the middle between two flushes. Could you also add a test that simulates this behavior?

You are right about corruption but it is also problematic with current master behavior, has to be "fixed" or improved in another PR as this PR is "simple" and only motivation is synchronizing on lower-end devices.
I'm not sure how I could test "flushes" tbh, have to check later but still I think it should be done in another PR refactoring whole thing.



> I realize that "bootstrap-flush-period" is in line with existing CLI arguments "bootstrap-from-directory" and "disable-bootstrap-block-validation" but all three are at odds with existing CLI argument "bootstrap" which means something completely different. Let's start a conversation about what to name the new commands to (or rename the old command to) to avoid confusion.

We could go with blocks "importing", I don't know but I also would say it deserves its own PR. :P

Issue #597, opened 2025-05-15
Track Status of Own Transaction


aszepieniec:
When you initiate a transaction via the dashboard, you get very little feedback on the status of that transaction. This lack of feedback leads to confusion and can even cause knowledgeable (and handsome) project contributors to err and initiate a *second* transaction in the mistaken belief that the first one was forgotten by the network. That mistake constitutes a loss of funds scenario when both transactions end up confirmed.

There is *some* feedback, namely a) the "transaction broadcast" message after sending on the send screen; and b) the mempool which shows the transaction and the fact that the user is the owner. However, this feedback is inadequate:
 - The "transaction broadcast" message is wrong because at this point the transaction has not been broadcast yet. In fact, the node still has to produce a `ProofCollection` for it. But this message does suggest that it is okay to turn off the node now because the rest of the network knows about the transaction now and will ensure its confirmation.
 - The mempool page shows that a transaction initiated by the user lives in the mempool, but it does not show whether it was broadcast. 
 - When a new block arrives, the mempool is cleared -- or at least, that's what it looks like. What really happens is that upgrade jobs are spawned for all transactions in the mempool and they are not shown until those jobs are complete because.

It is easy to suggest displaying some information without suggesting anything about where that information should come from. I haven't quite wrapped my head around the second part yet. So let me issue a call for discussion and frame it in terms of a suggestion for the easy part, which is *which information should be displayed?*

 - amount
 - recipient (address fingerprint)
 - timestamp of initiation
 - status: recorded / generated {primitive-witness, proof-collection, single-proof} (+ upgrading to {proof-collection, single-proof})? / broadcasted {proof-collection, single-proof} / confirmed in block

This information should probably be shown in a new screen. Or we can modify the "History" screen to show this data instead. That option would address #364 but would also admit scope creep because now you also need to show information about incoming transactions.

PR #609, opened 2025-06-10
feat: utxo_origin_block call

KaffinPX:
Can be used for "checking" status of outgoing UTXO(and txs).
related with #597

### DIFF
--- a/src/rpc_server.rs
+++ b/src/rpc_server.rs
@@ -4335,7 +4335,7 @@ mod tests {
     #[traced_test]
     #[test_strategy::proptest(async = "tokio", cases = 5)]
     async fn utxo_origin_block_test(
-        #[strategy(propcompose_txkernel_with_lengths(0usize, 1usize, 0usize))]
+        #[strategy(txkernel_with_lengths(0usize, 1usize, 0usize))]
         transaction_kernel: crate::models::blockchain::transaction::transaction_kernel::TransactionKernel,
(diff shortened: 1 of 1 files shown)


Sword-Smith:
Very nice. Please add a call to this endpoint from a test somewhere though. Could just be in the test that calls all RPC endpoints.

Sword-Smith:
Two questions:
1. Why not just return the block digest of the block containing the output? Isn't that what the caller is interested in?
2. Can you rebase against `origin/master`? The `pseudorandom_transaction_kernel` got removed. If not, I can just do the rebase and fix the test.

KaffinPX:
> * Why not just return the block digest of the block containing the output? Isn't that what the caller is interested in?

There's 2 potential "main" use-cases for this call, one is counting confirmations and one is checking inclusion and for counting confirmations we will need block data or "height", I can add a variant call called ``utxo_origin_digest`` if wanted, I'm not sure tbh.

> 2\. Can you rebase against origin/master? The pseudorandom_transaction_kernel got removed. If not, I can just do the rebase and fix the test.

Kk, will do after a decision about the return type.

Sword-Smith:
If you add this on top of your test, it works:

```rust
prop_assume!(!transaction_kernel.fee.is_negative());
```

The `set_new_tip` method requires that the block's transaction fee is non-negative. Otherwise, is panics.

PR #610, opened 2025-06-12
Refactor mempool: Preserve PW over block updates

Sword-Smith:
This PR works towards two goals: Improve the likelihood that weak computers have their transactions mined, and move closer to the vision of three-step mining as outlined [here](https://talk.neptune.cash/t/thoughts-on-the-mempool-issue-and-upgrading/52/26).

Three main changes:
1. Refactor mempool to never throw away `PrimitiveWitness` data.
2. Don't automatically run `SingleProof`'s "update" branch when receiving a new block, but let "update" be handled by proof upgraders.
3. Add new task for upgraders: The updating of `SingleProof` backed transactions.

- Since `ProofCollection`-backed transactions cannot be updated to be valid under a new mutator set, we instead update them with their `PrimitiveWitness` if one such is present. A `PrimitiveWitness` *will* be present in the mempool if the transaction was initiated locally. Previously, a `ProofCollection` backed transaction would always throw out the `PrimitiveWitness` but with this change, the `PrimitiveWitness` is always preserved in the mempool, if it was ever present there (i.e. if the transaction was initiated locally).
- Replace `TransactionOrigin` in the mempool with an ordered `UpgradePriority` that can be used by proof upgraders to pick the most favorable transaction for proof upgrading -- this especially helps pick the best transaction for "update".
- Remove execution of the "update" branch from being done automatically by any composer when a new block is received to being a task that only the proof upgrader performs.
- Only locally initiated transactions (as checked by the presence of a primitive witness) are updated immediately after the receival of a new block.

The central part of the code to understand is, in my opinion, this loop in mempool's `update_with_block`, as it defined the rules for what to do with transactions that did not get mined in the new block. Two questions are relevant for each transaction:
1. Should the transaction be updated to be valid under a new mutator set immediately?
2. Should the transaction stay in the mempool?

```rust
        for (tx_id, tx) in &self.tx_dictionary {
            if tx.transaction.kernel.inputs.is_empty() {
                debug!("Not updating transaction since empty transactions cannot be updated.");
                kick_outs.push(*tx_id);
                events.push(MempoolEvent::RemoveTx(tx.transaction.clone()));
                continue;
            }

            let (update_job, keep_in_mempool) = match &tx.transaction.proof {
                // Proof-collection backed transaction cannot be updated, but if
                // the transaction was initiated locally, the primitive witness
                // will be known, and it can be updated.
                TransactionProof::ProofCollection(_) => {
                    if let Some(pw) = &tx.primitive_witness {
                        let pw_update_job = PrimitiveWitnessUpdate::new(pw.to_owned());
                        let pw_update_job = MempoolUpdateJob::ProofCollection(pw_update_job);
                        (Some(pw_update_job), true)
                    } else {
                        (None, false)
                    }
                }

                // Primitive witness-backed transactions can easily be updated.
                TransactionProof::Witness(pw) => {
                    let pw_update_job = PrimitiveWitnessUpdate::new(pw.to_owned());
                    let pw_update_job = MempoolUpdateJob::PrimitiveWitness(pw_update_job);
                    (Some(pw_update_job), true)
                }

                // Single proofs can be updated. So they are kept in the
                // mempool in the expectation that someone will update them to
                // be valid under a new mutator set.
                //
                // If the transaction was initiated locally, *and* node can
                // produce single-proofs, transaction should be updated
                // immediately (and be kept in mempool).
                TransactionProof::SingleProof(sp) => {
                    if self.tx_proving_capability == TxProvingCapability::SingleProof
                        && tx.primitive_witness.is_some()
                    {
                        // Node initiated transaction and can update.
                        let update_sp = MempoolUpdateJob::SingleProof {
                            old_kernel: tx.transaction.kernel.clone(),
                            old_single_proof: sp.to_owned(),
                        };

                        (Some(update_sp), true)
                    } else {
                        (None, true)
                    }
                }
            };

            if let Some(update_job) = update_job {
                update_jobs.push(update_job);
            }

            if !keep_in_mempool {
                kick_outs.push(*tx_id);
                events.push(MempoolEvent::RemoveTx(tx.transaction.clone()));
                if !tx.upgrade_priority.is_irrelevant() {
                    warn!("Unable to update own transaction to new mutator set. You may need to create this transaction again. Removing {tx_id} from mempool.");
                }
            }
        }
```


src/main_loop/proof_upgrader.rs
```diff
@@ -201,25 +243,30 @@ impl UpgradeJob {
...
-                ..
-            } => {
-                let reciprocal = 1.0 / (collection.num_proofs() as f64);
-                gobbling_fee.lossy_f64_fraction_mul(reciprocal)
+            UpgradeJob::PrimitiveWitnessToProofCollection(pw_to_pc) => {
+                UpgradeIncentive::BalanceAffecting(
+                    pw_to_pc
+                        .primitive_witness
+                        .output_utxos
+                        .get_native_currency_amount(),
```
jan-ferdinand: The more we spend, the more important the transaction? Might be worth a comment.

src/main_loop/proof_upgrader.rs
```diff
@@ -518,12 +555,75 @@ impl UpgradeJob {
         (fee_notification_method, receiver_preimage)
     }
 
+    /// Build a single-proof backed gobbler transaction that can be used to
+    /// charge another transaction for upgrading a proof.
+    #[allow(clippy::too_many_arguments)]
```
jan-ferdinand: 1. 😕
2. Please use `expect` over `allow` whenever possible.

src/main_loop/upgrade_incentive.rs
```diff
@@ -0,0 +1,73 @@
+use crate::api::export::NativeCurrencyAmount;
+
+/// Enumerate the incentive to perform a transaction upgrade.
```
jan-ferdinand: While the type is an `enum`, someone using the type doesn't perform an enumeration. Maybe:

```rust
/// The magnitude of the incentive to perform a transaction upgrade.
```

src/main_loop/upgrade_incentive.rs
```diff
@@ -0,0 +1,73 @@
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum UpgradeIncentive {
+    /// The transaction is upgraded because fees can be gobbled.
+    Gobble(NativeCurrencyAmount),
+
+    /// The transaction is upgraded because we'd like to see it go through for
+    /// other reasons that fee-collection. Usually because it affects our
```
jan-ferdinand:
```suggestion
    /// other reasons than fee-collection. Usually because it affects our
```

src/main_loop/upgrade_incentive.rs
```diff
@@ -0,0 +1,73 @@
...
+    /// balance, in either positive or negative direction.
+    BalanceAffecting(NativeCurrencyAmount),
```
jan-ferdinand, line 14: (How) do the first two variants relate to each other?

src/main_loop/upgrade_incentive.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    /// On successful proof-upgrading, the new transaction will affect our
+    /// balance in the case fees were gobbled. Return the upgrade incentive as
+    /// it looks after a successful proof-upgrading.
+    pub(crate) fn after_upgrade(self) -> Self {
+        match self {
+            UpgradeIncentive::Critical => UpgradeIncentive::Critical,
+            UpgradeIncentive::Gobble(native_currency_amount) => {
+                UpgradeIncentive::BalanceAffecting(native_currency_amount)
+            }
```
jan-ferdinand, line 37: I like this.

src/main_loop/upgrade_incentive.rs
```diff
@@ -0,0 +1,73 @@
...
+            UpgradeIncentive::BalanceAffecting(native_currency_amount) => {
+                UpgradeIncentive::BalanceAffecting(native_currency_amount)
+            }
+        }
+    }
+}
+
+impl PartialOrd for UpgradeIncentive {
+    // Implemented such that balance affecting transactions are always
+    // prioritized over the chance to gobble fees.
```
jan-ferdinand: I think this comment belongs on the implementation of `Ord` more than here.

src/main_loop/upgrade_incentive.rs
```diff
@@ -0,0 +1,73 @@
...
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for UpgradeIncentive {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        // self.partial_cmp(other).expect("Ord must be implemented.")
```
jan-ferdinand:
```suggestion
```

src/main_loop/upgrade_incentive.rs
```diff
@@ -0,0 +1,73 @@
...
+                UpgradeIncentive::BalanceAffecting(self_amt),
+                UpgradeIncentive::BalanceAffecting(other_amt),
+            ) => self_amt.cmp(other_amt),
+            (UpgradeIncentive::BalanceAffecting(_), UpgradeIncentive::Critical) => {
+                std::cmp::Ordering::Less
+            }
+            (UpgradeIncentive::Critical, UpgradeIncentive::Critical) => std::cmp::Ordering::Equal,
+            (UpgradeIncentive::Critical, _) => std::cmp::Ordering::Greater,
+        }
+    }
```
jan-ferdinand: I feel like this can be simplified.

```suggestion
    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
        use std::cmp::Ordering;
        use UpgradeIncentive as Incentive;

        match (self, other) {
            (Incentive::Gobble(s), Incentive::Gobble(o))
            | (Incentive::BalanceAffecting(s), Incentive::BalanceAffecting(o)) => s.cmp(o),
            (Incentive::Gobble(_), _) => Ordering::Less,
            (Incentive::BalanceAffecting(_), _) => Ordering::Less,
            (Incentive::Critical, Incentive::Critical) => Ordering::Equal,
            _ => Ordering::Greater,
        }
    }
```

src/models/blockchain/transaction/primitive_witness.rs
```diff
@@ -95,6 +96,15 @@ impl SaltedUtxos {
                 .unwrap(),
         }
     }
+
+    /// The total value contained in the UTXOs, regardless of which other coins
+    /// are present (even if that makes the Neptune coins unspendable.)
```
jan-ferdinand: What does the part in the parenthesis mean? Nothing is mutated here, so how can anything become unspendable?

src/models/blockchain/type_scripts/native_currency_amount.rs
```diff
@@ -103,6 +104,7 @@ impl NativeCurrencyAmount {
     }
 
     /// Return the element that corresponds to 1 nau. Use in tests only.
+    #[cfg(test)]
```
jan-ferdinand, line 107: I suggest to move any `#[cfg(test)]` functions into the corresponding test module. Personally, I find it confusing to find test-only code in the middle of production code.

src/models/state/mempool.rs
```diff
@@ -87,36 +95,19 @@ pub enum MempoolEvent {
...
+#[derive(Debug, GetSize, Clone)]
+#[cfg_attr(test, derive(serde::Serialize))]
+struct MempoolTransaction {
+    transaction: Transaction,
 
-        write!(f, "{output}")
-    }
-}
+    /// The value of a transaction for the node operator.
+    #[get_size(ignore)]
```
jan-ferdinand, line 87: Comment why this is ignored?

src/models/state/mempool.rs
```diff
@@ -227,22 +250,70 @@ impl Mempool {
...
     }
 
-    /// Return the proof collection-supported transaction with highest
-    /// fee-density if mempool contains any such transactions. Otherwise, None.
-    pub(crate) fn most_dense_proof_collection(
+    /// Return the preferred single-proof backed transaction for the "update"
+    /// proof upgrade. Returns a transaction that is not synced to the tip
+    /// such that the caller can make the transaction synced again.
+    ///
+    /// Will always prioritize transactions with a positive upgrade priority.
```
jan-ferdinand: What's a “positive” upgrade priority? Anything but “irrelevant”? Maybe this line means “respects transactions' upgrade priorities”?

I suggest to change this line (and a few comments further down) from future tense to present tense. I find it unusual to see the future tense used here, unless you're making a promise to about future versions of this function.

src/models/state/mempool.rs
```diff
@@ -227,22 +250,70 @@ impl Mempool {
...
+    pub(crate) fn preferred_update(
+        &self,
+    ) -> Option<(&TransactionKernel, &NeptuneProof, UpgradePriority)> {
+        for txid in self
+            .upgrade_priority_iter()
+            .map(|(txid, _)| txid)
+            .chain(self.fee_density_iter().map(|(txid, _)| txid))
+        {
+            let candidate = self.tx_dictionary.get(&txid).unwrap();
```
jan-ferdinand, line 280: If all the rest is `continue`, why not here, too?

```suggestion
            let Some(candidate) = self.tx_dictionary.get(&txid) else { continue }; // maybe warn
```

src/models/state/mempool.rs
```diff
@@ -251,31 +322,72 @@ impl Mempool {
...
-    ) -> Option<([(&TransactionKernel, &Proof); 2], TransactionOrigin)> {
+    ) -> Option<([(TransactionKernel, Proof); 2], UpgradePriority)> {
         let mut ret = vec![];
-        let mut own_tx = false;
-        for (txid, _fee_density) in self.get_sorted_iter() {
+        let mut priority = UpgradePriority::Irrelevant;
+        for txid in self
+            .upgrade_priority_iter()
+            .map(|(txid, _)| txid)
+            .chain(self.fee_density_iter().map(|(txid, _)| txid))
```
jan-ferdinand, line 356: That's the third time I see that pattern. Seems like it's a thing.

src/models/state/mempool.rs
```diff
@@ -251,31 +322,72 @@ impl Mempool {
...
             if ret.len() == 2 {
-                return Some((ret.try_into().unwrap(), origin));
+                break;
             }
         }
 
-        None
+        let Ok(ret) = <Vec<_> as TryInto<[_; 2]>>::try_into(ret) else {
+            return None;
+        };
```
jan-ferdinand:
```suggestion
        let ret: [_; 2] = ret.try_into().ok()?;
```

src/models/state/mempool.rs
```diff
@@ -251,31 +322,72 @@ impl Mempool {
...
+        Some((
+            ret.map(|txid| {
+                let candidate = self.tx_dictionary.get(&txid).unwrap();
+                (
+                    candidate.transaction.kernel.to_owned(),
+                    candidate.transaction.proof.to_owned().into_single_proof(),
+                )
+            }),
+            priority,
+        ))
```
jan-ferdinand: Maybe even:

```suggestion
        let [left, right] = ret.try_into().ok()?;
        let left = &self.tx_dictionary.get(&left)?.transaction;
        let right = &self.tx_dictionary.get(&right)?.transaction;
        let candidates =
            [left, right].map(|t| (t.kernel.to_owned(), t.proof.to_owned().into_single_proof()));

        Some((candidates, priority))
```

src/models/state/mempool.rs
```diff
@@ -541,40 +699,23 @@ impl Mempool {
...
     /// Removes the transaction with the lowest [`FeeDensity`] from the mempool.
     /// Returns the removed value.
     ///
     /// Computes in θ(lg N)
     fn pop_min(&mut self) -> Option<(MempoolEvent, FeeDensity)> {
-        if let Some((transaction_digest, fee_density)) = self.queue.pop_min() {
-            if let Some(tx) = self.tx_dictionary.remove(&transaction_digest) {
-                debug_assert_eq!(self.tx_dictionary.len(), self.queue.len());
+        if let Some((txkid, fee_density)) = self.fee_densities.pop_min() {
+            if let Some(tx) = self.tx_dictionary.remove(&txkid) {
```
jan-ferdinand, line 731: What if the `remove()` returns `None`? Is that a sign that the mempool is in an inconsistent state? Should that be communicated somehow?

```rust
fn pop_min(&mut self) -> Option<(MempoolEvent, FeeDensity)> {
    let (txkid, fee_density) = self.fee_densities.pop_min()?;
    self.upgrade_priorities.remove(&txkid);
    let Some(tx) = self.tx_dictionary.remove(&txkid) else {
        error!("mempool is in inconsistent state");
        return None;
    };
    debug_assert_eq!(self.tx_dictionary.len(), self.fee_densities.len());

    Some((MempoolEvent::RemoveTx(tx.transaction), fee_density))
}
```

src/models/state/mempool.rs
```diff
@@ -700,72 +823,59 @@ impl Mempool {
...
                 kick_outs.push(*tx_id);
                 events.push(MempoolEvent::RemoveTx(tx.transaction.clone()));
                 continue;
             }
 
-            let can_upgrade_single_proof =
-                TxProvingCapability::SingleProof == tx_proving_capability;
-            let (update_job, can_update) = match &tx.transaction.proof {
+            let (update_job, keep_in_mempool) = match &tx.transaction.proof {
+                // Proof-collection backed transaction cannot be updated, but if
```
jan-ferdinand: Maybe hint that it's _possible_ to update a proof-collection backed transaction that does not have an associated primitive witness, but that the current policy is not to do so because it requires a big amount of resources, going through both the raise _and_ update paths?

src/models/state/mempool.rs
```diff
@@ -844,12 +968,21 @@ impl Mempool {
     /// Yields the `transaction_digest` in order of descending `fee_density`, since
     /// users (miner or transaction merger) will likely only care about the most valuable transactions
     /// Computes in O(N lg N)
-    pub fn get_sorted_iter(
+    pub fn fee_density_iter(
         &self,
-    ) -> Rev<IntoSortedIter<TransactionKernelId, FeeDensity, RandomState>> {
-        let dpq_clone = self.queue.clone();
+    ) -> Rev<DoubleEndedIterator<TransactionKernelId, FeeDensity, RandomState>> {
```
jan-ferdinand: For public functions, I suggest to return an intransparent type instead. Depending on the guarantees you want to make, this could be:

```suggestion
    ) -> impl std::iter::DoubleEndedIterator<Item = (TransactionKernelId, FeeDensity)> {
```

src/models/state/mempool/upgrade_priority.rs
```diff
@@ -0,0 +1,96 @@
...
+                std::cmp::Ordering::Greater
+            }
+            (UpgradePriority::Interested(self_amt), UpgradePriority::Interested(other_amt)) => {
+                self_amt.cmp(other_amt)
+            }
+            (UpgradePriority::Interested(_), UpgradePriority::Critical) => std::cmp::Ordering::Less,
+            (UpgradePriority::Critical, UpgradePriority::Critical) => std::cmp::Ordering::Equal,
+            (UpgradePriority::Critical, _) => std::cmp::Ordering::Greater,
+        }
+    }
```
jan-ferdinand:
```suggestion
    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
        use std::cmp::Ordering;
        use UpgradePriority as Priority;

        match (self, other) {
            (Priority::Irrelevant, Priority::Irrelevant) => Ordering::Equal,
            (Priority::Irrelevant, _) => Ordering::Less,
            (Priority::Interested(s), Priority::Interested(o)) => s.cmp(o),
            (Priority::Interested(_), _) => Ordering::Less,
            (Priority::Critical, Priority::Critical) => Ordering::Equal,
            _ => Ordering::Greater,
        }
    }
```

src/models/state/mempool/upgrade_priority.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+impl UpgradePriority {
+    /// Returns true if the priority is irrelevant.
+    pub(crate) fn is_irrelevant(&self) -> bool {
+        matches!(self, UpgradePriority::Irrelevant)
```
jan-ferdinand: No need to match.

```suggestion
        self == &Self::Irrelevant
```

src/models/state/mempool/upgrade_priority.rs
```diff
@@ -0,0 +1,96 @@
...
+    fn add(self, other: Self) -> Self::Output {
+        match (self, other) {
+            (UpgradePriority::Irrelevant, _) => other,
+            (_, UpgradePriority::Irrelevant) => self,
+            (UpgradePriority::Interested(self_amt), UpgradePriority::Interested(other_amt)) => {
+                UpgradePriority::Interested(self_amt + other_amt)
+            }
+            (_, UpgradePriority::Critical) | (UpgradePriority::Critical, _) => {
+                UpgradePriority::Critical
+            }
```
jan-ferdinand: Nicer formatting, but it's not important:

```suggestion
            (Self::Irrelevant, _) => other,
            (_, Self::Irrelevant) => self,
            (Self::Interested(l), Self::Interested(r)) => Self::Interested(l + r),
            (_, Self::Critical) | (Self::Critical, _) => Self::Critical,
```

src/models/state/wallet/transaction_output.rs
```diff
@@ -515,7 +515,8 @@ impl<I: Into<TxOutput>, T: IntoIterator<Item = I>> From<T> for TxOutputList {
 }
 
 impl TxOutputList {
-    /// calculates total amount in native currency
+    /// calculates total amount in native currency, regardless of which other
+    /// coins are present (even if that makes the Neptune coins unspendable).
```
jan-ferdinand: I don't understand what the part in parenthesis means.

src/models/state/wallet/wallet_state.rs
```diff
@@ -4919,39 +4923,28 @@ pub(crate) mod tests {
 
             // upgrade transaction again
             // this time mutator set data
+            // let upgrade_incentive = UpgradeIncentive::Gobble(())
```
jan-ferdinand:
```suggestion
```

jan-ferdinand (approved):
I must admit that I don't understand all implications of these changes; distributed protocols are hard to reason about. In general, the changes look good to me. I like the idea of tracking interest in a transaction from potential, in the form of fees, to concrete, once those fees have been claimed.

I've left questions, suggestions, and nits inline.

src/models/state/wallet/transaction_output.rs
Sword-Smith: A UTXO is defined as:
```rust
pub struct Utxo {
    lock_script_hash: Digest,
    coins: Vec<Coin>,
}
```

The `coins` field enumerates the type scripts that the UTXO interacts with. That's what the "coins" in that sentence refers to.

jan-ferdinand: In other words, “Neptune coins” does not mean “native currency”? That might be the source of my confusion.

src/models/state/mempool.rs
Sword-Smith: The mempool is in a broken state if that unwrap doesn't work. I want to just crash in that case, not accept a broken mempool state.

PR #611, opened 2025-06-13
Gracefully catch errors resulting from invalid blocks

aszepieniec:
The peer loop should act as a robust firewall against invalid blocks.
Nevertheless, it is worth ensuring that handlers behind that firewall
cannot crash even for invalid blocks.

In particular, function `Block::total_guesser_reward` and other
helper functions downstream from there, are undefined for
un-validated blocks. The result should be wrapped in a `Result`, in
order to avoid passing inconsistent values up the call graph. This
change cascades into many function return type modifications
and many `.unwrap`s and `.expect`s downstream.

The inconsistency was exposed by test
`rpc_server::rpc_server_tests::public_announcements_in_block_test`
which is now fixed instead of ignored. Thanks to @skaunov for PR #543
and the effort leading to this exposition.

Also:
 - Change logic for computing block proposal favorability: use
   `prev_block_digest` instead of block height. However, the old
   method (now suffixed with `_legacy`) continues to exist to
   enable unmodified processing of `PeerMessage`
   `BlockProposalNotification`, which comes with a block height and
   not a digest. (We cannot change `PeerMessage`, unfortunately.)
 - Rewrite handler for `PeerMessage` `BlockProposal`. Simplify and
   reduce indentation.

### DIFF
--- a/benches/wallet_state.rs
+++ b/benches/wallet_state.rs
@@ -133,7 +133,7 @@ mod maintain_membership_proofs {
             rt.block_on(async {
                 wallet_state
                     .update_wallet_state_with_new_block(
-                        &genesis.mutator_set_accumulator_after(),
+                        &genesis.mutator_set_accumulator_after().unwrap(),
                         &block1,
                     )
                     .await
@@ -156,7 +156,7 @@ mod maintain_membership_proofs {
                 rt.block_on(async {
                     wallet_state
                         .update_wallet_state_with_new_block(
-                            &block1.mutator_set_accumulator_after(),
+                            &block1.mutator_set_accumulator_after().unwrap(),
                             &block2,
                         )
                         .await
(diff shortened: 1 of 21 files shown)


Sword-Smith (approved):
Pretty happy with how this looks. Only question mark is the text in the `expect` method calls. They might lead developers to think that the blocks need to be valid to be stored? Maybe it would be more precise if they said that the MS update for a block from state must exist, or something similar.

aszepieniec:
Can you be more explicit about the counter-proposal?

> that the blocks need to be valid to be stored?

I don't think this is false, but to the extent that the `expect` messages convey this meaning, it is besides the point.

The meaning that should be conveyed (IMHO) is the syllogism:
 - the block is valid
 - all valid blocks have mutator set updates
 - therefore, this block has a mutator set update
 - therefore, it is safe to unwrap.

That said, I don't think I want to be quite as explicit as that. I'm sure some propositions can be left as context assumed to be familiar to the reader.

aszepieniec:
> Let me be more explicit then: "mutator set update from stored block must exist"?

1) More explicit still, please: which message(s) do you want to substitute for this?

2) By omitting mention of validity, this phrase makes it seem as though being stored is the relevant feature that makes a block's mutator set update exist. However, the block's validity is the more proximal cause. The chain of implications is *block is stored* => *block is valid* => *mutator set update exists*. As I read it, this proposal is skipping a step and therefore begs the question, *why* is it skipping a step?

Issue #215, opened 2024-10-29
Negative tests for `Block::is_valid`


aszepieniec:
- [ ] 1. Previous_block` is consistent with current block
   - [ ] a) Block height is previous plus one
   - [ ] b) Block header points to previous block
   - [ ] c) block mmr updated correctly
   - [ ] d) Block timestamp is greater than previous block timestamp
   - [ ] e) Target difficulty and cumulative proof-of-work were updated correctly
   - [ ] f) Block timestamp is less than host-time (utc) + 2 hours.
 - [ ] 2. Block is valid
   - [ ] a) block proof is valid
   - [ ] b) max block size is not exceeded
 - [ ] 3. The transaction is valid
   - [ ] a) Verify that MS removal records are valid, done against previous `mutator_set_accumulator`,
   - [ ] b) Verify that all removal records have unique index sets
   - [ ] c) verify that we can add `mutator_set_update` to previous `mutator_set_accumulator`,
   - [ ] c') and that it results in new block's `mutator_set_accumulator`
   - [ ] e) transaction timestamp <= block timestamp
   - [ ] f) transaction coinbase <= miner reward
   - [ ] g) transaction is valid (internally consistent)

Sword-Smith:
This function should return a result type with a custom error type, not just a boolean. That will make these negative tests *much* more robust.

skaunov:
The tx from the section "3." is always a merge tx, am I correct? 

I'm currently at "a) Verify that MS removal records are valid, done against previous mutator_set_accumulator", and the easiest way seems to spend something which is added to the block as a part of reward (coinbase or that merging part which prover can send to himself subtracting from the fee) -- is it possible somehow, or is the only way adding a removal record via an ordinary tx?

PR #625, opened 2025-07-03
#215

skaunov:
the failed checks can't download the proofs they need


src/models/blockchain/block/mod.rs
```diff
@@ -861,11 +861,9 @@ impl Block {
             return Err(BlockValidationError::TransactionTimestamp);
         }
 
-        let block_subsidy = Self::block_subsidy(self.kernel.header.height);
```
aszepieniec: Speaking about my personal preference: I find the expression `Self::block_subsidy(self.kernel.header.height)` complex enough to justify binding it to a well-named variable for the sake of improving code readability. That said, if the block subsidy is only a function of block height (and it is) then you could make an argument that `block_subsidy` should be a method on `BlockHeight` and `Block` could have a convenience method that returns `self.kernel.header.height.block_subsidy()`. In this case, the "complex" expression becomes `self.block_subsidy()`, which is not worth binding to a variable.

All of this is subjective preference, of course, which I am not interested in enforcing as part review process.

src/models/blockchain/block/mod.rs
```diff
@@ -1764,23 +1762,27 @@ pub(crate) mod tests {
             #[strategy((#ts_previous.0.value() + 6000)..=(BFieldElement::MAX - 1288490188500000))]
             ts_now: u64,
             // #[strategy((#ts_now + 6000)..(#ts_now + 1288490188500000))] ts_next: u64,
+            #[strategy(..(#b_previous.kernel.header.timestamp.0.value() + 60))] ts_kernel_min: u64,
+            // b_now.kernel.header.timestamp
```
aszepieniec: I note that I am spending a lot of time guessing the purpose of variables with abbreviated names. I assume that "ts" abbreviates "timestamp", which by the way is not clear from the variable's type but from its arbitrary strategy. Does `min` abbreviate "minimum" or "minutes"? If the latter, note that the unit of `Timestamp` is milliseconds.

When in doubt, prefer the more explicit variable name.

src/models/blockchain/block/mod.rs
```diff
@@ -2047,44 +2050,227 @@ pub(crate) mod tests {
...
+            b_now_big_coinbase.kernel.appendix =
+                BlockAppendix::new(BlockAppendix::consensus_claims(b_now_big_coinbase.body()));
+            prop_assert_eq![
+                BlockValidationError::CoinbaseTooBig,
+                b_now_big_coinbase
+                    .validate(&b_previous, now, Network::Main)
+                    .await
+                    .err()
+                    .unwrap()
+            ];
```
aszepieniec: Here is a suggestion to enable code reviewers to quickly verify that all variants of `BlockValidationError` are covered. Instantiating a list containing all variants. Then validation failure, remove the returned `BlockValidationError` from the list. In the end, the list should be empty.

aszepieniec: Or better yet, collect all returned variants into a list and then show that the list contains all variants.

src/models/blockchain/block/mod.rs
```diff
@@ -2047,44 +2050,227 @@ pub(crate) mod tests {
...
+            let b_now_timestamp_header = b_now.kernel.header.timestamp;
+            let mut tx_kernel =
+                TransactionKernelProxy::from(b_now.kernel.body.transaction_kernel.clone());
+            if b_now.kernel.body.transaction_kernel.timestamp > b_now_timestamp_header {
+                tx_kernel.timestamp = b_now_timestamp_header;
+            }
+            tx_kernel.inputs = Vec::new();
+            b_now.kernel.body.transaction_kernel = tx_kernel.into_kernel();
+            b_now.kernel.appendix =
+                BlockAppendix::new(BlockAppendix::consensus_claims(b_now.body()));
```
aszepieniec: Work in progress, I guess? I'm not sure if this block of code belongs to the previous (AppendixTooLarge) or next (CoinbaseTooBig).

aszepieniec (changes requested):
I like where this is going.

 - Please factor out one function for each error code. If the functions rely on common setup code then either factor out the set up code into functions as well or call all error code functions from one master test function which does the setup beforehand.
 - Please avoid new abbreviations in variable names.
 - To show that all error codes can be triggered, please collect them all into a list and show that the list is exhaustive.
 - Github is not showing the differences with respect to master correctly, probably because the branch is not based on master. Please rebase to make reviewing easier.

skaunov:
somehow it didn't linked with #215 yet

skaunov:
It was based on `master`, that just went forward while I was making this. That's a reason I publish some partial work done -- to have less unmerged rebases (since merged won't need more unproductive work).

>     * Please factor out one function for each error code. If the functions rely on common setup code then either factor out the set up code into functions as well or call all error code functions from one master test function which does the setup beforehand.

Np, but are you sure in that? My idea was to have the part which doesn't need a proof to be in a property test, and then hitting all from that one would reduce the fuzzying generation just to one instance. Also I don't think those will be sharing the setup as each `Error` type needs own slightly different setup (which seems to become unnecessarily bloated). If you are sure, then it'd wonderful if you'd take a minute to share the rationale -- it would be really educative to me! :pray: 

>     * To show that all error codes can be triggered, please collect them all into a list and show that the list is exhaustive.

Tbh, it didn't aim to trigger "all error codes", but those which are described in #215. Do you have an example of such a test util somewhere? I'm trying to grasp this: do you mean a higher-order test with `match` in it? :thinking:

src/models/blockchain/block/mod.rs
```diff
@@ -873,40 +870,76 @@ impl Block {
...
+            .iter()
+            .map(|removal_record| removal_record.absolute_indices.to_vec())
+            .collect_vec();
+        absolute_index_sets.sort();
+        absolute_index_sets.dedup();
+        if absolute_index_sets.len() != self.kernel.body.transaction_kernel.inputs.len() {
+            Err(BlockValidationError::RemovalRecordsUniqueness)
+        } else {
+            Ok(())
+        }
```
aszepieniec: While I find that factoring out these individual checks improves readability, I do recall that @Sword-Smith has the opposite preference. If I recall correctly, having all the logic in one place makes it easier to a) audit, and b) translate into tasm. @Sword-Smith: please add anything I'm missing.

src/models/blockchain/block/mod.rs
```diff
@@ -1531,17 +1568,20 @@ pub(crate) mod tests {
...
         use super::*;
         use crate::mine_loop::tests::make_coinbase_transaction_from_state;
+        use crate::models::proof_abstractions::verifier::cache_true_claim;
         use crate::models::state::tx_creation_config::TxCreationConfig;
         use crate::models::state::wallet::address::KeyType;
         use crate::tests::shared::blocks::fake_valid_successor_for_tests;
+        use crate::tests::shared::Randomness;
         use crate::triton_vm_job_queue::TritonVmJobPriority;
         use crate::{
             api::export::NeptuneProof, tests::shared::strategies::block::arbitrary_mostly,
```
aszepieniec: Multiple imports in one `use`-statement generate unnecessary merge conflicts. When the closing bracket is not on the same line as the opening bracket, it is often unclear how to resolve the merge conflict. For these reasons, we prefer to stick to "item" granularity -- by unmerging all possible use statements.

src/models/blockchain/block/mod.rs
```diff
@@ -1752,262 +1792,34 @@ pub(crate) mod tests {
...
-                Network::Main.genesis_difficulty()
-            } else {
-                difficulty_control(
-                    b_now.header().timestamp,
-                    b_previous.header().timestamp,
-                    b_previous.header().difficulty,
-                    Network::Main.target_block_interval(),
-                    b_previous.header().height,
+        proptest::prop_compose! {
+            fn n_strategy() (
```
aszepieniec: It is difficult to infer the purpose of this function from its code. Consider naming the function descriptively or adding an elucidating docstring.

src/models/blockchain/block/mod.rs
```diff
@@ -1752,262 +1792,34 @@ pub(crate) mod tests {
...
-                    b_now.header().timestamp,
-                    b_previous.header().timestamp,
-                    b_previous.header().difficulty,
-                    Network::Main.target_block_interval(),
-                    b_previous.header().height,
-                )
+        // #[traced_test]
+        // #[apply(shared_tokio_runtime)]
+        #[test]
+        fn n() {
```
aszepieniec: The purpose of this test is difficult to infer from its name. Please rename to something descriptive or add a comment. Suggestion: `can_trigger_all_error_codes`.

src/models/blockchain/block/mod.rs
```diff
@@ -2017,260 +1829,745 @@ pub(crate) mod tests {
             > Thorkil Schmidiger -- will never ask for money:
             I'm not sure it's possible to hit the `BlockValidationError::MutatorSetUpdatePossible` error.
```
aszepieniec: After investigating a little, I think `BlockValidationError::MutatorSetUpdateImpossible` and `BlockValidationError::NegativeFee` cannot both be triggered, and only whichever one comes first in `Block::validate` can.

With an eye to supporting future changes (in particular, future reasons why the mutator set update could be impossible), I suggest to modify the error codes as follows.

 - Drop `NegativeFee` from `BlockValidationError`.
 - Add new error type `MutatorSetUpdateError` which (for now) contains only one variant: `NegativeFee`.
 - Add a `MutatorSetUpdateError` object as associated data to `BlockValidationError::MutatorSetUpdateImpossible`.

src/models/blockchain/block/mod.rs
```diff
@@ -1752,262 +1792,34 @@ pub(crate) mod tests {
...
-            assert_eq![
-                BlockValidationError::RemovalRecordsValidity,
-                b_now_bad_spending
-                    .validate(&b_previous, now, Network::Main)
-                    .await
-                    .err()
-                    .unwrap()
-            ];
-
             /* > S:
```
aszepieniec: Conversations like this can be used to motivate design choices. However let's improve readability by:
 - stripping descriptors in names ("will never ask for money" is irrelevant and disctracting);
 - adhering to the max line length (80 chars);
 - killing comments that are no longer relevant (not the case here though).

src/models/blockchain/block/mod.rs
```diff
@@ -2017,260 +1829,745 @@ pub(crate) mod tests {
...
-            // assert_eq![
-            //     // BlockValidationError::MutatorSetUpdateImpossible,
-            //     BlockValidationError::MutatorSetUpdateIntegrity,
-            //     b_now_bad_update.validate(&b_previous, now, Network::Main).await.err().unwrap()
-            // ];
-
-            let mut b_now_kernel_ts = b_now.clone();
-            let mut tx_kernel_ts = TransactionKernelProxy::from(
-                b_now_kernel_ts.kernel.body.transaction_kernel.clone(),
+            err_variants.retain(|v| v != &BlockValidationError::MutatorSetUpdateImpossible);
```
aszepieniec: The comment above motivates the *why*, but it still takes some effort to figure out the *what* about what's going on. A comment can help.

```suggestion
            // filter out unreachable error codes
            err_variants.retain(|v| v != &BlockValidationError::MutatorSetUpdateImpossible);
```

src/models/blockchain/block/mod.rs
```diff
@@ -2017,260 +1829,745 @@ pub(crate) mod tests {
...
-            // Test CoinbaseTooBig
-            let mut b_now_big_coinbase = b_now.clone();
-            let mut tx_kernel_big_coinbase = TransactionKernelProxy::from(
-                b_now_big_coinbase.kernel.body.transaction_kernel.clone(),
+            /* S Kaunov, [12.07.2025 18:30]
+            >> ... So, you already did the negative test for this case back than: block_with_invalid_proof_fails it is, am I correct?
+
+            Thorkil Schmidiger -- will never ask for money, [15.07.2025 20:02]
+            > Yeah. I added that negative test. */
+            err_variants.retain(|v| v != &BlockValidationError::ProofValidity);
```
aszepieniec:
```suggestion
            // filter out error codes that were already checked by other tests
            err_variants.retain(|v| v != &BlockValidationError::ProofValidity);
```

src/models/blockchain/block/mod.rs
```diff
@@ -2017,260 +1829,745 @@ pub(crate) mod tests {
...
+                        fake_valid_successor_for_tests(
+                            &b_prev,
+                            ts,
+                            rness,
+                            Network::Main,
+                        ).await.validate(&b_prev, ts, Network::Main).await.is_ok()
+                    });
+                    prop_assert![t];
+                }
             );
```
aszepieniec: I'm having trouble parsing this. You're defining a proptest within a test, which verifies that arbitrary blocks can be validated successfully? If correct, why? If not, what am I missing?

src/models/blockchain/block/mod.rs
```diff
@@ -2017,260 +1829,745 @@ pub(crate) mod tests {
...
+                    });
+                    prop_assert_eq![err, t.err().unwrap()];
+                }
             );
-            tx_kernel_big_coinbase.coinbase = Some(
-                NativeCurrencyAmount::one()
-                    + Block::block_subsidy(b_now_big_coinbase.header().height),
+            err_variants.retain(|v| v != &err);
+
+            let err = BlockValidationError::RemovalRecordsValidity;
```
aszepieniec: The git commit hook disallows unrelated variable shadowing. (And CI will fail if you commit with `-n` anyway.) As `err` is only used once between every binding, you might as well inline the value. That would also benefit readability IMHO.

src/models/blockchain/block/mod.rs
```diff
@@ -2017,260 +1829,745 @@ pub(crate) mod tests {
...
+                        b_prev.kernel.header.timestamp + Timestamp(bfe![60]),
+                        Network::Main
+                    ));
+                    prop_assert_eq![err, t.err().unwrap()];
+                }
+            );
+            err_variants.retain(|v| v != &err);
+
+            dbg![err_variants];
+            // assert!(dbg![err_variants].is_empty())
```
aszepieniec: This line is kind of important, no?

aszepieniec (changes requested):
I left inline comments. General comments follow below.

Presently you are removing elements from a list that starts out exhaustive, and verifying emptiness at the end. I think it's cleaner to do it the opposite way: add elements to a list that starts out empty and verify exhaustiveness (exhaustivity? exhaustion?) at the end.

Please find a way to make every block of code that triggers one particular error code into a separate, named, function. Then the wrapper test calls a bunch of functions, collects the error codes they generate, and asserts that the produced collection is exhaustive.

> Np, but are you sure in that? My idea was to have the part which doesn't need a proof to be in a property test, and then hitting all from that one would reduce the fuzzying generation just to one instance. Also I don't think those will be sharing the setup as each Error type needs own slightly different setup (which seems to become unnecessarily bloated). If you are sure, then it'd wonderful if you'd take a minute to share the rationale -- it would be really educative to me! 🙏

Excuse me for the slow response. (I do not recall getting a notification. Feel free to tag me to ensure that I do get one, and poke me over other channels after a long while of radio silence.) 

Factor out a common instance generator. Call it from every individual function and then modify that instance to trigger the targeted error code. I do not see why that strategy results in code that is more bloated than any alternative.

I do think that this strategy is difficult to marry with a fuzzing proptest, but then I'm not sure what the added value of fuzzing is, or of proptest for that matter, in the context of verifying that all error codes can be triggered. As I see it, the purpose of verifying that all error codes can be triggered is two-fold:

 1. Identify redundant checks. (You got one. Good job!)
 2. Identify incomplete or missing checks. This identification comes as a result of thinking through the logic, which catching all error codes forces you to do.

I do not think that the use of proptest or fuzzing in general helps with either objective, although it doesn't hurt them either.

> Currently it didn't aim to trigger "all error codes", but those which are described/mentioned in #215.

Issue #215 is a snapshot of our thinking at the time it was last edited, which pre-dates `BlockValidationError`. A much more up-to-date snapshot is given by `BlockValidationError`. If you want to restrict scope to the errors mentioned in that issue, then the wrapper test function needs to explicitly account for the variants that are present on the enum and not in the issue. (In this case, please still write it so that we have a template to complete.)

src/models/blockchain/block/mod.rs
skaunov: I guess I'm with @Sword-Smith on this to some extent. And am still see the case (here) for moving out suitable functions for testability. I mean I guess testing can impact a code to this extent (until it guides to clean and clear division into those units) and it's a good practice. The moved out functions would require to make a quite different (malicious) block producer to hit them. And I'd also add to these the one that I didn't hit at all (I guess it's the one @aszepieniec noted be the other redundant which was found during this task); with this approach it can have the snippet showing how and why it couldn't be hit currently and show that if it could be it's still on guard.

skaunov:
@Sword-Smith , yeah property part have been absolutely redundant all the way it's more for me to delve into how the things around works.

@aszepieniec 
I guess I'll be preparing now the version without moving out the submethods and without that coverage part. And will be happy to extend the result with what you decide on this -- everything is here already conceptually, so if you'd like it to be covering all variants with the submethods I'll extend the contribution.

neptune-core/src/models/blockchain/block/block_body.rs
```diff
@@ -61,6 +61,7 @@ impl HasDiscriminant for BlockBodyField {
 // We likewise skip the field for `BFieldCodec`, and `GetSize` because there
 // exist no impls for `OnceLock<_>` so derive fails.
 #[derive(Clone, Debug, Serialize, Deserialize, BFieldCodec, GetSize, TasmObject)]
+// #[cfg_attr(any(test, feature = "arbitrary-impls"), derive(arbitrary::Arbitrary))]
```
Sword-Smith: Let's not add commented-out code to this repo.

neptune-core/src/models/blockchain/block/block_validation_error.rs
```diff
@@ -95,3 +96,645 @@ impl From<RemovalRecordListUnpackError> for BlockValidationError {
...
+                            tx_kernel_neg_coinbase.into_kernel();
+                        b_new.kernel.appendix = BlockAppendix::new(BlockAppendix::consensus_claims(b_new.body(), ConsensusRuleSet::infer_from(Network::Main, b_prev.header().height)));
+                        cache_true_claim(BlockProgram::claim(b_new.body(), &b_new.kernel.appendix)).await;
+                    }
+                } else {panic![]}
+
+                b_new.validate(&b_prev, ts, Network::Main).await
+            });
+            prop_assert_eq![err_checking, t.err().unwrap()];
+        });
```
Sword-Smith: Didn't we talk about having one simple test per error variant? This is the opposite.

skaunov: AFAIR it turned out to be a best between @aszepieniec demand for the check of exhaustion and yours for separation (ironically being none of two, yeah). I mean these `proptest!` inside are completely independent and I can quickly convert it to a test per variant if we ditching the check that it exhausts the variants. (If we decide that this check is needed I can quickly turn it back but then one way or another it ends up being packaged in one test.)

Sword-Smith: I'm sorry there have been conflicting messages. We still have some things to learn when it comes to collaboration on open source projects :)

Please just keep the tests as simple as possible. Use `block_with_empty_proof_fails` and `block_with_invalid_proof_fails` as templates for these negative tests. Write one test per variant of `BlockValidationError` that you can hit and hit as many of the variants as you can.

Sword-Smith: So yes, please ditch the check that it exhausts the variants.

Cargo.toml
```diff
@@ -109,4 +109,4 @@ build-override.opt-level = 3
 # The profile that 'dist' will build with
 [profile.dist]
 inherits = "release"
-lto = "thin"
+lto = "thin"
```
Sword-Smith, line 112: Try to keep your diffs minimal and don't add various whitespace changes throughout the codebase since it makes the reviewers task bigger than it has to and since it obscures the git history.

neptune-core/src/mine_loop.rs
```diff
@@ -1129,7 +1128,7 @@ pub(crate) mod tests {
                 .map(|_| pseudorandom_addition_record(rng.random()))
                 .collect_vec();
             (
-                make_mock_block_transaction_with_mutator_set_hash(
+                crate::tests::shared::mock_tx::make_mock_block_transaction_with_mutator_set_hash(
```
Sword-Smith: why?

neptune-core/src/models/blockchain/block/block_validation_error.rs
```diff
@@ -5,6 +5,7 @@ use crate::util_types::mutator_set::removal_record::removal_record_list::Removal
 ///
 /// Conversely, defines what it means for a block to be "valid".
 #[derive(Debug, Clone, Copy, thiserror::Error, PartialEq, Eq)]
+// #[cfg_attr(test, derive(strum::VariantArray))]
```
Sword-Smith: Avoid adding commented-out code, please.

Cargo.toml
skaunov: :+1: Are we even interested in the end of file blanks? What would happen if just turn off the tracking of these in Git?

neptune-core/src/tests/shared/blocks.rs
```diff
@@ -47,6 +47,8 @@ use crate::util_types::mutator_set::addition_record::AdditionRecord;
 use crate::util_types::mutator_set::mutator_set_accumulator::MutatorSetAccumulator;
 use crate::util_types::mutator_set::removal_record::RemovalRecord;
 
+pub const DIFFICULTY_LIMIT: u32 = 1_000_000_000;
```
Sword-Smith, line 50: Way too high. Tests take forever because valid PoW must be found. I changed this value to 20.000. I known you did not introduce this constant, I did, but you expanded its scope in a way that meant that the tests you add in this PR take forever.

I will fix this.

Sword-Smith:
Merged via CLI in 6ea4c12461f7982b8e1fd617005b829417fca5a5

PR #629, opened 2025-07-19
an arrangement for #215

skaunov:
I didn't go into all the details - just took the arrangement/ordering that I got during #215, so probably missed something (like in a couple of place I guess from some general assumptions which one is more costly). (Other changes were due to that Git hook.)


src/util_types/mutator_set/removal_record/removal_record_list.rs
```diff
@@ -342,7 +342,7 @@ impl RemovalRecordList {
                 .map(MmrMembershipProof::new),
         );
 
-        use itertools::EitherOrBoth::*;
+        use itertools::EitherOrBoth::{Both, Left, Right};
```
Sword-Smith, line 345: I prefer to avoid grouping elements like this in imports. Three lines are better: 

```rust
use itertools::EitherOrBoth::Both;
use itertools::EitherOrBoth::Left;
use itertools::EitherOrBoth::Right;
```

Sword-Smith:
Wrt. the mutator set update validity, the 1st step as you've probably noticed is now the unpacking of the removal records list. That function (RemovalRecordList::try_unpack) is tested for an inability to crash, and since it's the first thing that needs to be done for the mutator set update validity check, it's the main line of defense against malicious data. And the most exposed function in terms of panicking risks.

src/util_types/mutator_set/removal_record/removal_record_list.rs
skaunov: it was just `cargo clippy --fix` tbh

Sword-Smith:
> @Sword-Smith
> 
> > if that helps you write negative tests.
> 
> I see such helping as the lateral/secondary thing and primary to sort in the order of the checking cost. Am I correct `BlockProgram::verify` is the most expensive one?

I don't have data for each check. As far as I remember proof verification is around 10ms.  I can try to make qualified guesses on how expensive each guess is though:

- 0.a 1ns
- 0.b 1ns
- 0.c 1µs
- 0.d 10ns
- 0.e 100ns
- 0.f 10ns
- 0.g 10ns
- 1.a 20ns
- 1.b 1ns
- 1.c 10ns
- 1.d 10ms
- 1.e 5ms? (encoding is probably a bit expensive. Size is at least 1MB)
- 2.a 1ms
- 2.b 2ms?
- 2.c 10µs
- 2.d 1ms
- 2.e 10µs
- 2.f 10ns
- 2.g 10ns
- 2.h 5ns
- 2.i 5ns
- 2.j 5ns
- 2.k 5ns
- 2.l 5ns

However, downloading the block from a peer also takes time. Probably around 1 second. And if all verification is 10x faster than downloading the block, I'm not sure the order really matters that much for performance issues.

Sword-Smith:
Merged all but the rearranging of checks through 034f14bf613f1364b9551a07a1fe288f3c2fbbf5. @skaunov please close this PR if you're OK with my approach and inclusion of your proposed changes.

PR #630, opened 2025-07-20
Promote public announcements to first-class citizenship

aszepieniec:
The main objective of this PR is to expand the `TransactionDetailsBuilder` interface with support for setting arbitrary public announcements. (Almost) all changes constitute scaffolding in support of this objective.

Currently, there is only one kind of public announcement: encrypted UTXO notification. This state suffices for all use cases we have seen so far. However, in the future we will need to produce transactions with arbitrary public announcements. (Indeed, I need just this functionality over on branch [asz/redeem-endpoints](https://github.com/Neptune-Crypto/neptune-core/tree/asz/redeem-endpoints), although the particular kind of transaction being produced there should probably not make it to `master`).

Currently, the public announcements that are encryped UTXO notifications are produced on the fly from the transaction outputs when the transaction kernel is produced, and not built by the `TransactionDetailsBuilder`. I kept this separation of concerns in order to avoid populating `TransactionDetails` with duplicated (and therefore, possibly inconsistent) data. This choice does mean that we have to be mindful about the distinction between encrypted UTXO notifications and other public announcements when supplying them. One question for reviewers (besides the usual) is whether the interface and documentation here adequately dispels the potential confusion.


src/api/tx_initiation/builder/transaction_details_builder.rs
```diff
@@ -82,6 +85,20 @@ impl TransactionDetailsBuilder {
         self
     }
 
+    /// Add many public announcements.
+    ///
+    /// Use this method for public announcements that are *not* encrypted UTXO
+    /// notifications. The encrypted UTXO notifications are generated on the fly
+    /// at a later stage.
+    pub fn public_announcements<Iter: IntoIterator<Item = PublicAnnouncement>>(
```
skaunov: how do you like `pa_custom` name for the method?

aszepieniec:
> On a grande thought: do you think "public..." part in items naming belong at all? Would it look nicer if everything is renamed to just `Announcement`? I guess it's the right moment to do if so.

**Yes.**

aszepieniec:
The most recent commit, [ca7b3f5](https://github.com/Neptune-Crypto/neptune-core/pull/630/commits/ca7b3f515564f331487ae5d81cb0ed64e592ebcf), renames "public announcement" to "announcement". This is a breaking change: as tasm imports are sorted alphabetically, some tasm imports have switched order, resulting in different program hashes. This is okay for the reboot, but not good for the redemption branch.

src/models/state/transaction_details.rs
```diff
@@ -41,6 +44,9 @@ use crate::util_types::mutator_set::mutator_set_accumulator::MutatorSetAccumulat
 pub struct TransactionDetails {
     pub tx_inputs: TxInputList,
     pub tx_outputs: TxOutputList,
+
+    /// announcements *excluding* encrypted UTXO notifications.
```
dan-da, line 48: Perhaps call this field extra_announcements, to distinguish from the output utxo announcements.

Alternatively, all announcements (including from output utxos) could be placed in this field by the TransactionDetailsBuilder.   That would, I believe be symmetric with TransactionKernel.   A drawback is that it would make TransactionDetails potentially much larger for sending across the wire, especially when there are many output utxos.

src/api/tx_initiation/builder/transaction_details_builder.rs
```diff
@@ -82,6 +85,20 @@ impl TransactionDetailsBuilder {
...
+        mut self,
+        public_announcements: Iter,
+    ) -> Self {
+        self.public_announcements
+            .append(&mut public_announcements.into_iter().collect_vec());
```
dan-da:
```suggestion
self.public_announcements.extend(public_announcements)
```

* more concise & idiomatic
* more efficient: avoids intermediate allocation

src/api/tx_initiation/builder/transaction_details_builder.rs
```diff
@@ -261,7 +278,10 @@ impl TransactionDetailsBuilder {
                 .mutator_set_accumulator_after()
                 .expect("Block from state must have mutator set after"),
```
dan-da: This line is violating rule 3 in src/api/README-dev.md:

> 3. no unwrap(), expect(), or panic!() in this layer.

there should be a CreateTxError variant for this, and thus it should simply be:

```
tip_block.mutator_set_accumulator_after()?
```

dan-da (changes requested):
I left one change request and a couple suggestions.

PR #637, opened 2025-07-31
feat: add dump-db command to neptune-cli

dan-da:
context: 

I added this command to neptune-cli in case it is helpful for debugging issues with user's databases.

Both keys and values are stored as bytes inside level-db and are output (dumped) as hex encoded byte strings, with a space between each byte so it is easy to see visually identify each byte.

It requires an understanding of the expected data types to interpret the dumped data.

Example usage and output:

```
target/debug/neptune-cli dump-db testnet wallet`
Key (hex): 01 00 00 00 00 00 00 00 00
Value (hex): 28 00 00 00 00 00 00 00 d9 86 e9 b2 29 83 1f 77 9c ee 1a 55 6f 2a 63 d2 46 31 55 9e b9 b3 bb 8c 3d ac a8 d6 5d 1f 66 ac ca 1d 10 02 66 98 ad a8

Key (hex): 01 00 00 00 00 00 00 00 03
Value (hex): 28 00 00 00 00 00 00 00 d9 86 e9 b2 29 83 1f 77 9c ee 1a 55 6f 2a 63 d2 46 31 55 9e b9 b3 bb 8c 3d ac a8 d6 5d 1f 66 ac ca 1d 10 02 66 98 ad a8

Key (hex): 01 00 00 00 00 00 00 00 05
Value (hex): 08 00 00 00 00 00 00 00 01 00 00 00 00 00 00 00

Key (hex): 01 00 00 00 00 00 00 00 06
Value (hex): 08 00 00 00 00 00 00 00 01 00 00 00 00 00 00 00

Key (hex): 01 00 00 00 00 00 00 00 08
Value (hex): 02 00 00 00 00 00 00 00 01 00
```

commit message follows.

-----------------

This command enables dumping key/val pairs of all values in a given leveldb database as hex.

```
Usage: neptune-cli dump-db <NETWORK> <DB_STORE>

Arguments:
  <NETWORK>
  <DB_STORE>  [possible values: archival-block-mmr, banned-ips, block-index, mutator-set, wallet]
```

### DIFF
--- a/src/bin/neptune-cli.rs
+++ b/src/bin/neptune-cli.rs
@@ -21,6 +21,8 @@ use itertools::Itertools;
 use neptune_cash::api::tx_initiation::builder::tx_output_list_builder::OutputFormat;
 use neptune_cash::config_models::data_directory::DataDirectory;
 use neptune_cash::config_models::network::Network;
+use neptune_cash::database::storage::storage_schema::RustyKey;
+use neptune_cash::database::NeptuneLevelDb;
 use neptune_cash::models::blockchain::block::block_selector::BlockSelector;
 use neptune_cash::models::blockchain::type_scripts::native_currency_amount::NativeCurrencyAmount;
 use neptune_cash::models::state::wallet::address::KeyType;
@@ -37,6 +39,8 @@ use neptune_cash::models::state::wallet::wallet_status::WalletStatusExportFormat
 use neptune_cash::rpc_auth;
 use neptune_cash::rpc_server::error::RpcError;
 use neptune_cash::rpc_server::RPCClient;
+
+use neptune_cash::database::storage::storage_schema::RustyValue;
 use rand::Rng;
 use regex::Regex;
 use serde::Deserialize;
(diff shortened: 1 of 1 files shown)


Sword-Smith:
Very nice. I think we should add a warning not to share the output with others, unless there's a good reason. Not sure about the wording. To be fair, the wallet output can be used to deanonymize transactions, not to steal funds.

src/bin/neptune-cli.rs
```diff
@@ -159,6 +184,12 @@ enum Command {
     /// retrieve info about peers
     PeerInfo,
 
+    /// dump all key/val pairs in the indicated database
```
aszepieniec, line 187:
```suggestion
    /// Dump all key/val pairs in the indicated database.
    ///
    /// Note that the resulting data dump contains private information that would
    /// enable the reader to deanonymize transactions. Only share this information
    /// if the privacy risk is acceptable. This dump does not contain information
    /// that would enable the reader to spend UTXOs.
```

aszepieniec (approved):
LGTM

I had a stab at @Sword-Smith's suggested disclaimer. Note that the docstring is used for the command help info.

src/bin/neptune-cli.rs
Sword-Smith: For the wallet DB, this is correct. I think the other DBs don't leak anything money-related when dumped.

PR #640, opened 2025-08-03
style: repository into multiple crates

KaffinPX:
also I wanna propose removing wix as all binaries in this repo is already "low level", installation is overkill for a cli application.


aszepieniec:
> I wanna propose removing wix as all binaries in this repo is already "low level", installation is overkill for a cli application.

It is a good point. If an install wizard is necessary because unzipping is too difficult, then using a CLI application is definitely too difficult.

However, I think I prefer the optics of "I installed it but I don't know how to use it" over "I can't even install it".

aszepieniec:
Proof files (`test_data/*.proof`) should not be committed. Otherwise the repository grows too fast. That's what the [proof servers](https://github.com/Neptune-Crypto/proof-server) are for.

aszepieniec:
Merged via command line.

PR #642, opened 2025-08-05
feat: Add startup banner to CLI

hduoc2003:
Added a bold ASCII banner to neptune-core, neptune-cli, and neptune-dashboard for improved visual identity on startup.

image


src/bin/neptune-cli.rs
```diff
@@ -34,7 +34,7 @@ use neptune_cash::models::state::wallet::wallet_file::WalletFile;
 use neptune_cash::models::state::wallet::wallet_file::WalletFileContext;
 use neptune_cash::models::state::wallet::wallet_status::WalletStatus;
 use neptune_cash::models::state::wallet::wallet_status::WalletStatusExportFormat;
-use neptune_cash::rpc_auth;
+use neptune_cash::{display_banner, rpc_auth};
```
aszepieniec: We use "item" granularity for imports to avoid merge conflict headaches. Please un-merge these imports.

src/bin/neptune-dashboard.rs
```diff
@@ -9,7 +9,7 @@ use crossterm::terminal::LeaveAlternateScreen;
 use dashboard_src::dashboard_app::Config;
 use dashboard_src::dashboard_app::DashboardApp;
 use neptune_cash::config_models::data_directory::DataDirectory;
-use neptune_cash::rpc_auth;
+use neptune_cash::{display_banner, rpc_auth};
```
aszepieniec: Please un-merge.

src/bin/neptune-cli.rs
```diff
@@ -405,6 +405,7 @@ struct Config {
 
 #[tokio::main]
 async fn main() -> Result<()> {
+    display_banner();
```
aszepieniec: While I love the banner, I think it is noisy in the CLI tool. The CLI should be lean, producing as little output as possible, subject to meeting the user's demands. Please drop this line.

aszepieniec (changes requested):
Lovely contribution, thank you :-)

I left some minor change requests inline.

About the `neptune-banner.txt` asset file, could you explain to me how it works? What I'm concerned about is shipping the executable with a text file, and having that executable crash if the text file is not present. However, if the content of this txt file ends up as part of the executable, then it's all good in my book.

KaffinPX:
can we wait until crate separation PR bcs this would probably cause too many annoying conflicts?

src/bin/neptune-dashboard.rs
hduoc2003: It's done

src/bin/neptune-cli.rs
hduoc2003: It's done

src/bin/neptune-cli.rs
hduoc2003: It's done

aszepieniec:
I am getting identical pictures if I change the line 
```rust
println!("{}", NEPTUNE_BANNER.fg_rgb::<255, 255, 255>().bold());
```
into
```rust
println!("{}", NEPTUNE_BANNER);
```
.

At this point, new dependency `owo-colors` is not used. Is there something I'm not seeing? Does something break if we drop it? I prefer fewer dependencies to begin with but if there is not even any difference, there's a very easy case for dropping it ...

aszepieniec:
Closing in favor of #757.

PR #644, opened 2025-08-05
fix(wallet): persist key cache store correctly after derivation_index bump

KaffinPX:
on my case(same address, multiple UTXOs) all UTXOs after first UTXO was missing, wasn't getting discovered. Turns out reason is:

...bump_derivation_counter wasnt bumping addresses, only bumping derivation counter so while on scan mode, when first utxo was found derivation counter was getting bumped and that address wasn't getting scanned anymore by scan mode and other mode also wasnt scanning as it didnt add key to known keys cache.


KaffinPX:
> I'm confused. Where is the persisting happening?
> 
> I'm happy with these changes by the way.

image

Persistingn to known keys cache in this part by push.

PR #646, opened 2025-08-06
test: re-enable test that first blk-file contains valid blocks

Sword-Smith:
Since the block structure is different in production from what it is under the test flag, this tests, that checks the validity of the first ~120 mainnet blocks, cannot live under the test flag. So it's added as an integration test.

Unfortunately some test helper function were needed for this test to work, and since they can't be under the test flag, they are added as local functions in the new test. The alternative seems to be to pull the test helper functions in question out from under the test flag but that would require new dependencies in production code (namely HTTP client 'clienter') which I think is not preferable. So local functions was the best compromise I could come up with. Happy to have other people give me better ideas.

### DIFF
--- a/neptune-core/src/models/blockchain/block/mod.rs
+++ b/neptune-core/src/models/blockchain/block/mod.rs
@@ -1243,12 +1243,6 @@ pub(crate) mod tests {
     pub(crate) const PREMINE_MAX_SIZE: NativeCurrencyAmount = NativeCurrencyAmount::coins(831488);
 
     impl Block {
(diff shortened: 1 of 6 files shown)


skaunov (approved):
I didn't check whether the helpers are duplicated or simply moved out from the lib. If they're duplicated those would be cool to mark with a comment (I personally prefer to do it with some hashtag so it could be quickly and unambiguously searched with a _ctrl_+_shift_+_f_.) 

I doubt it will be touched again soon, so duplication seems to be not a problem, but if we want to be surely in sync for some reason a specific feature seems to be a fine fit here.

Generally moving this (kind of) test(s) to my eye looks as a step forward. :+1:

PR #647, opened 2025-08-06
fix(scan_mode): Bump known keys cache correctly

KaffinPX:
Fixed an issue where bumping the derivation counter did not generate or track the corresponding address.
This caused the scanning logic to skip addresses with valid UTXOs because they were neither scanned again
nor added to the known keys cache, leading to missed UTXOs across scan and normal modes.


aszepieniec (approved):
Looks good to me.

The previous PR had a refactor changing someting from a vec to an iterator. What was the motivation for dropping that?

KaffinPX:
> Looks good to me.
> 
> The previous PR had a refactor changing someting from a vec to an iterator. What was the motivation for dropping that?

It was for a planned future PR(index recovery by search), I decided to include it with that

PR #651, opened 2025-08-07
648 reduce block proposal switching frequency

### LINKED ISSUE #648: guesser: Don't automatically switch block proposal
When the node is guessing and a new block proposal with a higher fee comes in, the guesser should not automatically switch since the guesser-preprocessing step takes a non-trivial amount of time.

The best block proposal should probably still be relayed on the network but the mine loop should not switch unless the new proposal is better by X percent of the existing block proposal's guesser reward.

The economically optimal $X$ can be expressed as a function of current guesser reward $GR_0$ and the new guesser reward $GR_1$.

$X = \frac{GR_1}{GR_0} \geq \frac{1}{1 - \frac{PPT}{BTI'}}$, or $x = \frac{BTI'}{BTI' - PPT}$, where $BTI'$ is block target interval minus proving time, minus preprocessing time, and $PPT$ is the preprocessing time.

On a Threadripper 7995WX $PPT=60s$. Assuming a 120s proving time, this gives a $BTI'$ value of 7 minutes. So the fraction is approximately $\frac{7}{6}$, which translates to 15 %. If the new block proposal is better than the old by 15 %, a switch should be made. On weaker hardware, this value will be higher.

Also: No guessing should take place unless the block proposal has a minimum guesser fee of Y. X and Y should be configurable as CLI arguments.

Sword-Smith:
Closes #648. Currently a draft since a) there are a few tests missing, and b) we should discuss the default value of `minimum_guesser_improvement_fraction`. My calculations indicate that 10-30 % is a reasonable default value.

### DIFF
--- a/neptune-core/src/main_loop.rs
+++ b/neptune-core/src/main_loop.rs
@@ -1046,7 +1046,16 @@ impl MainLoopHandler {
                 let pmsg = MainToPeerTask::BlockProposalNotification((&*block).into());
                 self.main_to_peer_broadcast(pmsg);
 
-                self.main_to_miner_tx.send(MainToMiner::NewBlockProposal);
+                // Only notify miner if delta from current proposal is
+                // sufficient
+                if self
+                    .global_state_lock
+                    .lock_guard()
+                    .await
+                    .block_proposal_warrants_guess_restart(&block)
+                {
+                    self.main_to_miner_tx.send(MainToMiner::NewBlockProposal);
+                }
             }
             PeerTaskToMain::DisconnectFromLongestLivedPeer => {
                 let global_state = self.global_state_lock.lock_guard().await;
--- a/neptune-core/src/mine_loop.rs
+++ b/neptune-core/src/mine_loop.rs
@@ -710,41 +710,32 @@ pub(crate) async fn mine(
         let (guesser_tx, guesser_rx) = oneshot::channel::<NewBlockFound>();
         let (composer_tx, composer_rx) = oneshot::channel::<(Block, Vec<ExpectedUtxo>)>();
 
-        let maybe_proposal = global_state_lock
+        let proposal_meets_threshold = global_state_lock
             .lock_guard()
             .await
-            .mining_state
-            .block_proposal
-            .clone();
-        let guess = cli_args.guess;
-
+            .should_guess_on_block_proposal();
         let should_guess = !wait_for_confirmation
-            && guess
-            && maybe_proposal.is_some()
+            && cli_args.guess
+            && proposal_meets_threshold
             && !is_syncing
             && !pause_mine
             && is_connected;
-
-        // if start_guessing is true, then we are in a state change from
-        // inactive state to guessing state.
-        //
-        // if start_guessing is false and should_guess is true then we
-        // have already been guessing and are restarting with new params.
-        let start_guessing = matches!(
-            (mining_status, should_guess),
-            (MiningStatus::Inactive, true)
-        );
-
-        if start_guessing {
-            let proposal = maybe_proposal.unwrap(); // is_some() verified above
-            global_state_lock
-                .set_mining_status_to_guessing(proposal)
-                .await;
-        }
-
         let guesser_task: Option<JoinHandle<()>> = if should_guess {
-            // safe because above `is_some`
-            let proposal = maybe_proposal.unwrap();
+            let proposal = global_state_lock
+                .lock_guard()
+                .await
+                .mining_state
+                .block_proposal
+                .expect("Block proposal must be present when guesser threshold is met")
+                .clone();
+
+            // Only apply state change if we were not already guessing.
+            if matches!(mining_status, MiningStatus::Inactive) {
+                global_state_lock
+                    .set_mining_status_to_guessing(&proposal)
+                    .await;
+            }
+
             let guesser_key = global_state_lock
                 .lock_guard()
                 .await
@@ -757,7 +748,7 @@ pub(crate) async fn mine(
                 .await;
             let guesser_task = guess_nonce(
                 network,
-                proposal.to_owned(),
+                proposal,
                 latest_block_header,
                 guesser_tx,
                 GuessingConfiguration {
--- a/neptune-core/src/mine_loop.rs
+++ b/neptune-core/src/mine_loop.rs
@@ -718,8 +718,7 @@ pub(crate) async fn mine(
             && cli_args.guess
             && proposal_meets_threshold
             && !is_syncing
-            && !pause_mine
-            && is_connected;
+            && !pause_mine;
         let guesser_task: Option<JoinHandle<()>> = if should_guess {
             let proposal = global_state_lock
                 .lock_guard()
--- a/neptune-core/src/models/blockchain/block/mod.rs
+++ b/neptune-core/src/models/blockchain/block/mod.rs
@@ -1106,6 +1106,20 @@ impl Block {
         self.encode().len()
     }
 
+    /// A number showing how big the guesser reward is relative to the block
+    /// subsidy. Notice that this number can exceed 1 as it's defined a guesser
+    /// reward divided by block subsidy and with transaction fees, this value
+    /// can exceed 1.
+    ///
+    /// May not be used in any consensus-related setting, as precision is lost
+    /// because of the use of floats.
+    pub(crate) fn relative_guesser_reward(&self) -> Result<f64, BlockValidationError> {
+        let guesser_reward = self.total_guesser_reward()?;
+        let block_subsidy = Self::block_subsidy(self.header().height);
+
+        Ok(guesser_reward.to_nau_f64() / block_subsidy.to_nau_f64())
+    }
+
     /// The amount rewarded to the guesser who finds a valid nonce for this
     /// block.
     pub(crate) fn total_guesser_reward(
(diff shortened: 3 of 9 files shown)


neptune-core/src/main_loop.rs
```diff
@@ -1040,13 +1040,22 @@ impl MainLoopHandler {
...
                         BlockProposal::foreign_proposal(*block.clone());
-                }
 
-                // Notify all peers of the block proposal we just accepted
+                    global_state_mut.block_proposal_warrants_guess_restart(&block)
+                };
+
+                // Notify all peers of the block proposal we just accepted. This
+                // regardless of delta from previous suggestion (as long as
+                // delta) is positive.
```
aszepieniec: Very confusing comment. Please clarify.

Did you mean this?

 > **Do** this regardless of the **difference in guesser fee**, **relative to the** previous **proposal** (as long as **it** is positive **)**.

neptune-core/src/main_loop.rs
```diff
@@ -1040,13 +1040,22 @@ impl MainLoopHandler {
...
                 let pmsg = MainToPeerTask::BlockProposalNotification((&*block).into());
                 self.main_to_peer_broadcast(pmsg);
 
-                self.main_to_miner_tx.send(MainToMiner::NewBlockProposal);
+                // Only notify miner if delta from current proposal is enough.
+                // Unlike the informing of peers, this only happens when the
+                // delta is sufficiently big.
```
aszepieniec: The comments refer to a delta, but the actual code does not. Instead, the code uses (good, IMHO) abstractions like "whether the block proposal warrants a restart" or "whether we should inform the miner". It is confusing and I can't explain why better than saying because I expect the abstraction cascade to go the other way. Usually, the code is lower-level and the comments explain from an abstract high-level perspective what is happening. It's the reverse here.

IMHO the code is perfectly understandable and not in need of explanatory aid.

neptune-core/src/main_loop.rs
```diff
@@ -2703,6 +2715,128 @@ mod tests {
             instants.iter().copied().min_by(|l, r| l.cmp(r)).unwrap()
         );
     }
+
+    #[traced_test]
+    #[apply(shared_tokio_runtime)]
+    async fn received_new_block_proposal() {
```
aszepieniec: If test `received_new_block_proposal` fails, I have no idea what might have gone wrong. How about naming the test after the thing being tested? To that end, how about `should_switch_guessing_proposal_iff_sufficient_delta`?

neptune-core/src/mine_loop.rs
```diff
@@ -711,41 +704,29 @@ pub(crate) async fn mine(
...
-        let maybe_proposal = global_state_lock
+        let proposal_meets_threshold = global_state_lock
             .lock_guard()
             .await
-            .mining_state
-            .block_proposal
-            .clone();
-        let guess = cli_args.guess;
-
+            .should_guess_on_block_proposal();
```
aszepieniec: I am registering a similar confusion to the one I commented on earlier, except here no comments are involved. It's because the notions "proposal meets threshold" and "should guess on proposal" have different abstraction levels. I can infer that we probably should not be guessing on a proposal if it does not meet the threshold, but aren't there other requirements that a proposal must satisfy before guessing on it is warranted? If no, the function `should_guess_on_block_proposal` is improperly abstract and is better named like the variable: `proposal_meets_threshold`. If yes, what's the missing context that allows the code to equate these two things in this particular instance? That context is worth a comment.

neptune-core/src/mine_loop.rs
```diff
@@ -711,41 +704,29 @@ pub(crate) async fn mine(
...
         let should_guess = !wait_for_confirmation
-            && guess
-            && maybe_proposal.is_some()
```
aszepieniec, line 724: Why don't we care about having a proposal any more? Seems like a solid requirement to start guessing.

neptune-core/src/mine_loop.rs
```diff
@@ -711,41 +704,29 @@ pub(crate) async fn mine(
...
+            && cli_args.guess
+            && proposal_meets_threshold
             && !is_syncing
-            && !pause_mine
-            && is_connected;
```
aszepieniec, line 727: Why don't we care about being connected any more?

neptune-core/src/models/blockchain/block/mod.rs
```diff
@@ -1108,6 +1108,20 @@ impl Block {
         self.encode().len()
     }
 
+    /// A number showing how big the guesser reward is relative to the block
+    /// subsidy. Notice that this number can exceed 1 as it's defined a guesser
+    /// reward divided by block subsidy and with transaction fees, this value
+    /// can exceed 1.
```
aszepieniec:
```suggestion
    /// subsidy. Notice that this number can exceed 1 because of transaction fees.
```

neptune-core/src/models/state/block_proposal.rs
```diff
@@ -33,19 +33,11 @@ impl BlockProposal {
...
-    pub(crate) fn is_some(&self) -> bool {
-        !self.is_none()
-    }
-
-    pub(crate) fn is_none(&self) -> bool {
-        matches!(self, BlockProposal::None)
-    }
-
-    pub(crate) fn unwrap(&self) -> &Block {
+    pub(crate) fn expect(&self, msg: &str) -> &Block {
```
aszepieniec, line 36: <3

neptune-core/src/models/state/mod.rs
```diff
@@ -873,6 +874,52 @@ impl GlobalState {
...
+            // If currently composing, don't interrupt this process
+            MiningStatus::Composing(_) => false,
+
+            // If no proposal is being worked on, get started on one.
+            MiningStatus::Inactive => true,
+        }
+    }
+
+    /// Returns true iff the current block proposal should be guessed on.
+    pub(crate) fn should_guess_on_block_proposal(&self) -> bool {
```
aszepieniec: With a function named `should_guess_on_block_proposal`, I would expect the argument to be a block proposal. Pre-emptively dispel that confusion by disambiguating in the name, for instance: `should_guess_on_current_block_proposal`.

neptune-core/src/models/state/mod.rs
```diff
@@ -3573,106 +3620,165 @@ mod tests {
...
+                    received: big_guesser_fraction.total_guesser_reward().unwrap()
+                },
+                state
+                    .favor_incoming_block_proposal(
+                        big_prev_block_digest,
+                        big_guesser_fraction.total_guesser_reward().unwrap()
+                    )
+                    .unwrap_err(),
+                "Must favor existing over incoming equivalent"
+            );
```
aszepieniec, line 3784: I wonder if binding the left and right hand sides in this equation to well-named variables can reduce the mental overhead when parsing.

neptune-core/src/rpc_server.rs
```diff
@@ -4755,13 +4755,13 @@ mod tests {
...
             let bob_token = cookie_token(&bob).await;
-            assert!(!bob
-                .state
-                .lock_guard()
-                .await
-                .mining_state
-                .block_proposal
-                .is_some());
+            assert!(
+                matches!(
```
aszepieniec, line 4759: Some day (but not yet) this will be `assert_matches!`. [1]

[1]: https://doc.rust-lang.org/std/assert_matches/macro.assert_matches.html

neptune-core/src/mine_loop.rs
Sword-Smith: Because `proposal_meets_threshold` is false if no block proposal exists.

neptune-core/src/mine_loop.rs
Sword-Smith: Because of the `continue` on line 701 (above). This boolean is guaranteed to be true when the instruction pointer is here.

neptune-core/src/models/state/mod.rs
Sword-Smith: Test not added in this PR.

skaunov:
Functions named `relative_guesser_reward` are defined twice: once as the method, and another in the `tests`. I can imagine such approach could be nice, but can't see a reason for it here.

Also I wanted to ask if it ubiquitous enough to be a method? It seems to do fine as a local function in <./neptune-core/src/models/state/mod.rs>.

As for the comment on it it would be cool to save that extended version (personally I find it helpful) as another paragraph. Like the shorter one for the comfort of pros; and the other paragraph to read for those like me. X)

neptune-core/src/mine_loop.rs
aszepieniec: I think that explains a lot of my confusion. Here are two (mutually exclusive) design strategies that are less confusion-prone.

1. Rename the variable to more accurately reflect the conditions under which it is true, e.g., `proposal_is_available_and_meets_threshold`.
2. Refactor control flow so that this variable is in scope only if the block proposal exists.

neptune-core/src/models/state/mod.rs
```diff
@@ -873,6 +874,54 @@ impl GlobalState {
...
+
+            // If no proposal is being worked on, get started on one.
+            MiningStatus::Inactive => true,
+        }
+    }
+
+    /// Returns true iff the current block proposal is present and has a high
+    /// enough guesser fee to be worth guessing on. Always  returns true if
+    /// proposal was made locally.
+    pub(crate) fn current_block_proposal_meets_threshold(&self) -> bool {
```
aszepieniec, line 915: It is a bad idea to have a function whose name is a sentence describing the predicate it computes, compute a different predicate. Specifically, this function returns true if the current block proposal is an own composition that does not meet the threshold, whereas the function name suggests that it should return false. That's a recipe for mistakes.

Issue #654, opened 2025-08-11
Makefile


aszepieniec:
With crate separation, installing everything now requires 3 commands. Let's add a Makefile to turn it back into one.

aszepieniec:
Let's also add a command `make happy` to run the various linters, and add this command to the pre-push hook (and remove the linter commands from the pre-commit hook). Cfr. #655

Sword-Smith:
A Makefile already exists for this repo. Why don't we just expand on that?

PR #655, opened 2025-08-11
dx: Remove husky

KaffinPX:
Removed Husky “pre-commit” hook. It was redundant with CI checks and occasionally caused issues.


aszepieniec:
I hate the commit hook and would be glad to see it die. To avoid the delay, I end up using `-n` the overwhelming majority of the time, and the result is that the commit hook catches none of what it is supposed to catch. To make matters worse, `git rebase` has no `-n` option, leading to rebasings that consume unnecessary time and sometimes fail outright. So in summary:
 - time-consuming
 - misses own objective
 - hinders git discipline.

To address the motivation for keeping it, how about the following proposal?
 1. Add a Makefile so that you can write `make happy` and trigger the four commands.
 2. Add `make happy` to the *pre-push* hook, instead of the commit hook, so that committers are informed ahead of time that CI is going to fail.

Sword-Smith:
With #654 being closed, LGTM.

PR #658, opened 2025-08-12
test: Ensure deterministic STARK proofs

### LINKED ISSUE #657: Use deterministic proofs for tests
If the same test is started on two different machines simultaneously, they might produce two chains of proofs. If these proofs are used in blocks, the proofs are mutually incompatible since the block hash depends on the proof block proof. To remove this potential for a proof production race condition, proof generation should, in test mode, be deterministic.

Sword-Smith:
Ensure deterministic proofs in tests by setting the RNG seed used by the Triton VM prover. This breaks zero-knowledge of the proofs used for tests. This is motivated by the block hash depending on the block proof meaning that if two different machines start a test simultaneously, they would otherwise not generate the same proofs and, worse, will disagree on block hashes and mutator sets if these proofs are used for blocks.

This closes #657.

### DIFF
--- a/neptune-core/src/models/proof_abstractions/tasm/program.rs
+++ b/neptune-core/src/models/proof_abstractions/tasm/program.rs
@@ -248,12 +248,10 @@ pub mod tests {
         /// the subset of rust that the tasm-lang compiler understands. To run this
         /// program, call [`Self::run_rust`], which spawns a new thread, boots the
         /// environment, and executes the program.
-        #[cfg(test)]
         fn source(&self);
 
         /// Run the source program natively in rust, but with the emulated TritonVM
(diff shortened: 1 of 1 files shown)


aszepieniec (approved):
LGTM; LGTM!

Foremost: great catch! Sounds like a very confusing behavior to track down.

In other circumstances I might object to a name as long as `produce_non_zk_deterministic_proof_for_test` but in this case 

 - you are sacrificing ZK danger danger danger!
 - do not under any circumstances use this function outside of tests

are important messages to drive home. Moreover, it's not like anybody is wasting time and screen space writing this boilerplate because they use wrappers in the exact same way they used to.

PR #659, opened 2025-08-12
Composer updates tx if no synced in mempool

Sword-Smith:
A transaction included in a block must have its merge bit set which means it must go through at least one merge.

Previously two paths to a block transaction existed:
a) Get a transaction from mempool and merge it with the coinbase transaction
b) If no synced transaction exists in the mempool, make a nop transaction and merge it with the coinbase transaction.

The PR introduces a third path to a block transaction:
c) If an unsynced transaction exists in the mempool, update the transaction and merge it with the coinbase transaction.

Since updating is usually *a lot* faster than making a new single proof-backed transaction (padded height $2^{19}$ vs $2^{21}$) this is beneficial to the composer, as it allows them to compose faster. And this change benefits transaction throughput, so it's a win-win.


neptune-core/src/models/state/mempool.rs
```diff
@@ -272,6 +272,11 @@ impl Mempool {
                 continue;
             }
 
+            // Transactions with no inputs cannot be updated.
+            if candidate.transaction.kernel.inputs.is_empty() {
+                continue;
+            }
+
```
aszepieniec, line 279: Is this true? Transactions with no inputs still fix the mutator set hash, and that can be updated, no?

Sword-Smith: https://github.com/Neptune-Crypto/neptune-core/blob/2273b9a0d21b52b1249a12625775c2d75a8eb651/neptune-core/src/models/blockchain/transaction/validity/tasm/single_proof/update_branch.rs#L515-L525

:)

PR #663, opened 2025-08-18
docs: Update to match rebooted network

aszepieniec:


docs/src/consensus/block.md
```diff
@@ -24,7 +24,7 @@ Besides the kernel, blocks also contain proofs. The block proof is a STARK proof
 
 ## Validity
 
-**Note:** this section describes the validity rules for blocks at some future point when we have succinctness, not the current validity rules (although there is a significant overlap).
+**Note:** this section describes two distinct things. First is the validity rules for blocks on the presently deployed network. Second is the validity rules as they are intended to be on some future fork, for instance the one that introduces succinctness.
```
Sword-Smith: I don't like the use of the work "network" in this context, as it sounds like a reset of the monetary distribution which we don't intend. Can the word "network" be replaced by "consensus rules"?

docs/src/consensus/block.md
```diff
@@ -37,24 +37,40 @@ The genesis block is hardcoded in the source code, see `genesis_block` in `block
...
- - ***f)*** there is an ancestor block `luca` of the current block such that for each uncle block `uncle`
-   - `uncle` is valid
-   - `luca` is an ancestor of `uncle`
-   - neither `luca` nor any of the blocks between `luca` and the current block list `uncle` as an uncle block
- - ***g)*** the `version` matches that of its predecessor or is member of a predefined list of exceptions
- - ***h)*** the `height` is one greater than that of its predecessor
- - ***i)*** the `timestamp` is greater than that of its predecssor
- - ***j)*** the network statistics trackers are updated correctly
- - ***k)*** the variable network parameters are updated correctly.
+A block is incrementally valid, relative to a predecessor block, iff (all of):
```
Sword-Smith, line 42: What does "incrementally valid" mean? Does it mean that its validity depends on its parent's validity?

docs/src/consensus/block.md
```diff
@@ -37,24 +37,40 @@ The genesis block is hardcoded in the source code, see `genesis_block` in `block
...
+     - f) The current block's cumulative proof-of-work number equals that of the predecessor plus the predecessor's difficulty.
+     - g) The block timestamp is not set further in the future, relative to the host machine's clock, than the futuredating limit, which is 5 minutes.
+ 1. The block's stand-alone non-transaction data is correct, specifically (all of):
+     - a) The block appendix contains the expected claims. At present, the list of expected claims contains only one item: the claim that the block's transaction is valid.
+     - b) The block appendix does not contain too many claims. The limit is 500.
+     - c) The block proof is of the right type, namely `SingleProof`. (This variant lives on `BlockProof` and is distinct from the like-named variant living on `TransactionProof`.) The alternatives are `Genesis` and `Invalid`.
+     - d) The proof passes Triton VM verification. The input for the claim is the block mast hash; the output is all appendices; and the program is the `BlockProgram`. The `BlockProgram` merely proves the integral verification of all claims in the appendix.
```
Sword-Smith: The output is the hash of all appendix claims.

docs/src/consensus/block.md
```diff
@@ -37,24 +37,40 @@ The genesis block is hardcoded in the source code, see `genesis_block` in `block
...
+     - e) The block does not exceed the maximum size, which is set to 1'000'000 `BFieldElement`s, or 8 MB.
+ 2. The block's transaction is correct, specifically (all of):
+     - a) The removal records can be *unpacked*. Phrased differently, failure to unpack the removal records results in a format error.
+     - b) All removal records must be *removable* from the mutator set as it was after the predecessor block.
+     - c) All removal records must be unique.
```
Sword-Smith: It's the absolute index sets of the removal records that must be unique.

docs/src/consensus/block.md
```diff
@@ -37,24 +37,40 @@ The genesis block is hardcoded in the source code, see `genesis_block` in `block
...
+     - d) The transaction, along with the block's guesser fee addition records, gives rise to a valid mutator set update. This step can fail, for instance, if the transaction fee is negative.
+     - e) This mutator set update can be applied to the mutator set as it was after the previous block.
+     - f) The transaction timestamp does not exceed the block timestamp.
+     - g) The coinbase amount does not exceed the [block subsidy](mining) for this height.
+     - h) The coinbase amoune is not negative.
+     - i) The fee must not be negative.
+     - j) The number of inputs is not too large. The limit is set to 16384.
+     - k) The number of outputs is not too large. The limit is set to 16384.
+     - l) The number of announcements is not too large. The limit is set to 16384.
```
Sword-Smith, line 70: Love that you follow the rule names defined in `BlockValidationError` here!

docs/src/consensus/mining.md
```diff
@@ -0,0 +1,56 @@
...
+
+The [Tip5](https://eprint.iacr.org/2023/107) hash function is used internally.
+
+### Validating
+
+In addition to verifying the inequality between the block hash and the target, validating proof-of-work also involves verifying the authentication paths embedded in the proof-of-work data. Phrased differently, the guesser *must* produce a valid root and authentication paths in order for his block to be accepted.
+
+## Block Rewards
+
+In the beginning of Neptune Cash's life, every block is allowed to mint a certain number of Neptune Coins. This number is known as the *block subsidy*. The initial subsidy is set to `INITIAL_BLOCK_SUBSIDY = 128`. This subsidy is halved automatically every `BLOCKS_PER_GENERATION = 160815` blocks , which corresponds to approximately three years. On the rebooted network, the first generation consists of only 39505 blocks, accounting for the 21310 blocks that were mined on the legacy network.
```
Sword-Smith: not "39505" but "139505".

docs/src/consensus/succinctness.md
```diff
@@ -0,0 +1,9 @@
+# Succinctness
+
+Succinctness is the property of a blockchain that allows participants to synchronize to the netwrok with negligible resources as a function of the number of historical blocks and transactions on the network. On conventional blockchain networks such as Bitcoin, synchronizing trustlessly requires replaying all of history in order to verify it, and downloading it first.
```
Sword-Smith: typo: "netwrok" -> "network".

Sword-Smith (approved):

docs/src/consensus/block.md
aszepieniec: Good catch! To clarify, the output is the concatenation of claim hashes for all claims in the appendix.

docs/src/consensus/block.md
aszepieniec: My first reaction was, "that's implied." But it's not!

Specifically, you could have distinct removal records that have the same absolute index set. That's disallowed.

So: good catch!

Issue #649, opened 2025-08-07
mempool: Don't throw away merge inputs on successful merge


Sword-Smith:
Imagine two single proof-backed transactions A and B. They are merged into transaction C. After the successful merge a block that mines transaction A comes along. With the current mempool behavior, transaction B will be forgotten as it was thrown out of the mempool when C was received since it has overlapping inputs with C.

Mempool should store such merge inputs in a separate field `merge_inputs` which can then (potentially) be populated on `Mempool::insert` and be cleared in `Mempool::update_with_block`. The transactions in `merge_inputs` should then be checked for overlap with the newly received block and other transactions in the mempool, and if some of the transactions in `merge_inputs` are still valid and don't conflict with other mempool transactions, they should be used to populate the mempool.

PR #664, opened 2025-08-18
Thv/preserve merge inputs in mempool

Sword-Smith:
Cf. #649

### DIFF
--- a/neptune-core/src/models/state/mempool.rs
+++ b/neptune-core/src/models/state/mempool.rs
@@ -148,11 +148,13 @@ pub struct Mempool {
     max_total_size: usize,
 
     /// Contains transactions, with a mapping from transaction ID to
-    /// transaction.
+    /// transaction. Contains all transactions considered to be "in the
+    /// mempool".
     tx_dictionary: HashMap<TransactionKernelId, MempoolTransaction>,
 
     /// Allows the mempool to report transactions sorted by [`FeeDensity`] in
-    /// both descending and ascending order.
+    /// both descending and ascending order. Contains all transactions
+    /// considered to be "in the mempool".
     // This is relatively small compared to `tx_dictionary`
     #[get_size(ignore)]
     fee_densities: DoublePriorityQueue<TransactionKernelId, FeeDensity>,
@@ -231,55 +233,59 @@ impl Mempool {
         Ok(())
     }
 
-    /// Check if mempool contains the specified transaction with a higher
-    /// proof quality.
+    /// Check if mempool will accept a transaction for insertion.
     ///
     /// Returns true if the new transaction is either not known, or if it is
     /// known but has a higher proof quality than the one already in the
     /// mempool. Synced transactions (with up-to-date mutator sets) are
     /// considered of higher quality than unsynced transactions.
-    pub(crate) fn new_transaction_has_higher_proof_quality(
+    ///
+    /// Even though this function returns true, a transaction might still be
+    /// rejected for insertion if the mempool is full *and* the transaction has
+    /// a lower fee density that all transactions in the mempool.
+    pub(crate) fn accept_transaction(
         &self,
         new_tx_txid: TransactionKernelId,
         new_tx_proof_quality: TransactionProofQuality,
         new_tx_mutator_set_hash: Digest,
     ) -> bool {
-        if let Some(existing_tx) = self.tx_dictionary.get(&new_tx_txid) {
-            match existing_tx.transaction.proof.proof_quality() {
-                Ok(mempool_proof_quality) => {
-                    if mempool_proof_quality > new_tx_proof_quality {
-                        // New tx has lower proof quality.
-                        false
-                    } else if mempool_proof_quality == new_tx_proof_quality {
-                        // New tx has same proof quality. Check if new tx
-                        // represents a valid mutator set, if it does, return
-                        // true as the new transaction is more likely to be
-                        // included in a block when it's synced.
-                        existing_tx.transaction.kernel.mutator_set_hash != self.tip_mutator_set_hash
-                            && new_tx_mutator_set_hash == self.tip_mutator_set_hash
-                    } else {
-                        // New tx has higher proof quality.
-                        true
-                    }
-                }
-                Err(_) => {
-                    // Any proof quality is better than none.
-                    // This would indicate that this client has a transaction with
-                    // e.g. primitive witness in mempool and now the same transaction
-                    // with an associated proof is queried. That probably shouldn't
-                    // happen. Only if two nodes share the same secret key can
-                    // this happen, in which case, we want to accept the new
-                    // transaction, so we return true here.
-                    error!(
-                        "Failed to read proof quality for tx in mempool. txid: {}",
-                        new_tx_txid
-                    );
+        let Some(existing_tx) = self.tx_dictionary.get(&new_tx_txid) else {
+            // Transaction is not in mempool. Is it in the cache of conflicting
+            // transactions?
+            return !self.merge_input_cache.contains(&new_tx_txid);
+        };
+
+        match existing_tx.transaction.proof.proof_quality() {
+            Ok(mempool_proof_quality) => {
+                if mempool_proof_quality > new_tx_proof_quality {
+                    // New tx has lower proof quality.
+                    false
+                } else if mempool_proof_quality == new_tx_proof_quality {
+                    // New tx has same proof quality. Check if new tx
+                    // represents a valid mutator set, if it does, return
+                    // true as the new transaction is more likely to be
+                    // included in a block when it's synced.
+                    existing_tx.transaction.kernel.mutator_set_hash != self.tip_mutator_set_hash
+                        && new_tx_mutator_set_hash == self.tip_mutator_set_hash
+                } else {
+                    // New tx has higher proof quality.
                     true
                 }
             }
-        } else {
-            // Return true if transaction is not known to the mempool
-            true
+            Err(_) => {
+                // Any proof quality is better than none.
+                // This would indicate that this client has a transaction with
+                // e.g. primitive witness in mempool and now the same transaction
+                // with an associated proof is queried. That probably shouldn't
+                // happen. Only if two nodes share the same secret key can
+                // this happen, in which case, we want to accept the new
+                // transaction, so we return true here.
+                error!(
+                    "Failed to read proof quality for tx in mempool. txid: {}",
+                    new_tx_txid
+                );
+                true
+            }
         }
     }
 
@@ -1043,7 +1049,7 @@ impl Mempool {
                 self.merge_input_cache.get_size() * MAX_SIZE_OF_CACHE_FACTOR > (*self).get_size();
             if dominated_by_cache {
                 assert!(
-                    self.merge_input_cache.pop_front().is_some(),
+                    self.merge_input_cache.pop_oldest().is_some(),
                     "Dominated by cache but cannot remove element"
                 );
             } else {
@@ -2005,12 +2011,13 @@ mod tests {
         );
 
         let (((left, right), merged), _) = merge_tx_triplet(consensus_rule_set).await;
-        mempool.insert(left, UpgradePriority::Irrelevant);
-        mempool.insert(right, UpgradePriority::Irrelevant);
+        mempool.insert(left.clone(), UpgradePriority::Irrelevant);
+        mempool.insert(right.clone(), UpgradePriority::Irrelevant);
         assert_eq!(2, mempool.len());
 
         // mock that tip's mutator set hash matches that of transactions
-        mempool.tip_mutator_set_hash = merged.kernel.mutator_set_hash;
+        let tx_mutator_set_hash = merged.kernel.mutator_set_hash;
+        mempool.tip_mutator_set_hash = tx_mutator_set_hash;
 
         // Verify that `most_dense_single_proof_pair` returns expected value
         // now that two single proofs are in the mempool.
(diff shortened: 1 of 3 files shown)


neptune-core/src/models/state/mempool.rs
```diff
@@ -121,17 +123,28 @@ struct MempoolTransaction {
...
+/// considered a part of the mempool as they will not e.g. be returned for block
+/// construction. The cache is only used to avoid dropping transaction a if b is
+/// mined instead of c. See [`MergeInputCache`] for a more detailed explanation.
+///
 /// The mempool returns a list of events which should be handled by associated
 /// wallets to see unconfirmed balance updates. So all functions that can
 /// return events should be invoked from a context where listeners (like
 /// wallets) can be informed.
 #[derive(Debug, GetSize)]
+#[cfg_attr(test, derive(Clone))] // *never* use Clone outside of tests
```
aszepieniec: This comment is saying the same thing as the code. Maybe motivate *why* `Clone` should not be used outside of tests? Otherwise the next programmer will come along and think, "circumstances have changed, this commandment must be out of date". With the motivation, they can reproduce the conclusion and, if circumstances really have changed, reject it with justification.

neptune-core/src/models/state/mempool.rs
```diff
@@ -158,6 +171,14 @@ pub struct Mempool {
     /// self-initiated single-proof backed transactions should be updated when
     /// a new block is processed.
     tx_proving_capability: TxProvingCapability,
+
+    /// A list of single-proof backed transactions that were removed from the
+    /// mempool because they were inputs to a merge. So they are not in the
+    /// mempool because they conflict with another transaction there. When a
+    /// new block comes in, however, some of these transactions may become
+    /// "unconflicted" again. This list can only grow when [`Self::insert`] is
+    /// called and can shrink when [`Self::update_with_block`] is called.
```
aszepieniec, line 186: It's rather confusing to describe a field of `Mempool` as containing things that are not in the mempool anymore. Is there another choice of words that can avoid this confusion?

neptune-core/src/models/state/mempool.rs
```diff
@@ -458,6 +469,75 @@ impl Mempool {
         conflict_txs_in_mempool
     }
 
+    /// Returns `true` iff the new transaction is a merge of two transactions
+    /// with the existing mempool transaction as one of its inputs.
+    fn preserve_merge_input(
+        new_tx: &Transaction,
+        existing_tx: &Transaction,
+    ) -> Option<(TransactionKernel, NeptuneProof)> {
```
aszepieniec: This is a rather generic description and function signature. Shouldn't this function live on `Transaction` instead?

neptune-core/src/models/state/mempool.rs
```diff
@@ -458,6 +469,75 @@ impl Mempool {
...
+            .kernel
+            .inputs
+            .iter()
+            .map(|x| x.absolute_indices)
+            .collect();
+        for old_tx_input in existing_tx.kernel.inputs.iter() {
+            if !new_txs_inputs.contains(&old_tx_input.absolute_indices) {
+                return None;
+            }
+        }
```
aszepieniec: What about checking that the existing outputs are contained in the new outputs, and same with announcements?

neptune-core/src/models/state/mempool.rs
aszepieniec: Also, the doc string does not match with the function signature: the function does not return a boolean; it returns an option.

While I appreciate the convenience of having a function return the thing you're interested in if it exists over a function that just returns a boolean, I think that pattern is overkill here. For starters, it's not clear if you are returning the new or the existing transaction. Moreover, you might need a field or feature of the transaction and in this case cloning it is unnecessary.

neptune-core/src/models/state/mempool.rs
```diff
@@ -458,6 +469,75 @@ impl Mempool {
...
+        {
+            return None;
+        }
+
+        let TransactionProof::SingleProof(proof_existing_tx) = &existing_tx.proof else {
+            panic!("Transaction proof must be of type single proof");
+        };
+
+        Some((existing_tx.kernel.clone(), proof_existing_tx.to_owned()))
+    }
```
aszepieniec: It seems to me that two transactions with the same inputs and different outputs (but the same number of them) will pass this test when it shouldn't.

neptune-core/src/models/state/mempool.rs
```diff
@@ -560,9 +640,28 @@ impl Mempool {
...
+
+                        // Conditionally store existing transaction in conflict
+                        // cache.
+                        if let Some((old_kernel, old_proof)) =
+                            Self::preserve_merge_input(&new_tx.transaction, removed)
+                        {
+                            let upgrade_priority = self
+                                .upgrade_priorities
+                                .get(&conflicting_txid)
+                                .map(|x| *x.1)
```
aszepieniec, line 619:
```suggestion
                                .map(|(_,priority)| *priority)
```

`x.1` is rather opaque.

neptune-core/src/models/state/mempool.rs
```diff
@@ -560,9 +640,28 @@ impl Mempool {
...
+            let should_replace_conflict = new_tx_has_higher_proof_quality || better_fee_density;
+            if should_replace_conflict {
                 for conflicting_txid in conflicts {
                     if let Some(e) = self.remove(conflicting_txid) {
+                        let MempoolEvent::RemoveTx(removed) = &e else {
+                            panic!("remove must return remove event");
+                        };
+
+                        // Conditionally store existing transaction in conflict
+                        // cache.
```
aszepieniec: Is the conflict cache different from, or the same thing as, the merge input cache?

neptune-core/src/models/state/mempool.rs
```diff
@@ -827,15 +926,14 @@ impl Mempool {
         // contained by) SWBF indices coming from the block transaction. If they
         // are not disjoint, then remove the transaction from the mempool, as
         // it is now a double-spending transaction.
-        let swbf_index_set_union: HashSet<_> = new_block
+        let block_bf_set_union: HashSet<_> = new_block
```
aszepieniec, line 896: This name change is an improvement on one metric and a degradation on another.

 - `+`: clarify origin.
 - `-`: "`bf`" is a new and unexplained acronym; even when expanded to "bloom filter", what is a "bloom filter set"?

Suggestion: `block_swbf_index_set_union`.

neptune-core/src/models/state/mempool.rs
```diff
@@ -857,6 +955,27 @@ impl Mempool {
             events.extend(removed);
         }
 
+        // Restore transactions from blocks. Do this prior to the collection of
+        // update jobs since we migth restore a transaction that we need to
```
aszepieniec, line 926:
```suggestion
        // update jobs since we might restore a transaction that we need to
```

neptune-core/src/models/state/mempool.rs
```diff
@@ -857,6 +955,27 @@ impl Mempool {
...
+        // return as an update job, in case one of our own transactions got
```
aszepieniec, line 927:
```suggestion
        // return as an update job. This can happen, for instance, if one of our own transactions got
```

neptune-core/src/models/state/mempool.rs
```diff
@@ -857,6 +955,27 @@ impl Mempool {
             events.extend(removed);
         }
 
+        // Restore transactions from blocks. Do this prior to the collection of
```
aszepieniec, line 925: The sentence "Restore transactions from blocks." seems to describe something different from what's going on. Based on this sentence, I would expect:
 - we used to have transactions
 - for some reason they got deleted or corrupted
 - they still live on the blocks though
 - so we get them from there.

neptune-core/src/models/state/mempool.rs
```diff
@@ -934,10 +1053,10 @@ impl Mempool {
...
         }
 
-        // Maintaining the mutator set data could have increased the size of the
-        // transactions in the mempool. So we should shrink it to max size after
-        // applying the block.
-        self.shrink_to_max_size();
+        {
+            let removed = self.shrink_to_max_size();
+            events.extend(removed);
+        }
```
aszepieniec, line 1026:
```suggestion
            let removed = self.shrink_to_max_size();
            events.extend(removed);
```

If there is a reason to keep the braces, how about articulating that reason in a comment?

neptune-core/src/models/state/mempool.rs
```diff
@@ -1254,6 +1392,67 @@ mod tests {
         }
     }
 
+    /// Returns three transactions: Two transactions that are input to the
+    /// transaction-merge function, and the resulting merged transaction. Also
+    /// returns the mutator set these transactions are synced against.
+    async fn merge_tx_triplet(
```
aszepieniec, line 1368: The function name is ambiguous. It could be (imperative verb, object) just as well as (object).

(Also, twin is to pair as triplet is to triple.)

```suggestion
    async fn deterministic_merge_tx_triple(
```

neptune-core/src/models/state/mempool/merge_input_cache.rs
```diff
@@ -0,0 +1,95 @@
...
+#[derive(Debug, Clone, GetSize)]
+pub(super) struct MergeInputCacheElement {
+    pub(super) tx_kernel: TransactionKernel,
+    pub(super) single_proof: NeptuneProof,
+    pub(super) upgrade_priority: UpgradePriority,
+}
+
+/// The mempools cache of transactions that conflict with transactions in the
+/// mempool but might still be mined in the future in case the merged
+/// transaction that kicked out the elements contained herein is not mined.
```
aszepieniec, line 22: The mempool has a cache whose elements not part of the mempool.

Let's try to find a better way to describe what's going on.

neptune-core/src/models/state/mempool/merge_input_cache.rs
```diff
@@ -0,0 +1,95 @@
...
+/// transactions a, b, and c are simulataneously mineable. Because of this
+/// conflict, the insertion of c into the mempool removes transaction a and b.
+/// But if a new block is mined containing transaction a, then transaction b
+/// can still be mined (after an update). So without this cache, transaction b
+/// has now dissapeared from the mempool. The solution is to keep a and b
+/// around until one of their inputs are mined since both a and b might both be
+/// mined in case the winning composer ignored the c transaction, or the c
+/// transaction came in too late.
+#[derive(Debug, GetSize, Default)]
+#[cfg_attr(test, derive(Clone))] // *never* use Clone outside of tests
```
aszepieniec: The comment repeats the code but with different words. Instead, motivate.

neptune-core/src/models/state/mempool/merge_input_cache.rs
```diff
@@ -0,0 +1,95 @@
...
+pub(super) struct MergeInputCache {
+    elements: VecDeque<MergeInputCacheElement>,
+}
+
+impl MergeInputCache {
+    #[cfg(test)]
+    pub(super) fn len(&self) -> usize {
+        self.elements.len()
+    }
```
aszepieniec: Let's avoid mixing test and non-test code where-ever we can.

neptune-core/src/models/state/mempool/merge_input_cache.rs
```diff
@@ -0,0 +1,95 @@
...
+        tx_kernel: TransactionKernel,
+        single_proof: NeptuneProof,
+        upgrade_priority: UpgradePriority,
+    ) {
+        self.elements.push_back(MergeInputCacheElement {
+            tx_kernel,
+            single_proof,
+            upgrade_priority,
+        });
+    }
```
aszepieniec: How about implementing `Deref<Target=VecDeque<MergeInputCacheElement>> for MergeInputCache`? And `DerefMut` too obviously.

neptune-core/src/models/state/mempool/merge_input_cache.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    /// Update the merge cache with a new block. Returns the transactions that
+    /// can now go back into the mempool but does not filter for internal
+    /// conflicts. So returned elements may be mutually incompatible but are
+    /// guaranteed to be individually compatible with the new block. Clears the
+    /// cache.
+    pub(super) fn update_with_block(
+        &mut self,
+        block_bf_set_union: &HashSet<u128>,
```
aszepieniec, line 101:
```suggestion
        block_swbf_index_set_union: &HashSet<u128>,
```

Sword-Smith: I think "Bloom filter" is fine nomenclature. We don't need the qualifier "sliding window" imo.

Sword-Smith:
I've addressed many of your concerns. This PR solves a *concrete* problem that valid transactions are dropped from the mempool which means transactions have to be issued again. I would be happy to consider a PR for you where the conflicts are somehow marked on the transactions and not put into a separate cache. I tried but found it to be a very complicated task because of the events that you have to return (since the conflicting transactions should e.g. not count towards unconfirmed balances) and how you'd remove those tags afterwards. 

What I've built is something that performs quite well and works for the observed problem. I'm not going to let the perfect be the enemy of the good and allow this PR to be stalled indefinitely.

I agree that it's important to have a good definition of what "in the mempool" means. I've tried to provide that in the updated documentation. In short, from a data perspective, it's all the transactions in the `tx_dictionary` field.

Sword-Smith:
With regards to transaction tracking, there is more work to be done. A transaction initiator should have a way to ask if the input/output/announcement sets of their unconfirmed transactions are about to be mined, or they have been dropped from the mempool. The current architecture, both with and without this PR, can answer these questions by considering what's "in the mempool" (i.e. you don't have to look in `merge_input_cache`), since transactions will only end up in the cache if there is a transaction in the mempool that is the result of the merger of the initiated transactions.

neptune-core/src/models/state/mempool/merge_input_cache.rs
aszepieniec: The point remains that this variable does not contain a union of sets of block bloom filters, but that is what you read from the name.

aszepieniec:
Thanks for the clarification.

neptune-core/src/models/state/mempool/merge_input_cache.rs
Sword-Smith: Good point.

neptune-core/src/models/state/mempool.rs
Sword-Smith: It should have said "Restore transactions from merge input". Fixing.

neptune-core/src/models/state/mempool.rs
Sword-Smith: Shadowing.

PR #670, opened 2025-08-26
(Un)Freeze and Set Tip

aszepieniec:
This PR lifts three functionalities, along with endpoints, from branch `redeem` and/or repo `neptune-legacy`.

 1. Freeze the state update mechanism. Ignore incoming blocks. `neptune-cli freeze`
 2. Set the current tip to a stored block. `neptune-cli set-tip ecfae7[...]`
 3. Unfreeze the state update mechanism. Resume updates induced by announced blocks. `neptune-cli unfreeze`.

These features were originally developed for the redemption process. Specifically, to set fix the state to a given block in order to produce and verify UTXO redemption claims -- and to avoid interference from the state update mechanism when peers announce blocks.

They are considered for integration into master separately and for separate reasons.

1. Freeze and unfreeze: if you want to build a big transaction that spans many `Merge`rs and perhaps even many `Raise`s, then you need finish that operation before the next block arrives because otherwise the result will be out of date instant it is finished -- if you even can finish because you will end up applying `Update`s along the way. With state freezing and unfreezing, you can focus on building one transaction without worrying about state updates, and then synchronize it to the current state of the blockchain with one `Update`.
2. Set-tip: the 5-hour gap between blocks observed on 2025-08-25 shows that it is possible for the blockchain to stall when the guessing power and composing power are on separate forks of a block race. One way to resolve such a situation is to manually set the tip to the latest universal common ancestor and resync from there.


neptune-cli/src/main.rs
```diff
@@ -835,12 +856,12 @@ async fn main() -> Result<()> {
                 .block_digest(ctx, token, BlockSelector::Tip)
                 .await??
                 .unwrap_or_default();
-            println!("{head_hash}");
+            println!("{} / {:x}", head_hash, head_hash);
```
Sword-Smith: Is there a reason to print the hash in two formats? Let's just go with hex-string displaying, no?

neptune-cli/src/main.rs
```diff
@@ -835,12 +856,12 @@ async fn main() -> Result<()> {
...
         }
         Command::LatestTipDigests { n } => {
             let head_hashes = client.latest_tip_digests(ctx, token, n).await??;
             for hash in head_hashes {
-                println!("{hash}");
+                println!("{hash} / {hash:x}");
```
Sword-Smith: same

neptune-cli/src/parser/hex_digest.rs
```diff
@@ -0,0 +1,51 @@
...
+                .map_err(|_| DigestParseError::InvalidHexString(byte_str.to_owned()))?;
+        }
+
+        let digest = HexDigest(
+            Digest::try_from(arr).map_err(|_| DigestParseError::NonCanonicalRepresentation)?,
+        );
+
+        Ok(digest)
+    }
+}
```
Sword-Smith: Can we have a small happy-path unit test of this?

neptune-core/src/models/blockchain/type_scripts/native_currency_amount.rs
```diff
@@ -1090,4 +1091,16 @@ pub(crate) mod tests {
...
+        type Output = Self;
+
+        /// Panics on overflow. (Or, technically, if `isize` is a bigger type
+        /// than `i128` and casting from the former to the latter fails.)
+        fn mul(self, rhs: isize) -> Self::Output {
+            let cofactor = i128::try_from(rhs).unwrap();
+            let new_internal = self.0.checked_mul(cofactor).unwrap();
+            Self(new_internal)
+        }
+    }
```
Sword-Smith: We already have `scalar_mul` on `NativeCurrencyAmount`. Can't you use that instead?

neptune-cli/src/main.rs
aszepieniec: I did not want to step on any toes.

If it were up to me, all printings and loggings of digests would use the hex format.

Sword-Smith: You're the cryptographer. Feel free to make that change, either globally or as-you-go. Watch out with global search-and-replace though. They often break stuff.

aszepieniec: I doubt 8e47479e3c10af525849356fd3ab16ec8db17760 catches all occurrences but it certainly catches the vast majority.

neptune-cli/src/parser/hex_digest.rs
aszepieniec: Added in 9a166bc2.

neptune-core/src/models/blockchain/type_scripts/native_currency_amount.rs
aszepieniec: Do you mind motivating why `LIQUID_BLOCK_SUBSIDY.scalar_mul(3)` is preferable to `LIQUID_BLOCK_SUBSIDY * 3`?

Regardless, I applied the requested change in 83df205c because I suspect your preference is stronger than mine.

aszepieniec:
> Peer loop should ignore certain messages when the state is frozen. Otherwise you'll end up in infinite p2p block-related interactions. See PR #527 for how I achieve this.

I'm afraid there is not enough information in this comment or the link for me to determine what you mean. Please elaborate.

Can you describe these infinite p2p block-related interactions? In your commit I do see two functions `pub fn ignore_on_paused_state_updates(&self) -> bool` and how they are used. So I see *how* you achieve it; but not *what* you are achieving.

dan-da:
This also seems like it could be useful in conjunction with regtest mode and mine-blocks-to-wallet RPC for simulating re-orgs.

aszepieniec:
Rebased to use @Sword-Smith's pausing commit instead.

neptune-core/src/models/blockchain/type_scripts/native_currency_amount.rs
Sword-Smith: I'm honestly fine with either. I just prefer that we only have one option: either `scalar_mul` or this new option. Not both.

Sword-Smith (approved):

Sword-Smith:
LGTM.

Nice to have that isn't included here: Allow user to set tip based on height. Fail if multiple blocks at the same height exists.

PR #674, opened 2025-08-28
fix: ensure block data written before db

### LINKED ISSUE #471: core-dump with SIGBUS error in ArchivalState::get_block_from_block_record()
This occurred when restarting neptune-core after a power outage.   Almost certainly a blockchain file is corrupted.   Ideally we would handle it more gracefully than with a core-dump though.

1. Logs are not very informative.  (note RUST_LOG=debug)

```
2025-02-27T01:36:50.25544553Z  INFO ThreadId(01) neptune_cash: Got wallet state.
2025-02-27T01:36:50.256797527Z  INFO ThreadId(01) neptune_cash: Got block index database
2025-02-27T01:36:50.258092128Z  INFO ThreadId(01) neptune_cash: Got peer database
2025-02-27T01:36:50.259672771Z  INFO ThreadId(01) neptune_cash: Got archival mutator set
2025-02-27T01:36:50.261025302Z  INFO ThreadId(01) neptune_cash: Got archival block MMR
Bus error (core dumped)
```

Suggestion:  in addition to logging "Got ... X" after each action, we should also log "Attempting to ... X" before the action.   So the log would at least show what was being attempted.

2. Ran under GDB, to get a stack trace where the core dump occurs:

```
Thread 58 "tokio-runtime-w" received signal SIGBUS, Bus error.
[Switching to Thread 0x7fffd4a006c0 (LWP 4050)]
0x0000555555c9117f in serde::de::SeqAccess::next_element ()
(gdb) where
#0  0x0000555555c9117f in serde::de::SeqAccess::next_element ()
#1  0x0000555555c9db2f in <&mut bincode::de::Deserializer<R,O> as serde::de::Deserializer>::deserialize_struct ()
#2  0x0000555555c9ede3 in <&mut bincode::de::Deserializer<R,O> as serde::de::Deserializer>::deserialize_struct ()
#3  0x0000555555ca0326 in <&mut bincode::de::Deserializer<R,O> as serde::de::Deserializer>::deserialize_struct ()
#4  0x0000555555bbf917 in neptune_cash::models::state::archival_state::ArchivalState::get_block_from_block_record::{{closure}}::{{closure}} ()
#5  0x0000555555a188b7 in <tracing::instrument::Instrumented<T> as core::future::future::Future>::poll ()
#6  0x0000555555bdf946 in tokio::runtime::task::core::Core<T,S>::poll ()
#7  0x0000555555adcd9d in tokio::runtime::task::harness::Harness<T,S>::poll ()
#8  0x00005555560db0c6 in tokio::runtime::blocking::pool::Inner::run ()
#9  0x00005555561091b9 in std::sys::backtrace::__rust_begin_short_backtrace ()
#10 0x000055555610a83a in core::ops::function::FnOnce::call_once{{vtable.shim}} ()
#11 0x00005555565c534b in std::sys::pal::unix::thread::Thread::new::thread_start ()
#12 0x00007ffff7c9caa4 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:447
#13 0x00007ffff7d29c3c in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78
```

SIGBUS can indicate memory corruption or mmap file corruption.  In our case, ArchivalState is reading from mmap'd file(s).  I haven't dug into it, but a corrupted mmap file is almost certainly the cause.

The question then becomes:  how can we handle this more gracefully?    Is there a way to detect a corrupted mmap file without causing a SIGBUS?     Or maybe its possible to handle the SIGBUS without core-dump?

dan-da:
fixes #471

The issue was that block records were being written to DB before written to blkXX.dat files on disk, which resulted in mis-match when power is lost during a call to ArchivalState::store_block().  Upon startup, a SIGBUS occurs when reading the non-existent mmap'd data for the tip block that is referred to by the DB.

In the code it appears that the blkXX.dat files were being written first. However it was not flushing/syncing, and thus the OS is free to buffer for indeterminate amount of time, and in actuality the DB data was being written first.  (note that we use fsync flag for db writes).

Fixes:

1. root cause: calls mmap.flush() after writing block to ensure it is written to disk before writing matching record to DB.

2. avoid SIGBUS: verifies we do not attempt to read past end of file. Returns error with diagnostic info if so, which gets logged.

3. fix db: for situations where the mis-match has somehow occurred, it now detects the error and rolls the tip back by one block in the block_index DB in an attempt to fix.

---------------------

notes:

1. this PR is a bit quick/dirty.  Consider it a draft.   It fixed the issue for me, but does not include any tests.  Also, it would be nicer if get_tip_from_disk() would return an enum (thiserror) error for get_tip_or_rollback() to check instead of ugly check of err msg.

2. fix (3) is kind of optional.   With fixes (1) and (2) in place, theoretically it should never be necessary.  For me it was necessary because the DB was already borked.

3. fix (3) required two starts of neptune-core.  On the first start it fixed the block_index DB but then reported a mismatch with mutator-set DB and recommended to wipe all databases (yikes), and exited.   I started it a second time and everything loaded seemingly correctly and my wallet balance appears.

4. I have not tested that the mmap.flush() actually fixes the root cause.  However, it seems correct to me, and also I showed the code to an LLM that verified the problem and the fix.

5. reviewers:  edit/modify as you please.  I won't be making further changes to the PR.   It works for me to solve a problem, so I'm contributing it, but not going to spend further time on it.

### DIFF
--- a/neptune-core/src/models/state/archival_state.rs
+++ b/neptune-core/src/models/state/archival_state.rs
@@ -411,6 +411,16 @@ impl ArchivalState {
             };
             let mut mmap: memmap2::MmapMut = mmap.make_mut().unwrap();
             mmap.deref_mut()[..].copy_from_slice(&serialized_block);
+
+            // FIX for #471
+            // https://github.com/Neptune-Crypto/neptune-core/issues/471
+            //
+            // Flush the memory-mapped pages to the physical disk.
+            // This call will block until the data is safely persisted.
+            // This ensures block data is written to the blkXX.dat file before
+            // updating the DB.  Else we can have situations where the DB
+            // references a block that does not exist on disk.
+            mmap.flush().unwrap();
         })
         .await?;
 
@@ -482,6 +492,12 @@ impl ArchivalState {
         self.block_index_db.batch_write(batch).await;
 
         Ok(())
+    }
+
+    async fn set_tip_digest(&mut self, tip_digest: Digest) {
+        let key = BlockIndexKey::BlockTipDigest;
+        let val = BlockIndexValue::BlockTipDigest(tip_digest);
+        self.block_index_db.put(key, val).await
     }
 
     /// Write a newly found block to database and to disk, without setting it as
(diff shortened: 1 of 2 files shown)


Sword-Smith:
Thanks for this fix. Apart from scrutinizing it for correctness I'd also like to ensure that it doesn't introduce a big slowdown since our initial block download and state syncing is quite slow.

aszepieniec:
Fantastic display of investigatory prowess.

KaffinPX:
Uh this feels like the CTRL+C DB corruption issue caused by "unordered flush"(One item was getting flushed while other state wasn't), so in this case fix shouldn't be flushing tip and flushing it together with block?

Issue #682, opened 2025-09-01
composing does not relay incoming block proposals to peers.


dan-da:
At the moment I have a composing node that is also setup as the only node connected to the internet.   I noticed that other peers on the subnet (connected to composing node) are not receiving any block proposals.

I believe the issue is in the code below.   favor_incoming_block_proposal_legacy() returns Err(BlockProposalRejectError::Composing) so we enter the error case and never send the BlockProposalRequest.  Thus we never receive the BlockProposal.   Thus we never notify our peers either.

It seems like a composing node should still pass along the proposals to peers, even if it doesn't choose to use them itself.

```
            PeerMessage::BlockProposalNotification(block_proposal_notification) => {
                let verdict = self
                    .global_state_lock
                    .lock_guard()
                    .await
                    .favor_incoming_block_proposal_legacy(
                        block_proposal_notification.height,
                        block_proposal_notification.guesser_fee,
                    );
                match verdict {
                    Ok(_) => {
                        peer.send(PeerMessage::BlockProposalRequest(
                            BlockProposalRequest::new(block_proposal_notification.body_mast_hash),
                        ))
                        .await?
                    }   
                    Err(reject_reason) => {
                        debug!(
                        "Rejecting notification of block proposal with guesser fee {} from peer \
                        {}. Reason:\n{reject_reason}",
                        block_proposal_notification.guesser_fee.display_n_decimals(5),
                        self.peer_address
                    )
                    }
                }

                Ok(KEEP_CONNECTION_ALIVE)
            }
```

Sword-Smith:
It was designed this way intentionally, that if you're composing you never relay other compositions, only your own. I think that's how *most* composers would want their node to behave -- but if you think it should behave another way, would it be possible to add that option via a CLI argument?

dan-da:
> If you're running a mining pool that is also composing, it's because you want to use your own compositions, I'd say

That is one use case, but we can't know everyone's reasons/uses.   And keep in mind that "your own compositions" may be much slower than the network's.

In my case I wanted to try composing on one node only in case there are some very slow blocks where my slowish hardware might have a chance to finish composing before the block is found.  But I still want the guessing nodes to see the block proposals from remote nodes.

[...]

PR #683, opened 2025-09-01
composing node relays BlockProposal to peers

dan-da:
addresses #682

## commit msg

If a composing node has peers then it should still relay a BlockProposal to them even though the proposal is not useful to the node itself.

This is particulary useful when --restrict-peers-to-list is in use.

## notes

* without this, guessing nodes will not receive any BlockProposal if relying on composing node for internet comms.
* this is quick n dirty, made for my needs.  seems to work.  It could likely be more elegant.
* there is no CLI option to enable/disable relay behavior.  that could be added as per comments in #682
* feel free to edit however suits project's needs, or close if not wanted.


Sword-Smith:
Implemented in e7c1cee150383c2ad7d5de778c2da2d968fe0628

Issue #697, opened 2025-09-13
Limit size of ProofCollection-backed transactions for propagation?


Sword-Smith:
SingleProof-backed transactions never grow above 2MB (as long as they have less than ~1000 inputs). But ProofCollection backed transaction can become very big because of how many proofs that support them. Should we introduce a limit such that they are not relayed if they are bigger than N bytes, or have more than M inputs?

skaunov:
Let me subscribe to this one... I was thinking on your message which sparkled this issue, I guess. In #701 I'm thinking to add a dedicated second instance of request-response component just for this case. It's not that bad there, as adding a component is dirt-cheap, but the limits are usually set component-wise (not message type-wise). And in the topic after the (usual message) size limit it would spread just the advertisement ("notification") of that tx instead of the thing itself, giving it out directly to peers who ask. So DoS is solved there, but the edge case still being nuisance. =( And privacy takes a hit, but it does kinda anyway for this case.

So I'd be glad to see the edge case being just eliminated. **What are the cases when it can grow abnormally?** I mean it would be cool if those cases are of the kind when off the network transfer makes sense. (Similar to off the chain UTXO notification.)

I was also considered Kademlia providing (not publishing, just signalling the peer can give the item with that digest outside of Kademlia) to assist the source of a huge tx. But *I can't find a motivation for a node in the network to assist propagating a 20-80 MB tx for someone.* =( Tbh I was struggling to imagine how miners are set to handle those: it should be an edge case for them too, so how would they account a fine fee for it: like a progressively per additional MB? Or does it target the specific kind of miners and for them it's a "gold-bearing vein" they can upgrade and collect?

Sorry for an essay here. But this led me to the thought that "tx" topic is not enough (I suspected that, but postponed the thought). And **it should be divided for different kinds of miners. Could you help me to identify** this sub-division? (Let me mention @aszepieniec too in case other are busy or miss this comment.) If these huge tx are beneficial to a group of nodes then I'd better design appropriate component(s) so they could benefit on these without interrupting others.

skaunov:
I found the size limits have enough settings to be just parameterized in Gossip component, so no components interaction to design in `libp2p` the settings would be enough for this issue.

Still, pls, **help me to sub-divide messages** into more granular topics better.

Sword-Smith:
> **What are the cases when it can grow abnormally?**

[...]

Issue #660, opened 2025-08-13
Don't share local IPs


Sword-Smith:
Local IP addresses (ones that start with e.g. 192.168) should not be shared as part of the peer discovery protocol. And machines should not attempt to connect to these local IPs when connecting to an IP received over the peer discovery protocol.

For IPv4 the ranges considered local is provided in [this Stack Overflow answer](https://stackoverflow.com/a/22479334/2574407). For IPv6 addresses, [Wikipedia shows](https://en.wikipedia.org/wiki/Private_network) which are considered local. 

Rust even has methods for this for both [IPv4](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#method.is_private) and for [IPv6](https://doc.rust-lang.org/std/net/struct.Ipv6Addr.html#method.is_unique_local). Let's use those!

Issue #514, opened 2025-03-19
Don't allow `GlobalState` read lock on failed connection attempts


Sword-Smith:
Nodes that receive a lot of failed incoming connection attempts have been seen to hang a lot. I suspect that the problem is that all the tasks that handle incoming connections hold a read lock on `GlobalState` while checking if the connection is allowed. Then when an incoming message requires a write lock, the node has to wait for all read locks to be released.

A solution would be to factor out either `NetworkingState` or `PeerMap` from `GlobalState` such that failed incoming connections would not acquire locks that prevented the chain state or mempool from being updated.

dan-da:
in another issue I suggested that we should drop `state` in this code before spawning the answer_peer task:

```
                    let state = self.global_state_lock.lock_guard().await;
                    let main_to_peer_broadcast_rx_clone: broadcast::Receiver<MainToPeerTask> = self.main_to_peer_broadcast_tx.subscribe();
                    let peer_task_to_main_tx_clone: mpsc::Sender<PeerTaskToMain> = self.peer_task_to_main_tx.clone();
                    let own_handshake_data: HandshakeData = state.get_own_handshakedata();
                    let global_state_lock = self.global_state_lock.clone(); // bump arc refcount.
                    let incoming_peer_task_handle = tokio::task::Builder::new()
                        .name("answer_peer_wrapper")
                        .spawn(async move {
                        match answer_peer(
                            stream,
                            global_state_lock,
                            peer_address,
                            main_to_peer_broadcast_rx_clone,
                            peer_task_to_main_tx_clone,
                            own_handshake_data,
                        ).await {
                            Ok(()) => (),
                            Err(err) => error!("Got error: {:?}", err),
                        }
                    })?;

```

I don't see a smoking gun that present behavior is wrong/problematic, but it might be.   `state` does get dropped at the end of the scope and the answer_peer task executes asyncronously, so both should occur.   But it definitely would be be cleaner and more certain to drop it before spawning the task, esp since it isn't even used later.

Issue #430, opened 2025-02-13
Improve Peer Discovery


aszepieniec:
I am behind a NAT. The time spent running a node is irrelevant to the number of peer connections I get. I only ever get the connections to peers supplied by the `--peers` argument.

I should be able to initiate connections to peers whom I learned about from the bootstrap peers. But for some reason that is not happening.

dan-da:
I am seeing this also.  Yeah it should be that my node (behind NAT) learns about new peers from the first (--peers) it connects to, and then connects to the new ones, and so on, up to --max-num-peers.

this should be a pretty high prio to fix.

gump1995:
中国用户连接不上。请求帮助

aszepieniec:
> 中国用户连接不上。请求帮助
> [Chinese users can't connect. Please help.]

Noted. Do you think it is because of NAT? If not, can you speculate on what the cause might be? In that case I guess it should be a separate issue.

aszepieniec:
> The time spent running a node is irrelevant to the number of peer connections I get.

This is not quite true: I noticed  I got a fourth connection after only specifying three peers through the CLI.

Still, peer discovery is worth improving. Compared to other cryptocurrencies I've tried, it is slow.

aszepieniec:
Four months later, peer discovery continues to be pretty bad. Yes we have had other, higher-priority issues to focus on, but maybe part of the reason why no external contributors decided to take this task on is because of its poor definition. So in the hope to attract external contributors' attention (and otherwise to track our own progress) here is a proposed decomposition into subtasks.

skaunov:
Am I correct it's best to start with connections relaying (hole punching). What are the other nice to begin with parts?

PR #701, opened 2025-09-20
`libp2p`

skaunov:
Basic switching to `libp2p` to see what & how it affects in the existing code.

this line of work potentially can solve 
#660 
#430 
#514 
#90
#255
#36
#35
#697

### DIFF
--- a/neptune-core/src/application/loops/main_loop/swarm_def.rs
+++ b/neptune-core/src/application/loops/main_loop/swarm_def.rs
@@ -1,6 +1,4 @@
-mod behaviour;
-
-use std::{net::Ipv4Addr, time::Duration};
+use std::time::Duration;
 
 use libp2p::{gossipsub, identify, kad::{self, store::MemoryStore}, multiaddr::Protocol, noise, ping, request_response::{self, cbor}, swarm::{NetworkBehaviour, SwarmEvent}, yamux, Multiaddr, Swarm};
 


neptune-core-cli/src/main.rs
```diff
@@ -1076,9 +1076,9 @@ async fn main() -> Result<()> {
             client.clear_all_standings(ctx, token).await??;
             println!("Cleared all standings.");
         }
-        Command::ClearStandingByIp { ip } => {
-            client.clear_standing_by_ip(ctx, token, ip).await??;
-            println!("Cleared standing of {ip}");
+        Command::ClearStandingByPeer { id } => {
+            client.clear_standing_by_ip(ctx, token, id).await??;
```
aszepieniec, line 1080: Probably a good idea to rename `clear_standing_by_ip` too.

neptune-core/src/application/config/cli_args.rs
```diff
@@ -615,15 +617,8 @@ impl Args {
...
-    pub(crate) fn accept_block_proposal_from(&self, ip_address: IpAddr) -> bool {
-        let ip_address = match ip_address {
-            IpAddr::V4(_) => ip_address,
-            IpAddr::V6(v6) => match v6.to_ipv4() {
-                Some(v4) => std::net::IpAddr::V4(v4),
-                None => ip_address,
-            },
-        };
-        self.whitelisted_composers.is_empty() || self.whitelisted_composers.contains(&ip_address)
+    pub(crate) fn accept_block_proposal_from(&self, peer: &libp2p::PeerId) -> bool {
```
aszepieniec: Shouldn't this be a `Multiaddress` instead of a `PeerId`?

neptune-core/src/application/loops/main_loop/swarm_def.rs
```diff
@@ -0,0 +1,79 @@
+mod behaviour;
+
+use std::{net::Ipv4Addr, time::Duration};
+
+use libp2p::{gossipsub, identify, kad::{self, store::MemoryStore}, multiaddr::Protocol, ping, request_response::{self, cbor}, swarm::{NetworkBehaviour, SwarmEvent}, Multiaddr};
```
aszepieniec: We prefer "item" granularity because it gives less headaches when there is a merge conflict.

neptune-core/src/application/loops/main_loop/swarm_def.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use crate::protocol::peer::PeerMessage;
+
+#[derive(libp2p::swarm::NetworkBehaviour)]
+pub(crate) struct ComposedBehaviour {
```
aszepieniec, line 8: Why is it called `ComposedBehavior`? It feels like there is a clue in the name that helps to understand the big picture, but I can't understand the clue.

neptune-core/src/application/loops/main_loop/swarm_def/behaviour.rs
aszepieniec, line 1: Missing body.

neptune-core/src/application/loops/peer_loop.rs
```diff
@@ -291,47 +246,45 @@ impl PeerLoopHandler {
                 "blocks"
             }
         );
-        let now = self.now();
+        let now = Timestamp::now();
```
aszepieniec, line 249: This change is dangerous! Because in tests, you want to be able to set the time to a given timestamp.

neptune-core/src/application/loops/peer_loop.rs
```diff
@@ -780,88 +701,74 @@ impl PeerLoopHandler {
...
-                let Some(issued_challenge) = peer_state_info.sync_challenge else {
-                    warn!("Sync challenge response was not prompted.");
-                    self.punish(NegativePeerSanction::UnexpectedSyncChallengeResponse)
-                        .await?;
-                    return Ok(KEEP_CONNECTION_ALIVE);
-                };
-
-                // Reset the challenge, regardless of the response's success.
-                peer_state_info.sync_challenge = None;
+                if let Some(issued_challenge) = peer_state_info.sync_challenge {
```
aszepieniec, line 704: Judging by the indentation and changed line count, a lot has changed in this handler. Am I looking for a change in logic, or is this just a superficial change? If the former, please draw my attention to what changed (beyond the superficial). If the latter, please motivate.

neptune-core/src/application/loops/peer_loop.rs
```diff
@@ -1378,15 +1258,14 @@ impl PeerLoopHandler {
...
-                        .await?;
+                    chan.send(PeerTaskToMain::Sanction(peer, PeerSanction::Negative(NegativePeerSanction::CannotApplyTransactionToMutatorSet))).await?;
                     return Ok(KEEP_CONNECTION_ALIVE);
                 }
 
                 let tx_timestamp = transaction.kernel.timestamp;
 
                 // 6. Ignore if transaction is too old
-                let now = self.now();
+                let now = Timestamp::now();
```
aszepieniec, line 1268: Dangerous, please avoid.

neptune-core/src/protocol/peer.rs
```diff
@@ -454,8 +456,6 @@ pub(crate) enum PeerMessage {
     SyncChallenge(SyncChallenge),
     SyncChallengeResponse(Box<SyncChallengeResponse>),
 
-    BlockProposalNotification(BlockProposalNotification),
-
```
aszepieniec, line 463: So it is no longer possible to send a `BlockProposalNotification`? That sounds like a rather dramatic and unnecessary change. How did you come to the conclusion that this deletion is the right way forward?

Even if the peer loop is modified so that this message is never sent, I am apprehensive about deleting any variant from this enum. Doesn't that break backwards compatibility?

neptune-core/src/protocol/peer.rs
```diff
@@ -471,7 +471,7 @@ pub(crate) enum PeerMessage {
     TransactionRequest(TransactionKernelId),
     PeerListRequest,
     /// (socket address, instance_id)
-    PeerListResponse(Vec<(SocketAddr, u128)>),
+    PeerListResponse(Vec<Multiaddr>),
```
aszepieniec, line 474: Same concern about breaking backwards compatibility.

neptune-core/src/state/database.rs
```diff
@@ -157,7 +158,7 @@ impl BlockIndexValue {
 
 #[derive(Clone)]
 pub struct PeerDatabases {
-    pub peer_standings: NeptuneLevelDb<IpAddr, PeerStanding>,
+    pub peer_standings: NeptuneLevelDb<PeerId, PeerStanding>,
```
aszepieniec, line 161: I wonder if there is any randomness or other connection-specific nondeterminism in `PeerId`. Because if there is, then it seems that misbehaving peers would be able to escape their punishment simply by creating a new connection.

neptune-core/src/application/config/cli_args.rs
skaunov: It's a great question I was and keep asking myself such ones, and there should be number of places I missed. Here's my logic/idea is that an user specifies the multi-addresses to the app and it gets the `PeerId` from the address (explicit or via connection) and then accepts the proposals from that over any connection path.

neptune-core/src/application/loops/main_loop/swarm_def.rs
skaunov: If you remember when I started I was adapting Substrate implementation, and after some work I understood that they like was doing from so early that they reimplemented 60% of the library (without using `derive`). Turns out that straightly composing the ready implemented `Behaviour` takes away some control but leverages the maintenance efforts the library authors do and the default / best practices they bake into it.

I mean there are few places I'd make other decisions for Neptune specifics but those are quite small choices and shifting maintenance burden to the `libp2p` out-weights those optimizations completely.

So name have a history behind but insignificant; can be well changed to any other. X)

neptune-core/src/application/loops/peer_loop.rs
skaunov: Oh, it's the lateral thing... I was trying to understand why there are so much field element timestamps and it turned out half of them should be replaced with `Instant` (those which doesn't end in blockchain).

And this particular one change signifies that I feel like there's a better way for testing too. This one is nice and working - no problem. But I'd improve/develop it. X)

neptune-core/src/application/loops/peer_loop.rs
skaunov: On the next iteration I'd dismantle it completely, but now I tried to minimally touch the thing exactly for comparing (though I did some active reading like it took some effort to understand why there's early `return`). 

That said logic is intentionally mostly the same (even when it asks to be adapted). The take-away for me are
- remove `self` as incompatible (the needed parts of it become arguments for now),
- the single max height seen which drastically reduces communication. It's about the fact that the node watches the tip(s) now with the "block" topic subscription. I actually had a question on this.

I didn't went into the logic deep enough, so I wonder how complex would it be to switch to tracking the cumulative PoW instead of height. (Pls forget about backwards compatibility for now.) I mean currently the discrepancy between the two are solved by the syncing mode I guess, and it cleverly tied to mining. But is it possible to get the blocks of competing branches via the single subscription? Am I correct something like this already happens but separated by the peer tasks? How/when should we forget a defeated branch (now I guess this happens when the peer representing it chooses to sync)?

neptune-core/src/application/loops/peer_loop.rs
skaunov: That whole thing made me thinking hard about itself! X) I ended with the realization that I'd prefer taking `now: Timestamp` as an argument, then first thing measure `Duration` since the function start, and put the sum of `Duration::as_millis` and `now` when it's used. That would cover all the needs without conditional compilation.

neptune-core/src/protocol/peer.rs
skaunov: It's the power of a subscription. Instead of `...Notification` and the whole process down the line a peer would get (or send) the block itself (or tx, or proposal) and using those Gossip-sub mechanisms (which are essentially the same just much more elaborated) avoid additional back and forth aiming to fast discovery and single transfer of the data.

neptune-core/src/state/database.rs
skaunov: Yes! I was thinking about this too! I came to the conclusion that relaying will have the same effect so it's something to accept. (Anyway IP are often dynamic.) And hence standings saved to the DB should be just for long present peers - like those seen the last time, or having the connection much longer than others (it was a very recent thought I didn't put it in the code yet).

I feel it's ok to think about standings like about storing good peers to rely upon, mostly forgetting about those to avoid.

skaunov:
On the one hand it was a little overaggressive approach to see what happen without these. On the other hand I'd rather invert the question to have a justification for those variants to exist. So it's leaning to the required end with scratching the surface on improving
> Enum `MainToPeerTask` is being changed rather significantly, and I believe in excess of the changes required to make things work with libp2p. Is the purpose of this PR to just migrate to libp2p, or to migrate to libp2p _and_ to improve the p2p protocol in other ways?
> 
> If the first, please avoid changing the p2p protocol beyond what's necessary for libp2p.
> 
> If the second, can you motivate these changes?
> 
>     * `BlockProposalNotification` -> `BlockProposal`
> 
>     * `TransactionNotification` -> `NewTransaction`
> 
>     * delete `Disconnect` and `DisconnectAll`
> 
>     * delete `DisconnectFromLongestLivedPeer`
> 
> 

For `...Notification` it's the fact that Gossip-sub is well suited precisely for that task: publish some data and manage notification of its spreading; I guess our capability to optimize this part will lag behind their development without a chance. 

As for `Disconnect...` I'd say it's too granular approach when switching to `libp2p::Swarm`. I developed a fine shutdown part for it; and for connections management there more correct API/facilities from the config options to the black-white lists component.
_________________
The info is scattered indeed. =( No single place which would have everything needed, sadly. On that path <https://docs.libp2p.io/concepts/introduction/overview/> is one of the most useful entrypoint.
> What are good resources to learn about the basic libp2p mechanisms being invoked and how they work? I found https://docs.libp2p.io/ already but there is a lot there. 

I was minding backwards compatibility as I saw you want to care about it, but at some point I concluded it's infeasible, the only way is running in parallel. =( I understand some/general motivation for the compatibility, but not entirely. **Could you make the rationale for it,** so it could be addressed by other means probably.
> I'm very concerned about breaking backwards compatibility, and the changes to `PeerMessage` strike me like they would break it. If we take this software upgrade, will nodes running the new software be able to connect to nodes running the old software? As long as the answer is "no" we have to go back to the drawing board and revise the architecture.

skaunov:
The fresh commit adds hole-punching. Up to this point it was mainly about finding something that would really prevent adding `libp2p`. Based on the discussion and gained experience I feel this **worth to continue and develop into the system.** Lack of further discussion leaves me just with a **parallel** path, *which isn't that bad and probably would win for now* anyway.

neptune-core/src/protocol/peer.rs
KaffinPX: But a notification might get pinged multiple times over same peer. Basically a "gossip" can get directed into that peer back?

skaunov: I did quite a study how that work but still limited to applying it here. So my current understanding is it doesn't guarantee the single outburst and single receive but trying to achieve that. A peer indicates to his peers the messages that he doesn't want to receive (basically those he already has). So a message he gossiped has low probability of being sent to him. If I understood you correctly (not sure what you meant by "pinged" though).

What bugs me quite badly is the fact that `impl Default` are done nicely but leans to quick networks with small messages. I guess for Neptune it will take quite some tuning to the optimal parameters. My guess we don't need the heartbeat every 1 sec, we will want history longer than 3 heartbeats, and this already are fundamentals which entails seeking optimums for all the other parameters. :shrug:

PR #702, opened 2025-09-22
Asz/681 improve mempool dashboard screen

aszepieniec:
- Remove 2 columns from Mempool table
 - Fix sign error in those two columns
 - Add UTXOs screen
 - Add RPC endpoint for UTXOs screen

### DIFF
--- a/neptune-core/src/state/wallet/wallet_state.rs
+++ b/neptune-core/src/state/wallet/wallet_state.rs
@@ -557,6 +557,8 @@ impl WalletState {
         }
     }
 
+    /// Get an iterator over (utxo, aocl_leaf_index) pairs corresponding to
+    /// own inputs into transactions that live in the mempool.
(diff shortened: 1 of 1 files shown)


neptune-core/src/state/wallet/wallet_state.rs
```diff
@@ -557,11 +557,11 @@ impl WalletState {
         }
     }
 
-    pub fn mempool_spent_utxos_iter(&self) -> impl Iterator<Item = &Utxo> {
+    pub fn mempool_spent_utxos_iter(&self) -> impl Iterator<Item = (&Utxo, &u64)> {
```
Sword-Smith, line 562: Can I get a docstring comment about what this function returns? The `u64` is a append-only commitment list leaf index, right? Just write that, that the 2nd element is an AOCL leaf index.

PR #708, opened 2025-09-24
devops: Faster compilation by lowering some `opt-level`s

Sword-Smith:
Also make use of incremental compilation in dev and test profiles.

Since we usually do not have to create Triton VM proofs when the tests are running, we might as well compile with less optimization. I personally only ever use the dev profile in my IDE, so that should compile as fast as possible, without regard to the performance of the compiler output.


Sword-Smith:
> Great!
> 
> I wonder what happens in the 0.01% of cases though, were we do generate proofs in the course of test runs. Is there a convenient command or CLI flag to override optimization settings?

You can use:
`RUSTFLAGS="-C opt-level=3" cargo test`.

PR #711, opened 2025-09-25
fix(RPC): Filter out duplicate `UiUtxo`s

### LINKED ISSUE #710: Duplicate UTXOs on neptune-dashboard
After sync, dashboard UTXO tab reports duplicates of "existing" UTXOs as "expected".

Image

aszepieniec:
The RPC call `list-utxos` returns `UiUtxo`s from three sources:
 1. the wallet database of monitored UTXOs;
 2. the wallet database of expected UTXOs;
 3. the mempool.

It is possible for the same `UiUtxo` to be present in more than one of these sources. Previously that would generate duplicate entries in the list of `UiUtxo`s returned by `list-utxos`. This commit changes that behavior by filtering out duplicates.

Furthermore, it changes the order of (2) and (3), so that if the `UiUtxo` comes from the mempool, it is not marked as "expected" when happens to also be present in the expected UTXOs database.

Lastly, the signature of mempool function `mempool_unspent_utxos_iter` is changed to accommodate this duplicate filter.

Fixes #710.

### DIFF
--- a/neptune-core/src/application/rpc/server.rs
+++ b/neptune-core/src/application/rpc/server.rs
@@ -3579,6 +3579,16 @@ impl RPC for NeptuneRPCServer {
             .await)
     }
 
+    /// Get [`UiUtxo`] from three sources:
+    /// 1) the wallet database for monitored UTXOs (these are confirmed);
+    /// 2) the mempool (these are pending); and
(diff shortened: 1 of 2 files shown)


skaunov (commented):
Having little experience with this part I see no problems here. I created #712 addressing small things which hindered me reading the change.

PR #721, opened 2025-10-01
719 dont create new address on dashboard open

### LINKED ISSUE #719: Don't create new address whenever dashboard is opened
Annoying bug that generates a new address whenever dashboard is opened, or whenever receive screen is accessed.

> Auto-generating receiving address removes control from the user, is unintuitive, and can lead to loss of funds because of how the recovery mechanism works if a wallet is restored from a seed, because of how many empty/non used addresses there are.

Sword-Smith:
Closes #719: Don't create new address/bump key derivation index every time the dashboard is opened.

### DIFF
--- a/neptune-core/src/application/rpc/server.rs
+++ b/neptune-core/src/application/rpc/server.rs
@@ -4213,7 +4213,7 @@ pub mod error {
         #[error("Cannot restore membership proofs: {0}")]
         CannotRestoreMembershipProofs(String),
 
-        #[error("Wallet key counter is zero. This value should always be positive after wallet initialization")]
+        #[error("Wallet key counter is zero. Must be positive after init")]
         WalletKeyCounterIsZero,
     }
 
--- a/neptune-dashboard/src/receive_screen.rs
+++ b/neptune-dashboard/src/receive_screen.rs
@@ -68,9 +68,8 @@ impl ReceiveScreen {
             let escalatable_event = self.escalatable_event.clone();
 
             tokio::spawn(async move {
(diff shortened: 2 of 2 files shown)


neptune-core/src/application/rpc/server.rs
```diff
@@ -345,6 +346,11 @@ pub trait RPC {
     /// ```
     async fn confirmations(token: auth::Token) -> RpcResult<Option<BlockHeight>>;
 
+    /// Return the most recently generated generation address. Does not add a
+    /// new key to the wallet, and does not change the derivation index.
+    async fn latest_generation_address(token: auth::Token)
```
dan-da: This is not a forward thinking method as it does not support any future key types.  Also, why limit to just one?  I would suggest this more flexible API instead:

```
fn latest_addresses_by_key_type(key_type: KeyType, start: usize, max: usize) -> Vec<ReceivingAddress>
```

ideally all our neptune-core RPCs that return multiple values should support paging, but one has to start somewhere.

dan-da: approach nack.

I do not see the point of showing the newest address in the Receive screen, as there is a good chance it may have been used already, which thus encourages address re-use.

The neptune-proton wallet deals with this issue in a different way that does not require any new RPC endpoint.

On the `Receive` screen it simply has a button to generate a new address (either Symmetric or Generation).   No RPC call is made at all until/unless the user clicks that button.

If user really wishes to see latest address, they can always visit the Addresses screen.

That's all.  simple.

Issue #724, opened 2025-10-03
Consider using multi-rpc to export single rpc impl via rest, json-rpc, tarpc


dan-da:
I just published a little crate [multi-rpc](https://crates.io/crates/multi-rpc) that provides macros which enable one to define a single trait and implementation for RPC methods, and it auto generates bindings for rest, jsonrpc, and tarpc.

Each rpc mechanism can be enabled (or not) via feature-flag.

So this is a basically zero maintenance way we can enable non-rust RPC clients to access our existing RPC APIs as well as any future APIs.

neptune-rpc would be the first app to use it for anything serious, so we might encounter some unforseen issues during integration, but hopefully nothing unsolveable.

Each RPC mechanism would be on its own port.

I could work on a PR after initial release of neptune-proton wallet.   I wanted to make this impl known now though in case of any duplicate work/efforts.

KaffinPX:
I actually completed a RPC framework, it's designed to support multiple transports but just JSON, it can serve around 200k calls/sec for a blockchain height call. Hoping to deprecate tarpc slowly and migrating to a general purpose ETH-like RPC...

dan-da:
Well, it's good for the project to have options to choose from I suppose.

It's unfortunate to duplicate work in this area, but then I don't recall ever being consulted as to your proposed design, so it is news to me.  Is the design documented or proposed somewhere?  In team meetings and on github I have consistently recommended a backwards compatible approach to cross-language rpc support.

I do not find deprecating tarpc to be a good goal necessarily.   It supports rust types cleanly and is easy to build rust clients with.  We presently have neptune-cli, neptune-dashboard, and now neptune-proton clients that make heavy use of it.  tarpc also supports streaming; we don't presently use that but it could be useful for efficiently notifying wallets of new mempool tx.

To me the obvious solution is to keep the present RPC endpoints essentially unchanged, but add bindings for additional cross-language protocols such as json-rpc and REST.   or just one of those.  Maintaining such bindings manually would be a bit annoying and brittle, hence I built these macros to automate.

The multi-rpc approach is backwards compatible, lightweight and only took a few days to implement, so no big loss if it is not chosen.

I estimate 1-2 days to integrate multi-rpc with neptune-core for an iniital draft PR, supporting all 3 rpc protocols.  The work would consist of:
1. minor refactoring of the RPC impls to remove the tarpc context params.
2. add multi-rpcs macros.
3. test.

KaffinPX:
[...]

PR #725, opened 2025-10-04
feat: support jsonrpc and rest rpc interfaces

### LINKED ISSUE #62: Discussion: tarpc (RPC layer) is rust specific.
related to #6.

I was surprised to find out that tarpc is a rust-specific RPC system, which of course would mean that all clients either need to be written in rust, or develop a tarpc impl for that language.   I think this would be shooting ourselves in the foot when it comes to getting exchange integration, onboard 3rd party wallet developers, etc.  Maybe this has already been considered/discussed?

I get that tarpc is slick, easy to implement, and provides some nice features like auto generation of an RPCClient from a trait, and full traceability between client/server.

My thought is that perhaps we can keep tarpc eg for the dashboard and future "official" wallet(s), but also add jsonrpc for compatibility with everything else.  Hopefully the jsonrpc would just be a wrapper around the existing tarpc API functions.   Or maybe those funcs will need an rpc abstraction layer.

jsonrpc of course has its warts.  It is pretty verbose, for one.  But it has also been kind of lingua-franca amongst cryptocurrencies and exchanges.  It is also very simple to write by hand if need be, and debug requests/responses manually, a nice property.   So I would think that we could have the jsonrpc for compat and easy use, and then tarpc is available for anyone that needs better performance.

Of course any cryptocurrency written in rust would face similar decisions so it may be worth surveying what others have done.  My information may be out-dated.  And I do notice that the rust jsonrpc crates haven't been updated in a couple of years which is not a good sign.

dan-da:
closes #62 #724 

adds a dep on crate [multi-rpc](https://crates.io/crates/multi-rpc).

This crate provides macros that wrap the RPC trait and generates wrappers for tarpc, jsonrpsee, and rest (via axum).

The basic idea is: one RPC implementation, multiple ways to access.

### motivation

1. There is a well known need for an RPC solution in neptune-core that is language agnostic.  One of my first issues was [discussing this](https://github.com/Neptune-Crypto/neptune-core/issues/62) need.  Now that we have launched mainnet, community interest and need is heating up.

2. neptune-proton wallet has architectural potential to become (fork) into a web wallet.  For this, neptune-core needs multi-wallet functionality.    The VXB team has implemented RPC APIs for multi-wallet.  We could integrate those APIs as tarpc APIs (which would work for neptune-proton), but to do it cleanly and also support the VXB wallet, we should expose them as REST apis.    I do not wish to fragment our RPCs into tarpc APIs vs REST APIs, hence I looked for a solution that enables us to maintain a single set of API methods, which can then be called by thin REST or tarpc wrappers.

3. json-rpc is used by bitcoin-core (and forks) and widely in crypto industry, so its nice to support it also.

4. It would be possible to manually implement wrappers for tarpc, rest, json-rpc, etc.  But that is tedious and brittle.  I wanted a solution that is automatic.  neptune-core devs need only maintain a single set of endpoint methods (and docs).

### status

This is a draft PR intended to demonstrate the concept/direction/simplicity.

tarpc and json-rpc are working.  REST server runs, but the routes are not wired up.  Just need to add some #[rest] attributes to methods.

I understand there is another RPC PR incoming from @KaffinPX, so it will be up to reviewers to decide which direction to take.

I'm happy to finish up this PR *if* reviewers like the direction and give a green light.   I estimate 1-3 days to add some polish and get it ready.  In particular:

1. add #[rest] attributes to endpoint impl methods, to define routes for REST.
4. should store task handle(s) to the rpc servers for clean(er) shutdown.  presently ctrl-c must be used twice.
5. make ports configurable for jsron-rpc and REST.
6. support custom protocol configs, eg large framesize in tarpc.
7. get the tests building and passing.
8. more testing of json-rpc and REST usage.

### example usage, tarpc client.

works just as before:

```
$ ./target/debug/neptune-cli block-info tip
height: 3
digest: 1082f85f827b3fdac5731b9c62241d866a28164930651f090c7c2c78a277aaef912d3713ba280000
size: 1166
prev_block_digest: 2e787b0691d47e7cf16ab3ed9cd2a6e8aafa1644c8f4aab23f335b9242514fda8018106645fe0300
timestamp: 2025-09-21T21:47:37.069-07:00
cumulative_proof_of_work: 18000
difficulty: 6000
num_inputs: 0
num_outputs: 2
num_announcements: 2
coinbase_amount: 128.00000000
fee: 64.00000000
is_genesis: false
is_tip: true
is_canonical: true
sibling_blocks:
```

### example usage, json-rpc client.

```
$ COOKIE=$(neptune-cookie-regtest.sh) && ~/dev/jsonrpc-cli/jsonrpc-cli http://localhost:9791/ block_info '{"token":{"Cookie": '"$COOKIE"'},"block_selector":"Tip"}'

{
    "Ok": {
        "coinbase_amount": 5.12e+32,
        "cumulative_proof_of_work": [
            18000,
            0,
            0,
            0,
            0,
            0
        ],
        "difficulty": [
            6000,
            0,
            0,
            0,
            0
        ],
        "digest": "1082f85f827b3fdac5731b9c62241d866a28164930651f090c7c2c78a277aaef912d3713ba280000",
        "fee": 2.56e+32,
        "height": 3,
        "is_canonical": true,
        "is_genesis": false,
        "is_tip": true,
        "num_announcements": 2,
        "num_inputs": 0,
        "num_outputs": 2,
        "prev_block_digest": "2e787b0691d47e7cf16ab3ed9cd2a6e8aafa1644c8f4aab23f335b9242514fda8018106645fe0300",
        "sibling_blocks": [],
        "size": 1166,
        "timestamp": 1758516457069
    }
}
```

neptune-cookie-regtest.sh:

a helper script for obtaining the rpc token in a json format (array).

```
xxd -p -c 32 ~/.local/share/neptune/regtest/.cookie | tr -d '\n' | sed -e 's/\(..\)/0x\1,/g' -e 's/,$/\n/' | python3 -c 'import sys; import json; print(json.dumps([int(x, 16) for x in sys.stdin.read().strip().split(",") if x]))'
```

raw http:

```
POST / HTTP/1.0
Accept: text/xml,application/xml,application/xhtml+xml,text/html,text/plain,image/png,image/jpeg,image/gif,*/*
Accept-Language: en-us
Content-Type: application/json
Host: localhost
User-Agent: Mozilla/4.0 (compatible; jsonrpc-cli HTTP Client; Linux)
Connection: Close
Content-Length: 218

{"jsonrpc":"2.0","method":"block_info","id":938110218,"params":{"token":{"Cookie":[173,75,65,96,191,123,205,98,123,36,155,50,224,207,7,2,119,20,101,39,172,214,2,18,2,117,180,167,76,232,62,106]},"block_selector":"Tip"}}

HTTP/1.0 200 OK
content-type: application/json; charset=utf-8
content-length: 585
date: Sat, 04 Oct 2025 02:44:42 GMT

{"jsonrpc":"2.0","id":938110218,"result":{"Ok":{"coinbase_amount":512000000000000000000000000000000,"cumulative_proof_of_work":[18000,0,0,0,0,0],"difficulty":[6000,0,0,0,0],"digest":"1082f85f827b3fdac5731b9c62241d866a28164930651f090c7c2c78a277aaef912d3713ba280000","fee":256000000000000005090463958695936,"height":3,"is_canonical":true,"is_genesis":false,"is_tip":true,"num_announcements":2,"num_inputs":0,"num_outputs":2,"prev_block_digest":"2e787b0691d47e7cf16ab3ed9cd2a6e8aafa1644c8f4aab23f335b9242514fda8018106645fe0300","sibling_blocks":[],"size":1166,"timestamp":1758516457069}}}
```

### example usage, REST client.

todo.


dan-da:
regarding neptune-dashboard and neptune-cli not building in this draft PR:

that is because the multi-rpc crate puts the generated code into separate namespaces, which results in the ClientRPC generated by tarpc not being found in original location.   I think I will remove that behavior or make it optional somehow.  Anyway, with a minor tweak to multi-rpc those should compile without any modification required.

PR #726, opened 2025-10-04
feat: HTTP-JSON RPC framework (server-side)

KaffinPX:


neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -0,0 +1,70 @@
...
+        Ok(())
+    }
+
+    #[apply(shared_tokio_runtime)]
+    async fn test_height_is_correct() -> Result<()> {
+        let rpc_server = test_rpc_server().await;
+        assert_eq!(0, rpc_server.height().await.height);
+        Ok(())
+    }
+}
```
Sword-Smith: This is great.

neptune-core/src/application/json_rpc/README.md
```diff
@@ -0,0 +1,43 @@
...
+     ```
+   * The first function should just call the second one.
+
+---
+
+## Server
+
+1. **Implement the `_call` method**
+
+   * In `json_rpc::server::service::RpcApi`, implement the function with the `_call` suffix.
```
Sword-Smith, line 39: I'm a bit confused by this. Are you adding two functions to the trait? Wouldn't you have to implement both functions then? Or is there some macro magic that treats function names ending with "_call" in a special way?

neptune-core/src/application/json_rpc/README.md
```diff
@@ -0,0 +1,43 @@
...
+## Core
+
+1. **Add a new operation**
+
+   * In `json_rpc::core::api::ops::RpcApiOps`, create a new variant (e.g. `SubmitBlock`).
+
+2. **Create request and response types**
+
+   * In `json_rpc::core::model::message`, define a matching pair of structs:
+     `SubmitBlockRequest` and `SubmitBlockResponse`.
```
Sword-Smith, line 14: Good idea with explicit request/reponse types and not relying on tuples and whatnot.

neptune-core/src/application/json_rpc/README.md
Sword-Smith: Or are you defining the function implementation directly in the trait for the function without the "_call" suffix?

KaffinPX: The first function is for the future Rust client, it just calls the _call function on implementations(Directly on trait) (We could remove it but then Rust clients also would have to use SubmitBlockRequest directly).

neptune-core/src/application/json_rpc/server/service.rs
Sword-Smith: Small nit here. These tests can never return error, so you can simplify the function signature by not having a return value (`()` return value) and dropping the last `Ok(())` line.

KaffinPX: Just a placeholder for now, need to extend network_is_consistent test(And standardify naming like that) to test over multiple networks but thanks for reminding, will fix soon

Sword-Smith:
Intial thoughts. Currently reviewing.

Formatting nits: 
- Get rid of trailing whitespaces
- avoid grouped imports (avoid this: `use serde::{Deserialize, Serialize};`, use two lines instead)
- don't extend comments beyond 80 columns, including docstrings.

Should we change the `Chain` namespace to `Light`? Because in the `Globalstate` definition, `chain` covers both `LightState` and `ArchivalState` whereas your use of "chain" maps to `LightState` AFAIK.

neptune-core/src/application/json_rpc/core/model/block/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+        common::RpcBFieldElements,
+    },
+    protocol::consensus::block::{block_kernel::BlockKernel, Block, BlockProof},
+};
+
+pub mod body;
+pub mod header;
+pub mod transaction_kernel;
+
+pub type RpcBlockProof = Option<RpcBFieldElements>;
```
Sword-Smith, line 15: Are `Option<T>` a good idea for JSON encoding? Are they uniform across implementations?

neptune-core/src/application/json_rpc/core/model/block/transaction_kernel.rs
```diff
@@ -0,0 +1,95 @@
...
+    api::export::Timestamp,
+    application::json_rpc::core::model::common::{RpcBFieldElements, RpcNativeCurrencyAmount},
+    protocol::consensus::transaction::transaction_kernel::TransactionKernel,
+    util_types::mutator_set::removal_record::{
+        absolute_index_set::AbsoluteIndexSet, chunk_dictionary::ChunkDictionary, RemovalRecord,
+    },
+};
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Default)]
+pub struct RpcChunkDictionary(pub BTreeMap<u64, (Vec<Digest>, Vec<u32>)>);
```
Sword-Smith, line 18: Seems like a pretty wild type to JSON encode, no? 

What about this instead?
```rust
pub struct RpcChunkDictionary {
    authenticated_chunks: Vec<RpcIndexedChunk>,
}

pub struct RpcIndexedChunk {
    index: u64,
    authenticated_chunk: RpcAuthenticatedChunk,
}

pub struct RpcAuthenticatedChunk {
    relative_indices: Vec<u32>,
    authentication_path: Vec<Digest>,
}
```

neptune-core/src/application/json_rpc/core/model/block/transaction_kernel.rs
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RpcTransactionKernel {
+    pub inputs: Vec<RpcRemovalRecord>,
+    pub outputs: Vec<Digest>,
+    pub announcements: Vec<RpcBFieldElements>,
```
Sword-Smith, line 55: What's wrong with `Vec<Vec<BFieldElement>>`?

neptune-core/src/application/json_rpc/core/model/block/transaction_kernel.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub fee: RpcNativeCurrencyAmount,
+    pub coinbase: Option<RpcNativeCurrencyAmount>,
```
Sword-Smith, line 57: Are `Option<T>` types OK for JSON serialization?

neptune-core/src/application/json_rpc/core/model/common.rs
```diff
@@ -0,0 +1,99 @@
...
+impl<'de> Deserialize<'de> for RpcNativeCurrencyAmount {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let s = String::deserialize(deserializer)?;
+        let nau = i128::from_str(&s).map_err(serde::de::Error::custom)?;
+        Ok(RpcNativeCurrencyAmount(NativeCurrencyAmount::from_nau(nau)))
+    }
+}
```
Sword-Smith, line 40: You're again using a tuple struct here. Is that really OK? If it is, please explain why it's OK.

neptune-core/src/application/json_rpc/server/http.rs
```diff
@@ -0,0 +1,125 @@
...
+        let router = RpcMethods::new_router(api, namespaces);
+
+        let app = Router::new()
+            .route("/", post(Self::rpc_handler))
+            .with_state(Arc::new(router));
+
+        axum::serve(listener, app).await.unwrap();
+    }
+
+    async fn rpc_handler(
```
Sword-Smith, line 109: So all endpoints are located at the servers root address `/`. And the concrete endpoint is selected with a "method" field in request body, which needs to be JSON encoded. So to reach the `foo` endpoint in the `bar` namespace, send a POST request with this body to `/`:
```json
{
    "method": "bar_foo",
    "params": [],
}
```

Maybe add some docstrings here showing how these endpoints are addressed?

neptune-core/src/application/json_rpc/server/http.rs
```diff
@@ -0,0 +1,125 @@
...
+            server::{http::RpcServer, service::tests::test_rpc_server},
+        },
+        tests::shared_tokio_runtime,
+    };
+    use axum::{extract::State, Json};
+    use macro_rules_attr::apply;
+    use serde_json::json;
+
+    #[apply(shared_tokio_runtime)]
+    async fn namespace_isolates_correctly() {
```
Sword-Smith, line 150: So this is a test that verifies that deactivated namespaces are actually deactivated? I like the idea. But we can make this test more robust, I think:

```rust
    #[apply(shared_tokio_runtime)]
    async fn namespace_isolates_correctly() {
        let router_no_chain = Arc::new(RpcMethods::new_router(
            Arc::new(test_rpc_server().await),
            HashSet::from([Namespace::Node]),
        ));
        let router_with_chain = Arc::new(RpcMethods::new_router(
            Arc::new(test_rpc_server().await),
            HashSet::from([Namespace::Node, Namespace::Chain]),
        ));

        async fn make_rpc_request(router: Arc<RpcRouter>, method: &str) -> RpcResponse {
            let request = RpcRequest {
                jsonrpc: Some("2.0".to_string()),
                method: method.to_string(),
                params: json!([]),
                id: None,
            };
            RpcServer::rpc_handler(State(router), Ok(Json(request)))
                .await
                .0
        }

        let node_network_res = make_rpc_request(router_no_chain.clone(), "node_network").await;
        assert!(
            matches!(node_network_res, RpcResponse::Success { .. }),
            "Expected success for node_network, got: {:?}",
            node_network_res
        );

        let chain_height_method_name = "chain_height";
        let chain_height_res_bad =
            make_rpc_request(router_no_chain, chain_height_method_name).await;
        assert!(
            matches!(
                chain_height_res_bad,
                RpcResponse::Error {
                    error: RpcError::MethodNotFound,
                    ..
                }
            ),
            "Expected MethodNotFound error for chain_height, got: {:?}",
            chain_height_res_bad
        );
        let chain_height_res_good =
            make_rpc_request(router_with_chain.clone(), chain_height_method_name).await;
        assert!(
            matches!(chain_height_res_good, RpcResponse::Success { .. }),
            "Expected success for chain_height, got: {:?}",
            chain_height_res_good
        );
    }
```

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    #[apply(shared_tokio_runtime)]
+    async fn height_is_correct() {
+        let rpc_server = test_rpc_server().await;
+        assert_eq!(0, rpc_server.height().await.height);
+    }
+
+    #[test_strategy::proptest(async = "tokio", cases = 5)]
+    async fn block_calls_are_consistent(
```
Sword-Smith: Excellent test 🥇

neptune-core/src/protocol/consensus/block/block_body.rs
```diff
@@ -80,15 +80,15 @@ pub struct BlockBody {
     /// This field represents the state of the MS *after* applying the update
     /// induced by the transaction, but *before* applying the update induced by
     /// guesser fees (and perhaps later composer fees).
-    pub(super) mutator_set_accumulator: MutatorSetAccumulator,
+    pub mutator_set_accumulator: MutatorSetAccumulator,
```
Sword-Smith, line 85: This `pub` scares me. Because it was made `pub(super)` to prevent this field from being read, since what you're usually after is the output of the method `mutator_set_accumulator_after`. So you're opening a pitfall for later developers here. Is there a way to avoid this?

neptune-core/src/application/json_rpc/core/model/common.rs
KaffinPX: On Tuples serde (de)serialize the inside directly, it's mainly used for avoiding nesting so it won't be like { amount: String } but just String.

neptune-core/src/application/json_rpc/core/model/block/transaction_kernel.rs
KaffinPX: Yeah, it will be de/serialized as `null`

neptune-core/src/application/json_rpc/core/model/block/transaction_kernel.rs
KaffinPX: This way it gets de/serialized as hex, more compact than Vec<BFieldElement> and better DX.

neptune-core/src/application/json_rpc/core/model/block/transaction_kernel.rs
KaffinPX: As we were processing anyway, thought it would make sense to index it; it gets de/serialized as an object { id: [[], []]}.

neptune-core/src/application/json_rpc/core/model/block/mod.rs
KaffinPX: Yeah its `null | T`.

KaffinPX:
> There's also this: `pub struct RpcActiveWindow(pub Vec<u32>);`. Don't you need named fields for uniform JSON encoding? You can just use the regular `ActiveWindow` here:
> 
> ```rust
> pub struct ActiveWindow {
>     // It's OK to store this in memory, since it's on the size of kilobytes, not gigabytes.
>     pub sbf: Vec<u32>,
> }
> ```

This way we can avoid nesting, as it was a single value struct.

neptune-core/src/protocol/consensus/block/block_body.rs
KaffinPX: Maybe we should return `mutator_set_accumulator_after` accumulator on RPC? (So we cant hmm) -- we decided to add a comment.

neptune-core/src/application/json_rpc/core/model/block/transaction_kernel.rs
Sword-Smith: Did you mean `index: [[], []]`, where `index` is the chunk index? That can work, I guess.

neptune-core/src/application/json_rpc/core/model/message.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+#[derive(Clone, Copy, Debug, Serialize, Deserialize_tuple)]
+#[serde(rename_all = "camelCase")]
+pub struct HeightRequest {}
+
+#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct HeightResponse {
+    pub height: u64, // This technically could exceed JavaScript's safe int limits but practically it would take thousand(?) years.
```
Sword-Smith: Comment exceeds max length of 80 characters.

neptune-core/src/application/json_rpc/server/http.rs
```diff
@@ -0,0 +1,187 @@
...
+    /// - `jsonrpc`: `"2.0"`
+    /// - `method`: The RPC method to call, formatted as `namespace_method`
+    /// - `params`: An array of parameters to pass to the method
+    /// - `id` (optional): Request identifier for matching responses
+    ///
+    /// # Response
+    ///
+    /// Returns a JSON-RPC 2.0 response:
+    /// - On success: `{ "jsonrpc": "2.0", "id": <request_id>, "result": <method_result> }`
+    /// - On error: `{ "jsonrpc": "2.0", "id": <request_id>, "error": { "code": <error_code>, "message": <error_message> } }`
```
Sword-Smith: Max line width exceeded. Max is 80 characters for comments.

PR #727, opened 2025-10-06
fork!: Change PoW algorithm to make proposal-switching free

Sword-Smith:
Change the PoW algorithm in order to increase transaction throughput.

A small algorithmical change to the PoW algorithm which constitutes a hard-fork since old versions of the software will consider new blocks invalid.

The current PoW algorithm has a preprocessing step that takes ~12 seconds on a very fast GPU and ~40 seconds on a fast CPU. This preprocessing step is invalidated by a new block proposal, meaning that the guessers are heavily disincentivized from switching block proposals while mining. The preprocessing step involves building a large Merkle tree, and it is the leafs of this Merkle tree that depends on the concrete block proposal. With this commit, the leafs instead depend only on the previous' block's hash which means that switching between block proposals is free as long as the parent of the two proposals is the same.

The reason that this change is good for transaction throughput is that it makes it much more likely that the competitive miner will switch to whatever proposal that pays them the highest fee. And including merging another fee-paying transaction into the block proposal increases the reward for the guesser. In the scenario where a mining pool decides both the block proof and the guessser reward (and pays out users for their near-misses) this change *also* benefits transaction throughput since the preprocessing step (being up to 42GB) will be done locally at each user, and if they can reuse the preprocessed data for a new block proposal/task sent by the mining pool, the mining pool is more likely to update the task regularly as new transactions are merged into the coinbase.


Sword-Smith:
Problem with current approach @aszepieniec. I think this line allows guessers to guess on multiple proposals at once. Which we probably don't want:

```rust
let buffer_hash = Tip5::hash_pair(self.root, leaf_prefix);
```

Whole validation algorithm:
```rust
    pub(super) fn validate(
        self,
        auth_paths: PowMastPaths,
        target: Digest,
        consensus_rule_set: ConsensusRuleSet,
        parent_digest: Digest,
    ) -> Result<(), PowValidationError> {
        let leaf_prefix = match consensus_rule_set {
            ConsensusRuleSet::Reboot => auth_paths.commit(),
            ConsensusRuleSet::HardforkAlpha => parent_digest,
        };
        let buffer_hash = Tip5::hash_pair(self.root, leaf_prefix);
        let (index_a, index_b) = Self::indices(buffer_hash, self.nonce);

        let (leaf_a, leaf_b) = if consensus_rule_set == ConsensusRuleSet::Reboot {
            (
                Self::leaf(leaf_prefix, index_a),
                Self::leaf(leaf_prefix, index_b),
            )
        } else {
            let index_a = u64::from(Self::bitreverse(
                index_a.try_into().unwrap(),
                Self::MERKLE_TREE_HEIGHT as u32,
            ));
            let index_b = u64::from(Self::bitreverse(
                index_b.try_into().unwrap(),
                Self::MERKLE_TREE_HEIGHT as u32,
            ));
            (
                Self::leaf(leaf_prefix, index_a),
                Self::leaf(leaf_prefix, index_b),
            )
        };

        if !MTree::verify(self.root, index_a as usize, &self.path_a, leaf_a) {
            return Err(PowValidationError::PathAInvalid);
        }
        if !MTree::verify(self.root, index_b as usize, &self.path_b, leaf_b) {
            return Err(PowValidationError::PathBInvalid);
        }

        let pow_digest = auth_paths.fast_mast_hash(self);
        let meets_threshold = pow_digest <= target;
        if !meets_threshold {
            return Err(PowValidationError::ThresholdNotMet);
        }

        Ok(())
    }
```

I suggest we change the problematic line to:
```rust
let buffer_hash = Tip5::hash_pair(self.root, auth_paths.commit());
```

1. Is compatible with current `ConsensusRuleSet::Reboot`.
2. Prevents guessers from reusing guesses across multiple proposals

For transaction throughput it's important that guessers are incentivized to guess on the most valuable proposals (and not just *any*) as that's the best way to incentivize the mining of transactions.

Sword-Smith:
> Problem with current approach @aszepieniec. I think this line allows guessers to guess on multiple proposals at once ...

Motivated by this I added a new required test above: Ensure that opened indices into the Merkle tree change when the proposal changes, not just when the parent block changes.

neptune-core/src/application/loops/peer_loop.rs
```diff
@@ -2407,10 +2417,10 @@ mod tests {
             Ok(())
         }
 
-        /// Return three blocks:
+        /// Return four blocks:
```
jan-ferdinand, line 2420:
```suggestion
        /// Return three blocks:
```

out of date?

neptune-core/src/application/loops/peer_loop.rs
```diff
@@ -3747,6 +3761,146 @@ mod tests {
...
+
+        #[traced_test]
+        #[test]
+        fn hardfork_alpha_happy_case() {
+            // Ensure that valid blocks are all accepted across hardfork alpha.
+            // Scenario: Current height is hardfork minus 2. Then receives peer
+            // notification of first block after hardfork. Must accept all
+            // blocks from peer, not punish peer, and send blocks to main
+            // loop.
+            let init_block_heigth = BlockHeight::from(14999u64);
```
jan-ferdinand, line 3797: Depend on hardfork const instead of duplicating its value here.

neptune-core/src/application/loops/peer_loop.rs
```diff
@@ -3747,6 +3761,146 @@ mod tests {
...
+                    ConsensusRuleSet::Reboot
+                ));
+
+                // Declare expected order of messages
+                let mock = Mock::new(vec![
+                    Action::Read(PeerMessage::BlockNotification(
+                        (&first_block_after_hardfork).into(),
+                    )),
+                    Action::Write(PeerMessage::BlockRequestByHeight(BlockHeight::from(
+                        15000u64,
```
jan-ferdinand, line 3876: and here

neptune-core/src/lib.rs
```diff
@@ -5,7 +5,7 @@
 // enables nightly feature async_fn_track_caller for crate feature log-slow-write-lock.
 // log-slow-write-lock logs warning when a write-lock is held longer than 100 millis.
 // to enable: cargo +nightly build --features log-slow-write-lock
-#![cfg_attr(feature = "track-lock-location", feature(async_fn_track_caller))]
+// #![cfg_attr(feature = "track-lock-location", feature(async_fn_track_caller))]
```
jan-ferdinand: Is this a left-over from debugging? It seems like a useful cfg to me.

neptune-core/src/protocol/consensus/block/pow.rs
```diff
@@ -274,18 +276,34 @@ impl<const MERKLE_TREE_HEIGHT: usize> Display for Pow<MERKLE_TREE_HEIGHT> {
...
+    prev_block_digest: Digest,
+}
+
+impl<const MERKLE_TREE_HEIGHT: usize> GuesserBuffer<MERKLE_TREE_HEIGHT> {
+    /// A commitment that refers to both the Merkle tree of the guesser buffer
+    /// and the current proposal being guessed on, designed in such a way that
+    /// the Merkle tree root must be known before indices can be picked.
+    /// Combined with a nonce, the opened indices can be derived from this
+    /// value.
+    pub fn index_pricker_preimage_from_root(
```
jan-ferdinand:
```suggestion
    pub fn index_picker_preimage_from_root(
```

neptune-core/src/protocol/consensus/block/pow.rs
```diff
@@ -274,18 +276,34 @@ impl<const MERKLE_TREE_HEIGHT: usize> Display for Pow<MERKLE_TREE_HEIGHT> {
...
+        merkle_tree_root: Digest,
+        mast_auth_paths: &PowMastPaths,
+    ) -> Digest {
+        Tip5::hash_pair(merkle_tree_root, mast_auth_paths.commit())
+    }
 
-    /// Authentication paths for all fields but the PoW field
-    mast_auth_paths: PowMastPaths,
+    /// Like [Self::index_pricker_preimage_from_root]
```
jan-ferdinand: Feels like either this or the alike function is superfluous.

neptune-core/src/protocol/consensus/block/pow.rs
```diff
@@ -583,28 +688,170 @@ pub(crate) mod tests {
...
+        assert_eq!(14, Pow::<10>::bitreverse(7, 4));
+        assert_eq!(3, Pow::<10>::bitreverse(7, 2));
+        assert_eq!(1, Pow::<10>::bitreverse(7, 1));
+        assert_eq!(7, Pow::<10>::bitreverse(14, 4));
+        assert_eq!(3, Pow::<10>::bitreverse(12, 4));
+        assert_eq!(19, Pow::<10>::bitreverse(100, 7));
+    }
+
+    #[proptest]
+    fn bitreverse_is_symmetric(#[strategy(arb())] k: u32, #[strategy(1u32..=32)] log2_n: u32) {
```
jan-ferdinand:
```suggestion
    fn bitreverse_is_symmetric(k: u32, #[strategy(1u32..=32)] log2_n: u32) {
```

There is an implementation of `proptest::Arbitrary` for all primitive types, so the `arbitrary::Arbitrary` to `proptest::Arbitrary` bridge that `proptest_arbitrary_interop` provides is not necessary.

neptune-core/src/protocol/consensus/block/pow.rs
```diff
@@ -583,28 +688,170 @@ pub(crate) mod tests {
...
+        let mask: u32 = if log2_n == 32 {
+            u32::MAX
+        } else {
+            (1u32 << log2_n) - 1
+        };
```
jan-ferdinand:
```suggestion
        let mask = u32::MAX >> (32 - log2_n);
```

neptune-core/src/protocol/consensus/block/pow.rs
```diff
@@ -583,28 +688,170 @@ pub(crate) mod tests {
...
+            buffer.index_picker_preimage(&auth_paths1),
+            Digest::default(),
+        );
+        let indices_2 = Pow::<10>::indices(
+            buffer.index_picker_preimage(&auth_paths2),
+            Digest::default(),
+        );
+
+        assert_ne!(indices_1, indices_2);
+    }
```
jan-ferdinand:
```suggestion
    /// Ensure that indices cannot be reused over two proposals that share the
    /// same parent.
    #[proptest(cases = 10)]
    fn indices_depend_on_concrete_proposal_hardfork_alpha(
        #[strategy(arb())] prev_block_digest: Digest,
        #[strategy(arb())] auth_paths_1: PowMastPaths,
        #[strategy(arb())] auth_paths_2: PowMastPaths,
        #[strategy(arb())] nonce: Digest,
    ) {
        const MERKLE_TREE_HEIGHT: usize = 10;

        let buffer = Pow::<MERKLE_TREE_HEIGHT>::preprocess(
            auth_paths_1,
            None,
            ConsensusRuleSet::HardforkAlpha,
            prev_block_digest,
        );

        let indices_1 =
            Pow::<MERKLE_TREE_HEIGHT>::indices(buffer.index_picker_preimage(&auth_paths_1), nonce);
        let indices_2 =
            Pow::<MERKLE_TREE_HEIGHT>::indices(buffer.index_picker_preimage(&auth_paths_2), nonce);

        assert_ne!(indices_1, indices_2);
    }
```

Requires `#[cfg_attr(any(test, feature = "arbitrary-impls"), derive(arbitrary::Arbitrary))]` annotation for `PowMastPaths`.

neptune-core/src/protocol/consensus/block/pow.rs
```diff
@@ -583,28 +688,170 @@ pub(crate) mod tests {
...
+
+    #[test]
```
jan-ferdinand: This also feels like a proptest.

neptune-core/src/application/loops/peer_loop.rs
Sword-Smith: Good point! I can actually return a 4th block here to harden two tests.

neptune-core/src/application/loops/peer_loop.rs
Sword-Smith: Disagree. I prefer not to have indirections in this test. And I want it to fail if someone changes the block height for when the hard fork happens.

neptune-core/src/lib.rs
Sword-Smith: agreed

neptune-core/src/protocol/consensus/block/pow.rs
Sword-Smith: OK :)

neptune-core/src/protocol/consensus/block/pow.rs
Sword-Smith: Why not? :)

neptune-core/src/protocol/consensus/block/pow.rs
Sword-Smith: Got it.

neptune-core/src/protocol/consensus/block/pow.rs
Sword-Smith: I like it as it is because both the verifier and the prover needs to call this. And the verifier does not have the buffer, only the Merkle root, whereas the prover has the buffer, so they can use the method instead of the associated function.

neptune-core/src/protocol/consensus/block/pow.rs
```diff
@@ -274,18 +277,34 @@ impl<const MERKLE_TREE_HEIGHT: usize> Display for Pow<MERKLE_TREE_HEIGHT> {
...
+    prev_block_digest: Digest,
+}
+
+impl<const MERKLE_TREE_HEIGHT: usize> GuesserBuffer<MERKLE_TREE_HEIGHT> {
+    /// A commitment that refers to both the Merkle tree of the guesser buffer
+    /// and the current proposal being guessed on, designed in such a way that
+    /// the Merkle tree root must be known before indices can be picked.
+    /// Combined with a nonce, the opened indices can be derived from this
+    /// value.
+    pub fn index_picker_preimage_from_root(
```
aszepieniec: Excellent formalism, but as far as I can see there is no need to make this function public. Private will do.

neptune-core/src/application/loops/mine_loop.rs
```diff
@@ -231,22 +240,37 @@ fn guess_worker(
...
         .unwrap();
+
+    let index_picker_preimage = guesser_buffer.index_picker_preimage(&mast_auth_paths);
     let guess_result = pool.install(|| {
         rayon::iter::repeat(0)
-            .map_init(rand::rng, |rng, _i| {
-                guess_nonce_iteration(&guesser_buffer, threshold, rng, &sender)
-            })
+            .map_init(
+                || rng.clone().unwrap_or(std_rng_from_thread_rng()),
```
aszepieniec, line 262: Can you convince me that this generates a different RNG per rayon thread?

aszepieniec (commented):
What happened to file `preprocess_task.rs`? It contained some useful interfacing between async and non-async code that will be difficult to reproduce. Unfortunately, it seems to have been rebased away.

lib.rs:8: `#![cfg_attr(feature = "track-lock-location", feature(async_fn_track_caller))]` Does not compile on stable rust.

Under HFa rules, preprocessing is independent of the block proposal. The only requirement is the prev_block_digest. As a result:
 - preprocessing can start before the block proposal is finished (not necessarily good);
 - you can switch to another block proposal for free (necessarily good: high tx confirmation rate).

The indices depend on a) the PoW MAST paths, and on b) the nonce.

We want to ensure that every hash invocation occurring inside the tight loop fixes the nonce, the proposal, and the guesser buffer. If we ensure these qualities then there is no way for the clever optimizer to reuse (or save) hashes across different iterations.

The index picker preimage depends on the guesser buffer Merkle root and the proposal's MAST auth paths. The indices depend on that and the nonce.

Therefore:
 - If the predecessor block is different, then the guesser buffer is different, and consequently the root is different. To obtain paths valid relative to this root, the miner must store two buffers or recompute on the fly the missing parts of the paths.
 - If the predecessor block is the same but the proposal is different (up to PoW data), then the index picker preimage is different, which results in different path indices being sampled with reasonably high probability (more on this later).
 
If the miner uses the same nonce to guess on two different proposals which have the same predecessor, and there happens to be a collision in the path indices, then the bulk of the hashes in the tight loop can be reused. Specifically, the encoding of `Pow` is identical and so is its hash. I count 7 permutations instead of 39. In the case of two proposals, the probability of this event is small enough to not care: 2^-58. However, a malicious strategy involves sitting on a large number k of proposals and computing the path indices for all k proposals takes 8 hashes (?) per index pair. Bear with me for some probability calculations.

Of the 2^58 possible index pairs, each one has probability (1-2^-58)^k of not being sampled at all, and one minus that gives the probability of any given index pair being sampled once or more. Multiply that with the total number of index pairs to get the expected number of unique pairs after sampling k times with repetition: 2^58 * (1 - (1 - 2^-58)^k). So the expected number of collisions is k - 2^58 * (1 - (1 - 2^-58)^k).

k = 2^ 1  -> E[#coll] =  3.46944695195361418882384896278381347656250000000000000000000000000000000000000000000000000e-18
k = 2^ 10  -> E[#coll] =  1.81721304670645407755600038204345509918940712495986238547234553052233997780675469727477754e-12
k = 2^ 20  -> E[#coll] =  1.90734681382078349612633122391242953432722558829097633772290943850314942173699061499155706e-6
k = 2^ 30  -> E[#coll] =  1.99999999565382799437962544667008217171070416879076367096722052605292644172618887807248906
k = 2^ 40  -> E[#coll] =  2.09714933333396912154662710620707230176903836932520861762517803150974286467752614993385633e6

So after collecting k = 2^30 proposals we may reasonably expect one colliding pair of indices. If we enter into the tight guessing loop with these two proposals then we can spend 39 hashes to trial one (proposal, nonce) pair, and 7 more hashes to trial the second.

Sword-Smith:
Feedback received on another channel:

> _anonymous:_ 
the reason this change will make 0 difference to transaction throughput is because all dominant pools use private composers
_sword_smith_: quoting PR: "In the scenario where a mining pool decides both the block proof and the guessser reward (and pays out users for their near-misses) this change also benefits transaction throughput since the preprocessing step (being up to 42GB) will be done locally at each user, and if they can reuse the preprocessed data for a new block proposal/task sent by the mining pool, the mining pool is more likely to update the task regularly as new transactions are merged into the coinbase."
_anonymous_: the only incentive being, more tx per block is better for the project
_sword_smith_: More tx per block means higher guesser fee.
_anonymous_: oh. right. makes sense now, thanks

aszepieniec (approved):

PR #729, opened 2025-10-07
`libp2p` for running in parallel with the current peer loop

skaunov:
continuation of #701 from a more recent commit

it has a lot of comments to delete when pass the reviews -- I left those to better convey the implementation decisions 

Also given the size of this change it makes sense to split/separate some things which you're OK with to merge earlier. This should ease all the aspects of handling it; I just had some hard time guessing which parts could get a go --- let me listen from you first! :tophat: 

# ~~todos~~
never mind, it was reverted
> I was surprised how many todos the code held. I thought just to invent some other tag for my current todos, but I truly think the system is beyond the point when this was acceptable. Some of them 3 y.o.! My point is they should be converted to the tracker issues and replaced by the reference to the issue. I'll bring them back if I'm wrong on this. But I hope after your first natural rejection reaction you would take less than 2 hours to map the issues to your todos. My first contribution was a small issue from the tracker, not a todo from the code, as an example.
> 
> And ofc I'm converting all the todos I've added around into the issues when the review is passed! :pencil2:

### DIFF
--- a/neptune-core-cli/src/main.rs
+++ b/neptune-core-cli/src/main.rs
@@ -1183,7 +1183,7 @@ async fn main() -> Result<()> {
                         &data_directory,
                         network,
                         tx_artifacts.all_offchain_notifications(),
-                        None, //  parse receiver tags from cmd-line.
+                        None, // todo:  parse receiver tags from cmd-line.
                     )?
                 }
                 Err(e) => eprintln!("{e}"),
@@ -1415,7 +1415,8 @@ fn process_utxo_notifications(
         println!("\n*** Utxo Transfer Files ***\n");
     }
 
-    // It would be better if this timestamp was read from the created transaction.
+    // TODO: It would be better if this timestamp was read from the created
+    // transaction.
     let timestamp = SystemTime::now()
         .duration_since(UNIX_EPOCH)
         .unwrap()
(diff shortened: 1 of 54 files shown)


skaunov:
I was asking for an advice on this few weeks ago in the dev Tg group. I mean I'd add Rust tests but I can't think out of how to do it with this thing. Any attempt/approach to it ends for me as building a separate container network which would be doing such testing. It looks like a right thing to do, but I had not enough motivation for it yet, and also wanted to have more eyes on this first, and discuss better ideas.

Sword-Smith:
By end-to-end testing I meant how I could start neptune-core with this new peer loop.

skaunov:
> ...has no formalized style guide (other than what cargo fmt and clippy enforce), I'll let it be up to the individual developer when they write _their_ code (within reason). But please don't fix comments unrelated to this PR after your personal preferences.

I would say sticking to some tooling is really better, thanks no such "formalized style guide" without a tool is there. :thumb_up: Let me avoid of committing those, as not fixing is hard when it's part of the reading. I would also speculate the code belongs more to those who is actively working on it today; I mean why the idea of iterative collaboration/development would stop when a line is commented. :upside_down_face: And for a fault change the review is here --- sincerely sorry again for the additional strain; I'll be trying to keep the comment changes local.

skaunov:
@Sword-Smith 
> By end-to-end testing I meant how I could start neptune-core with this new peer loop.

saving here for a note: [`str1ke`](@sploitzberg) [offered to help with running the thing on a test-net](https://t.me/neptune_dev/1623)

PR #736, opened 2025-10-14
fix(Mempool): Cover edge case for initiating Update

aszepieniec:
This commit fixes an edge case in the mempool logic that results in transactions initiated by ourselves to be ignored for Updates as new blocks are found, even when we were the ones to produce a SingleProof for them.

The edge case comes about as follows:
 - We initiate the transaction and produce a SingleProof for it.
 - We broadcast the transaction.
 - Someone else merges the transaction with some other transaction and broadcasts that.
 - That merge-result is inserted into our mempool and this insertion moves the original transaction into the cache (without primitive witness).
 - The other operand of the merger is mined. So our transaction is fetched from the cache and inserted back into the mempool (without primitive witness).
 - A new block is mined.
 - We do not Update our transaction because
 - it does not have an associated primitive witness <-- error is here.

The correct criterion for deciding whether a transaction should be updated is whether it is critical; not whether it has a primitive witness. Critical means "initiated by us". Transactions cannot have a primitive witness without being critical; but they obviously can be critical without having a primitive witness. More to the point, the primitive witness is not relevant data for performing an Update job.

### DIFF
--- a/neptune-core/src/state/mempool.rs
+++ b/neptune-core/src/state/mempool.rs
@@ -1038,12 +1038,19 @@ impl Mempool {
                 // mempool in the expectation that someone will update them to
                 // be valid under a new mutator set.
                 //
-                // If the transaction was initiated locally, *and* node can
-                // produce single-proofs, transaction should be updated
-                // immediately (and be kept in mempool).
+                // If (the transaction was initiated locally or deemed
+                // critical), *and* node can produce single-proofs, transaction
+                // should be updated immediately (and be kept in mempool).
+                //
+                // Note: Do not check for the presence of a primitive witness.
+                // This information is irrelevant for Update tasks and moreover
+                // not always present for critical transactions. For instance:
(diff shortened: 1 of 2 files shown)


neptune-core/src/state/mempool.rs
```diff
@@ -1038,12 +1038,19 @@ impl Mempool {
                 // mempool in the expectation that someone will update them to
                 // be valid under a new mutator set.
                 //
-                // If the transaction was initiated locally, *and* node can
-                // produce single-proofs, transaction should be updated
-                // immediately (and be kept in mempool).
+                // If (the transaction was initiated locally or deemed
+                // critical), *and* node can produce single-proofs, transaction
```
Sword-Smith, line 1042: deemed critical should be sufficient here, no?

Sword-Smith:
Merged via CLI. In 91b9a3d656df55585f31df4f185cd1988c2a2a82.

PR #738, opened 2025-10-15
Dos protection against incoming connections

Sword-Smith:
Some DOS mitigation implementations: 
- shorten timeout on handshake
- Don't allocate errors on failure to send handshakes
- Limit number of spawned tokio tasks for incoming connections
- Check database for ban before accepting or sending handshake on incoming connection.

### DIFF
--- a/neptune-core/src/application/loops/main_loop.rs
+++ b/neptune-core/src/application/loops/main_loop.rs
@@ -1708,7 +1708,15 @@ impl MainLoopHandler {
 
                 // Handle incoming connections from peer
                 Ok((stream, peer_address)) = self.incoming_peer_listener.accept() => {
-                    if !precheck_incoming_connection_is_allowed(self.global_state_lock.cli(), peer_address.ip()) {
+                    let ip = peer_address.ip();
+                    if !precheck_incoming_connection_is_allowed(self.global_state_lock.cli(), ip) {
+                        continue;
+                    }
+
+                    // Is this IP banned through database entry?
+                    let peer_banned = self.global_state_lock.lock_guard().await.net.peer_databases.peer_standings.get(ip).await.is_some_and(|x| x.is_bad());
+                    if peer_banned {
+                        debug!("Banned peer {ip} attempted incoming connection. Hanging up.");
                         continue;
                     }
 


neptune-core/src/application/loops/connect_to_peers.rs
```diff
@@ -339,17 +341,27 @@ where
     >::new(length_delimited, SymmetricalBincode::default());
 
     // Complete Neptune handshake
-    let Some(PeerMessage::Handshake {
-        magic_value,
-        data: peer_handshake_data,
-    }) = peer.try_next().await?
-    else {
-        bail!("Didn't get handshake on connection attempt, from {peer_address}");
+    let maybe_msg = timeout(Duration::from_secs(4), peer.try_next()).await;
```
jan-ferdinand: Make the timeout configurable to allow simple adaptation under bad network conditions?

neptune-core/src/application/loops/connect_to_peers.rs
```diff
@@ -339,17 +341,27 @@ where
...
+    let (magic_value, peer_handshake) = match maybe_msg {
+        Ok(Ok(Some(PeerMessage::Handshake { magic_value, data }))) => (magic_value, data),
+        Ok(Ok(_)) => {
+            tracing::warn!(%peer_address, "unexpected message instead of handshake");
+            return Ok(()); // no heavy anyhow::Error, just close
```
jan-ferdinand: I can see why you do this, but it feels odd. How about a doc comment regarding the function's behavior:

“A returned `Result::Error` always indicates that the connection was not initiated. A returned `Result::Ok` means that either (a) the connection was initiated, or (b) the handshake was not completed. The reason for this behavior …”.

You might also consider a funky return type like `core::result::Result<(), Result<()>>`:
- `Ok(())` means connection was initiated
- `Err(Ok(())` is lightweight and indicates an early abort
- `Err(Err(…))` is heavyweight and indicates a different kind of error

It might be possible to introduce a new error enum that has a lightweight variant in the early-abort case and a heavyweight variant in the late-abort case.

neptune-core/src/application/loops/main_loop.rs
```diff
@@ -1696,16 +1708,31 @@ impl MainLoopHandler {
...
+                    // acquisitions.
+                    let permit = incoming_connections_limit.clone().acquire_owned().await?;
+
                     let state = self.global_state_lock.lock_guard().await;
                     let main_to_peer_broadcast_rx_clone: broadcast::Receiver<MainToPeerTask> = self.main_to_peer_broadcast_tx.subscribe();
                     let peer_task_to_main_tx_clone: mpsc::Sender<PeerTaskToMain> = self.peer_task_to_main_tx.clone();
                     let own_handshake_data: HandshakeData = state.get_own_handshakedata();
                     let global_state_lock = self.global_state_lock.clone(); // bump arc refcount.
                     let incoming_peer_task_handle = tokio::task::spawn(async move {
+                        let _permit = permit; // drop releases slot
```
jan-ferdinand: This is not dropping the permit. If you want to drop the permit here, I suggest to be explicit: `drop(permit);`
If you want to use arcane language features, use `let _ = permit;`. [See also](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#explanation-15).

jan-ferdinand: Is it important that dropping the permit happens in the async closure? It's not obvious to me, so it might warrant a comment. If it's obvious to more experienced async programmers, then it probably doesn't need a comment.

jan-ferdinand:
Doesn't the semaphore make DOSing even easier? Now, an attacker only has to keep a rather limited number of connection attempts open at the same time. Might well be that I don't understand the solution well enough.

neptune-core/src/application/loops/connect_to_peers.rs
Sword-Smith: OK. Defaults to four seconds.

neptune-core/src/application/loops/connect_to_peers.rs
Sword-Smith: Added the comment as a docstring to the function.

neptune-core/src/application/loops/main_loop.rs
Sword-Smith: The value is dropped when the connection to the peer is closed. That's when the total counter is incremented again to allow for one more connection. So the semaphore acquisition is released (counter incremented) when the `_permit` value is dropped from scope.

The drop happens when the spawned tokio task is being closed/`answer_peer` returns.

PR #743, opened 2025-10-21
network-security: Prevent infinite sync challenges

Sword-Smith:
An attack I spotted through the logs of a node: peer connects, peer sends a sync challenge which requires 20 blocks to be read from disk and sent over the network, and then disconnects. Continues this forever.

To mitigate this attack, each sync challenge is now punished at 50 points reputation. So after 20 such sync challenges from the same IP address, this IP address will be banned.

### DIFF
--- a/neptune-core/src/application/loops/peer_loop.rs
+++ b/neptune-core/src/application/loops/peer_loop.rs
@@ -734,6 +734,12 @@ impl PeerLoopHandler {
 
                     info!("Got sync challenge from {}", self.peer_address.ip());
 
+                    // Sync challenges are *always* punished to prevent a
+                    // malicious peer from spamming these, as they are expensive
+                    // to respond to.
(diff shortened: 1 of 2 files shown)


KaffinPX:
> There seems a danger here that some honest nodes could end up getting kind of permanently banned from the network.
> 
> In the past, on low bandwidth connection I saw my node repeatedly issuing sync challenges because it never got the responses in time simply due to bandwidth.
> 
> The problem is I think especially hazardous right now because there are only 2-3 known bootstrapping nodes to connect to (from README). So if those seed nodes ban a node, it becomes much harder for them to get on the network at all. For better resiliency, the network needs some kind of DNS system like bitcoin uses that provides a list of seed nodes.

I think eliminating nodes with "very bad" internet connection is okay as they affect P2P latency in a negative way

PR #744, opened 2025-10-21
refactor(peer_loop): Avoid receiving same tx many times

Sword-Smith:
Add a delay to each peer loop in an attempt to mitigate race conditions on the receival of transactions.


dan-da:
I think there is a code smell here, and possible inefficiencies.

fwiw, LLM independently agrees.  for consideration:
[gemini_analysis.md](https://github.com/user-attachments/files/23028764/gemini_analysis.md)

PR #745, opened 2025-10-22
json_rpc: Add `archival` namespace

KaffinPX:
- Fix faulty RpcBfes behavior and add more tests
- Allow `<Digest>/<N>` input directly on BlockSelector (without prefix)
- Add tests for BlockSelector
- Add archival calls parallel to chain_tip* calls (+ digest method)
- Check consistency of archival calls like on tip calls
- Adds a way to check if a block is on canonical chain

### DIFF
--- a/neptune-core/src/application/json_rpc/server/service.rs
+++ b/neptune-core/src/application/json_rpc/server/service.rs
@@ -369,10 +369,9 @@ pub mod tests {
 
     #[test_strategy::proptest(async = "tokio", cases = 5)]
     async fn get_block_calls_are_consistent(
-        #[strategy(0usize..8)] _num_inputs: usize,
         #[strategy(0usize..8)] _num_outputs: usize,
         #[strategy(0usize..8)] _num_announcements: usize,
-        #[strategy(txkernel::with_lengths(#_num_inputs, #_num_outputs, #_num_announcements, true))]
+        #[strategy(txkernel::with_lengths(0, #_num_outputs, #_num_announcements, true))]
     tx_block1: crate::protocol::consensus::transaction::transaction_kernel::TransactionKernel,
     ) {
         let mut rpc_server = test_rpc_server().await;


neptune-core/src/application/json_rpc/core/model/common.rs
```diff
@@ -92,12 +153,63 @@ mod tests {
...
-        let bytes = vec![1u8, 2, 3, 4];
-        let bfes = bytes_to_bfes(&bytes);
-
+        let bfes = vec![
+            BFieldElement::new(0),
+            BFieldElement::new(1),
+            BFieldElement::new(255),
+            BFieldElement::new(0x1234567890ABCDEF),
+            BFieldElement::new(u64::MAX),
+        ];
```
Sword-Smith: You can use `bfe_array!` here.

neptune-core/src/application/json_rpc/core/api/ops.rs
```diff
@@ -53,6 +53,27 @@ pub enum RpcMethods {
...
+    GetBlockHeader,
+
+    #[namespace(Namespace::Archival)]
+    GetBlockBody,
+
+    #[namespace(Namespace::Archival)]
+    GetBlockTransactionKernel,
+
+    #[namespace(Namespace::Archival)]
+    GetBlockAnnouncements,
```
Sword-Smith, line 85: Should we drop the "Get" prefix from these names? So `Block` instead of `GetBlock`?

KaffinPX: It's actually intentional, On Ethereum and similar JSON-RPC APIs, methods without the get_ prefix (like chain_height or node_network) return simple state values that reflect the node’s current state(Like getters on JavaScript), rather than fetching a separate data object

KaffinPX: Maybe I should document about this naming convention on somewhere?

neptune-core/src/state/wallet/address/mod.rs
```diff
@@ -5,7 +5,7 @@
 //!
 //! (especially since we now have a key type with no corresponding address)
 mod addressable_key;
-pub mod common;
+mod common;
```
skaunov, line 8: I know just the two reasons of making something less `pub`:
- cut the (docs) maintenance effort,
- protect from misuse.

And I'd be glad to learn an other case to it! Would you care to describe how this improves/develops something. :nerd_face:

KaffinPX: this is the protect from misuse case, check bfes_to_bytes and bytes_to_bfes functions -- they were specialized for address-only usage but I thought it was okay to use for all cases(This PR also fixes that behavior). On an old PR it was made public by me but it was a mistake for sure.

neptune-core/src/application/json_rpc/core/api/rpc.rs
```diff
@@ -56,4 +63,75 @@ pub trait RpcApi: Sync + Send {
...
+    }
+    async fn get_block_kernel_call(&self, request: GetBlockKernelRequest)
+        -> GetBlockKernelResponse;
+
+    async fn get_block_header(&self, selector: RpcBlockSelector) -> GetBlockHeaderResponse {
+        self.get_block_header_call(GetBlockHeaderRequest { selector })
+            .await
+    }
+    async fn get_block_header_call(&self, request: GetBlockHeaderRequest)
+        -> GetBlockHeaderResponse;
```
Sword-Smith, line 113: Can you show the categories here by just adding a comment? Like so:
/* Archival */

neptune-core/src/application/json_rpc/core/api/ops.rs
```diff
@@ -53,6 +56,33 @@ pub enum RpcMethods {
...
+
+    #[namespace(Namespace::Archival)]
+    IsBlockCanonical,
```
Sword-Smith, line 88: This doesn't cover all archival calls we currently have in the tarpc server. I can think of `utxo_digest`, `utxo_origin_block` and `block_digests_by_height` where the latter returns a **list** of block digests with a specific height, whereas you new endpoint `GetBlockDigest(block_height)` only returns the canonical block with that height. The former is nice to have in order to inspect forks/chain reorganizations.

neptune-core/src/application/json_rpc/core/model/common.rs
```diff
@@ -92,12 +154,55 @@ mod tests {
...
-        let bytes = vec![1u8, 2, 3, 4];
-        let bfes = bytes_to_bfes(&bytes);
-
+        let bfes = bfe_vec![
+            BFieldElement::new(0),
+            BFieldElement::new(1),
+            BFieldElement::new(255),
+            BFieldElement::new(0x1234567890ABCDEF),
+            BFieldElement::new(u64::MAX),
+        ];
```
Sword-Smith: I think you can just use `bfe_vec![0, 1, 255, 0x1234567890ABCDEF, u64::MAX]` here. No need for the `BFieldElement::new` constructor call.

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -89,17 +99,183 @@ impl RpcApi for RpcServer {
...
     }
+
+    async fn get_block_digest_call(
+        &self,
+        request: GetBlockDigestRequest,
+    ) -> GetBlockDigestResponse {
+        let state = self.state.lock_guard().await;
+
+        GetBlockDigestResponse {
+            digest: request.selector.as_digest(&state).await,
```
Sword-Smith: Danger!!!

You're acquiring the read lock on state twice here. If any other thread requests a write lock after the 1st read lock has been granted but before the 2nd has, that will cause a deadlock. Since no read locks are granted once a write lock is requested.

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -89,17 +99,183 @@ impl RpcApi for RpcServer {
...
+    }
+
+    async fn get_block_kernel_call(
+        &self,
+        request: GetBlockKernelRequest,
+    ) -> GetBlockKernelResponse {
+        let state = self.state.lock_guard().await;
+        let archival_state = state.chain.archival_state();
+
+        let kernel = match request.selector.as_digest(&state).await {
```
Sword-Smith: Deadlock!

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -89,17 +99,183 @@ impl RpcApi for RpcServer {
...
+    }
+
+    async fn get_block_header_call(
+        &self,
+        request: GetBlockHeaderRequest,
+    ) -> GetBlockHeaderResponse {
+        let state = self.state.lock_guard().await;
+        let archival_state = state.chain.archival_state();
+
+        let header = match request.selector.as_digest(&state).await {
```
Sword-Smith: Deadlock!

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -167,4 +342,94 @@ pub mod tests {
         let announcements = rpc_server.tip_announcements().await.announcements;
         assert_eq!(transaction_kernel.announcements, announcements);
     }
+
+    #[test_strategy::proptest(async = "tokio", cases = 5)]
+    async fn get_block_calls_are_consistent(
+        #[strategy(txkernel::with_lengths(0, 2, 2, true))]
```
Sword-Smith: These size parameters can also be randomized:
```rust
 #[strategy(0usize..8)] _num_inputs: usize,
 #[strategy(0usize..8)] _num_outputs: usize,
 #[strategy(0usize..8)] _num_announcements: usize,
#[strategy(txkernel::with_lengths(#_num_inputs, #_num_outputs, #_num_announcements, true))]
```

Sword-Smith (changes requested):
Please check my feedback. Not too happy with lock handling. Let's see if we can do better somehow.

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -89,17 +99,183 @@ impl RpcApi for RpcServer {
...
+        GetBlockDigestResponse {
+            digest: request.selector.as_digest(&state).await,
+        }
+    }
+
+    async fn get_block_call(&self, request: GetBlockRequest) -> GetBlockResponse {
+        let state = self.state.lock_guard().await;
+        let archival_state = state.chain.archival_state();
+
+        let block = match request.selector.as_digest(&state).await {
```
Sword-Smith, line 131: Not super happy with these bindings here. They make lock acquisition a bit hard to reason about for me. Can you maybe drop the binding of `archival_state`?

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -89,17 +99,207 @@ impl RpcApi for RpcServer {
...
+                .chain
+                .archival_state()
+                .block_height_to_block_digests(request.height.into())
+                .await,
+        }
+    }
+    async fn get_block_call(&self, request: GetBlockRequest) -> GetBlockResponse {
+        let state = self.state.lock_guard().await;
+
+        let block = match request.selector.as_digest(&state).await {
```
Sword-Smith, line 131: What do you think about this pattern instead? 

```rust

        let digest = request.selector.as_digest(&state).await;
        let block = match digest {
            ...

```

PR #748, opened 2025-10-23
Move `Semaphore::new` out of `loop` so it'd be used in a meaningful manner.

skaunov:
@Sword-Smith , it's first of all on your recent development. Idk if my review comment was just missed or intentionally skipped, but I took some time to look into that again and am still feel like this repair is needed. Sorry if I got something backwards and just bothering with this again. =(

### DIFF
--- a/neptune-core/src/application/loops/peer_loop.rs
+++ b/neptune-core/src/application/loops/peer_loop.rs
@@ -3589,16 +3589,16 @@ mod tests {
                     block_3.clone().try_into().unwrap(),
                 ))),
                 Action::Write(PeerMessage::BlockRequestByHash(block_2.hash())),
-
-                // Now make the interruption of the block reconciliation process
+                // ____
+                // Now make the interruption of the block reconciliation process.
                 Action::Read(PeerMessage::BlockNotification((&block_5).into())),
-
+                // ____
                 // Complete the block reconciliation process by requesting the last block
                 // in this process, to get back to a mutually known block.
                 Action::Read(PeerMessage::Block(Box::new(
                     block_2.clone().try_into().unwrap(),
                 ))),
-
+                // ____
                 // Then anticipate the request of the block that was announced
                 // in the interruption.
                 // Note that we cannot anticipate the response, as only the main
@@ -3703,13 +3703,10 @@ mod tests {
                     block_3.clone().try_into().unwrap(),
                 ))),
                 Action::Write(PeerMessage::BlockRequestByHash(block_2.hash())),
-
                 // Now make the interruption of the block reconciliation process
                 Action::Read(PeerMessage::PeerListRequest),
-
                 // Answer the request for a peer list
                 Action::Write(PeerMessage::PeerListResponse(expected_peer_list_resp)),
-                
                 // Complete the block reconciliation process by requesting the last block
                 // in this process, to get back to a mutually known block.
                 Action::Read(PeerMessage::Block(Box::new(


neptune-core/src/application/loops/main_loop.rs
```diff
@@ -1674,25 +1674,22 @@ impl MainLoopHandler {
...
-            // incoming connection is closed. This value is set much higher
-            // than the configured max number of peers since it's only intended
-            // to be used in case of heavy DOS.
-            let incoming_connections_limit = Arc::new(Semaphore::new(
-                self.global_state_lock.cli().max_num_peers * 2 + 4,
-            ));
+        /* Use a semaphore to limit number of incoming connections. Should only be relevant as a countermeasure against a DOS. Each
+        incoming connection must acquire a permit. If none is free, the below call to `acquire_owned` will only be resolved when an
+        incoming connection is closed. This value is set much higher than the configured max number of peers since it's only intended
+        to be used in case of heavy DOS. */
```
Sword-Smith: Can you limit comment line width to 80 characters?

neptune-core/src/application/loops/main_loop.rs
```diff
@@ -1674,25 +1674,22 @@ impl MainLoopHandler {
...
 
+        let exit_code: i32 = loop {
             select! {
                 Ok(()) = signal::ctrl_c() => {
                     info!("Detected Ctrl+c signal.");
                     break SUCCESS_EXIT_CODE;
                 }
 
-                // Monitor for SIGTERM, SIGINT, and SIGQUIT.
+                // Monitor for `SIGTERM`, `SIGINT`, and `SIGQUIT`.
```
Sword-Smith: I prefer these comments as they were, without the backticks, and without the period at the end.

neptune-core/src/application/loops/peer_loop.rs
```diff
@@ -4476,8 +4476,7 @@ mod tests {
         #[traced_test]
         #[apply(shared_tokio_runtime)]
         async fn bad_sync_challenge_height_greater_than_tip() {
-            // Criterium: Challenge height may not exceed that of tip in the
-            // request.
+            // Criterium: Challenge height may not exceed that of tip in the request.
```
Sword-Smith: This exceeds max comment length of 80 characters.

Sword-Smith (changes requested):
Happy to accept this PR. Please keep PR to the bug fix though, and don't change comments, or exceed max line width for comment lines of 80 characters.

You getting rid of a dependency? Cool if we don't need it! Can you make a separate commit for that though? Same PR is OK. Also: You need to commit the updated `Cargo.lock` file that results from the deleted dependency.

neptune-core/src/application/loops/peer_loop.rs
skaunov: That's interesting: `rustfmt` thinks it *is* 80 width, I guess it counts the width of the comment itself, not where it ends; and it seems like it's a good idea to stick to its behaviour.

jan-ferdinand: Note that rustfmt's options [`comment_width`](https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=comment_width#comment_width) and [`wrap_comments`](https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=wrap_comments) are unstable, which means that unfortunately, you won't get automatic comment formatting when using the stable toolchain. (Only) if using nightly, you can configure rustfmt accordingly.

skaunov:
@jan-ferdinand , thank you for the example. (Btw, it's a sad surprise that replying in email kicks the discussion out of that suggestion thread. :disappointed:)

"not prettily" is quite the description of it! You're the maintainers... In my opinion this works, especially given it will stabilize eventually. 

I consider the bigger problem is that someone will need to review those ten dozens of files affected. I went over few and they have tricky stuff sometimes.

Sword-Smith:
Thanks for this PR. I decided to just cherry-pick two changes: the fix for the semaphore, and the removal of the unneeded dependency. You are credited with both those commits.

I'm OK talking about comment formatting in general but then it should be in a separate commit, and avoid some of the changes introduced here like adding periods after short imperative sentences, changing `/* */`  to `//` etc.

PR #752, opened 2025-10-29
feat(cli_args): add min relay per input for pc-backed txs

Sword-Smith:
Allow users to set a min-relay fee for proof collection-backed transactions. If the fee per input for a proof collection-backed transaction is less than this value, the transaction is never requested, never stored to the mempool, and never relayed to other peers.

Concrete numbers:
The default value is set to 0.0005 NPT / input. So for a standard 2-in/2-out transaction, the minimum relay fee is 0.001 NPT, 1/1000 NPT. For a PC-backed transaction with 100 inputs, the minimum relay fee is 0.05 NPT, 1/20 NPT.

These default values can obviously be changed in the future if the price of NPT was to rise significantly.

### DIFF
--- a/neptune-core/src/application/config/cli_args.rs
+++ b/neptune-core/src/application/config/cli_args.rs
@@ -881,6 +881,51 @@ mod tests {
     }
 
     #[test]
+    fn relay_transaction_unit_test() {
+        let cli = Args::default();
+        assert!(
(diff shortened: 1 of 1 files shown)


Sword-Smith:
@dan-da How should we handle the case where the node generates a transaction that does not meet its own relay criteria?

PR #754, opened 2025-10-31
feat(dashboard): allow clearing peer standing through dashboard ui

imalfect:
this PR adds the clear peer standing functionality to the dashboard.

by clicking 'x', the focused peer's standing is reset to 0
I didn't add a notifier label (like in the send screen) as the result is visible immediately (refresh is triggered upon action completion), but can be added if needed.

To support this action I added a `clear_peer_standing` function to the dashboard rpc client which bridges `clear_standing_by_ip`

### DIFF
--- a/neptune-core/src/protocol/peer.rs
+++ b/neptune-core/src/protocol/peer.rs
@@ -1016,6 +1016,9 @@ impl rand::distr::Distribution<NegativePeerSanction> for rand::distr::StandardUn
             31 => NegativePeerSanction::UnwantedMessage,
 
             32 => NegativePeerSanction::NoStandingFoundMaybeCrash,
+
+            33 => NegativePeerSanction::ReceivedSyncChallenge,
+            34 => NegativePeerSanction::UnrelayableTransaction,
             _ => unreachable!(),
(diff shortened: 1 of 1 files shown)


imalfect:
thanks, fixed!

to accommodate for the functionality, I moved the mock data generation from rpc functions to a state that the mockup rpc client now has, thank you @aszepieniec for help :)

note: instead of clearing peer standing with mocked data it replaces the peer with another random one (so any UI interaction is shown), since the clear_standing function from peer.standing cannot be interacted with outside of the core crate.

aszepieniec:
Here is something that *does* seem related to this PR. In mock mode, press x several times. After a few times I observe this:

```
thread 'tokio-runtime-worker' (36986) panicked at /home/alan/repos/neptune-core/neptune-core/src/protocol/peer.rs:1019:18:
internal error: entered unreachable code
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
Aborted (core dumped)
```

aszepieniec (approved):
LGTM

PR #757, opened 2025-11-02
UX: Add banner

aszepieniec:
Adds a banner as the first thing in the standard output of `neptune-core`.


aszepieniec:
Resuming from #642.

IIRC the objections were
 - [x] wait for crate separation (done)
 - [x] introduces unnecessary* dependency on `owo-colors` (I took the liberty to drop this dependency)

So we are ready for merging right?

Sword-Smith (approved):

PR #763, opened 2025-11-09
JSON-RPC: Add UTXO-related calls and `--unsafe-rpc`

KaffinPX:
- Prepare server for multiple transports (a simple refactorization).
- Add an extra restriction mechanism to RPC for scenarios like limiting search depth on `FindUtxoOrigin` calls.
- Two new RPC methods: `GetUtxoDigest` and `FindUtxoOrigin`.
- Expose `AdditionRecord` to RPC properly.

### DIFF
--- a/neptune-core/src/application/config/cli_args.rs
+++ b/neptune-core/src/application/config/cli_args.rs
@@ -559,10 +559,15 @@ pub struct Args {
     )]
     pub rpc_modules: Vec<Namespace>,
 
-    /// Allow unsafe RPC methods over all transports (e.g., HTTP).
+    /// Enable unsafe RPC methods over all transports (e.g., HTTP).
     ///
-    /// WARNING: Enabling this will expose dangerous RPC methods,
-    /// Use this only in isolated, controlled environments.
+    /// WARNING: Enabling this exposes dangerous RPC behavior and should only be used in
+    /// isolated, controlled environments.
+    ///
+    /// This may result in:
+    /// - Denial of Service (DoS) by triggering heavy or expensive computations.
+    /// - Exposure of internal node state or metrics that could aid an attacker.
+    /// - Misconfigurations leading to unexpected or unsafe behavior.
     #[clap(long)]
     pub unsafe_rpc: bool,
 }
--- a/neptune-core/src/application/json_rpc/core/model/block/transaction_kernel.rs
+++ b/neptune-core/src/application/json_rpc/core/model/block/transaction_kernel.rs
@@ -51,8 +51,8 @@ impl From<RemovalRecord> for RpcRemovalRecord {
 #[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, Default)]
 pub struct RpcAdditionRecord(pub Digest);
 
-impl From<&AdditionRecord> for RpcAdditionRecord {
-    fn from(record: &AdditionRecord) -> Self {
+impl From<AdditionRecord> for RpcAdditionRecord {
+    fn from(record: AdditionRecord) -> Self {
         Self(record.canonical_commitment)
     }
 }
@@ -95,7 +95,7 @@ impl From<&TransactionKernel> for RpcTransactionKernel {
     fn from(kernel: &TransactionKernel) -> Self {
         Self {
             inputs: kernel.inputs.clone().into_iter().map(Into::into).collect(),
-            outputs: kernel.outputs.iter().map(Into::into).collect(),
+            outputs: kernel.outputs.iter().copied().map(Into::into).collect(),
             announcements: kernel
                 .announcements
                 .iter()
--- a/neptune-core/src/application/json_rpc/core/model/message.rs
+++ b/neptune-core/src/application/json_rpc/core/model/message.rs
@@ -234,7 +234,7 @@ pub struct IsBlockCanonicalResponse {
 #[derive(Clone, Copy, Debug, Serialize_tuple, Deserialize_tuple)]
 #[serde(rename_all = "camelCase")]
 pub struct GetUtxoDigestRequest {
-    pub index: u64,
+    pub leaf_index: u64,
 }
 
 #[derive(Clone, Copy, Debug, Serialize, Deserialize)]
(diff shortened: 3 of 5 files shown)


neptune-core/src/application/config/cli_args.rs
```diff
@@ -558,6 +558,13 @@ pub struct Args {
         value_name = "NAMESPACES"
     )]
     pub rpc_modules: Vec<Namespace>,
+
+    /// Allow unsafe RPC methods over all transports (e.g., HTTP).
+    ///
+    /// WARNING: Enabling this will expose dangerous RPC methods,
+    /// Use this only in isolated, controlled environments.
```
Sword-Smith: So what exactly can happen here if set? For now it's just a DoS, right? I mean, there are no secret keys that can be leaked if this is set, I think. Maybe use some more words on what the dangers are here.

neptune-core/src/application/json_rpc/core/model/block/transaction_kernel.rs
```diff
@@ -47,11 +48,28 @@ impl From<RemovalRecord> for RpcRemovalRecord {
     }
 }
 
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, Default)]
+pub struct RpcAdditionRecord(pub Digest);
```
Sword-Smith, line 52: This data struct exists to make the JSON representation of an addition record smaller? Such that it's only a digest and not a struct with a field whose value is a digest?

neptune-core/src/application/json_rpc/core/model/block/transaction_kernel.rs
```diff
@@ -47,11 +48,28 @@ impl From<RemovalRecord> for RpcRemovalRecord {
...
+
+impl From<&AdditionRecord> for RpcAdditionRecord {
+    fn from(record: &AdditionRecord) -> Self {
+        Self(record.canonical_commitment)
+    }
+}
```
Sword-Smith: Addition record is `Copy` so maybe implement `From<AdditionRecord>` instead here?

neptune-core/src/application/json_rpc/core/model/block/transaction_kernel.rs
KaffinPX: Yup, to avoid nesting to be exact

neptune-core/src/application/json_rpc/core/model/block/transaction_kernel.rs
KaffinPX: I think I made it reference so I could use .into() directly on iterator (it was being returned as a reference because its fetched over a getter fn)

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -301,6 +301,43 @@ impl RpcApi for RpcServer {
...
+
+    async fn get_utxo_digest_call(
+        &self,
+        request: GetUtxoDigestRequest,
+    ) -> RpcResult<GetUtxoDigestResponse> {
+        let state = self.state.lock_guard().await;
+        let aocl = &state.chain.archival_state().archival_mutator_set.ams().aocl;
+
+        Ok(GetUtxoDigestResponse {
+            digest: match request.index > 0 && request.index < aocl.num_leafs().await {
```
Sword-Smith: Off-by-one error here. `request.index = 0` is allowed.

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -301,6 +301,43 @@ impl RpcApi for RpcServer {
...
+                true => Some(aocl.get_leaf_async(request.index).await),
+                false => None,
+            },
+        })
+    }
```
Sword-Smith, line 315: Let's use `try_get_leaf` instead. This way you avoid having to re-implement a bounds-check.

```rust
    async fn get_utxo_digest_call(
        &self,
        request: GetUtxoDigestRequest,
    ) -> RpcResult<GetUtxoDigestResponse> {
        let aocl = &state.chain.archival_state().archival_mutator_set.ams().aocl;

        let maybe_digest = aocl.try_get_leaf(request.index).await;
        Ok(GetUtxoDigestResponse {
            digest: maybe_digest,
        })
    }
```

neptune-core/src/application/json_rpc/core/model/message.rs
```diff
@@ -230,3 +230,30 @@ pub struct IsBlockCanonicalRequest {
 pub struct IsBlockCanonicalResponse {
     pub canonical: bool,
 }
+
+#[derive(Clone, Copy, Debug, Serialize_tuple, Deserialize_tuple)]
+#[serde(rename_all = "camelCase")]
+pub struct GetUtxoDigestRequest {
+    pub index: u64,
```
Sword-Smith: An MMR has multiple concepts of indices: node index, leaf index, Merkle-tree index. This index here is being used as a leaf index, so let's rename it to `leaf_index`.

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -301,6 +301,43 @@ impl RpcApi for RpcServer {
...
+
+    async fn find_utxo_origin_call(
+        &self,
+        request: FindUtxoOriginRequest,
+    ) -> RpcResult<FindUtxoOriginResponse> {
+        let allowed_search_depth = if self.unrestricted {
+            request.search_depth
+        } else {
+            Some(request.search_depth.unwrap_or(100).min(100))
```
Sword-Smith, line 324: Do you want to log a warning if a greater depth is requested than what's allowed? I.e. if the value is bounded by the 100.

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -522,4 +558,65 @@ pub mod tests {
...
+    #[test_strategy::proptest(async = "tokio", cases = 5)]
+    async fn utxo_calls_are_consistent(
+        #[strategy(0usize..8)] _num_outputs: usize,
+        #[strategy(txkernel::with_lengths(0usize, #_num_outputs, 0usize, true))]
+        transaction_kernel: crate::protocol::consensus::transaction::transaction_kernel::TransactionKernel,
+    ) {
+        let mut rpc_server = test_rpc_server().await;
+
+        // Before new block check size of aocl leaves so we can know exact index of new outputs
+        let aocl_leaves = rpc_server
```
Sword-Smith: Rename to `num_aocl_leafs`

neptune-core/src/application/json_rpc/server/service.rs
KaffinPX: Someone might spam this call with more than allowed depth to annoy RPCs maintainer so let's dont I think

PR #765, opened 2025-11-09
fix(dashboard): Mock RPC call `latest-address`

aszepieniec:
Commits 489853b and 9c35136 add a RPC end-point for getting the most recent address and displaying it by default in the dashboard. However, the mock RPC server, which is used when running the dashboard in mock mode, does not know this method and consequently when it is called control is passed to the real RPC server, leading to a crash.

This commit mocks the endpoint, fixing the cash.


jan-ferdinand:
> fixing the cash

Nice one.

imalfect (approved):
lgtm!

PR #766, opened 2025-11-09
Small Dashboard Features

aszepieniec:
1. Escape focus in Peers screen by pressing `q`.
 2. Jump up/down by one page using PgUp/PgDn.

### DIFF
--- a/neptune-dashboard/src/mock_rpc_client.rs
+++ b/neptune-dashboard/src/mock_rpc_client.rs
@@ -57,7 +57,7 @@ impl MockRpcClient {
     fn init_mock_data() -> MockState {
         let mut rng = StdRng::from_seed(rng().random());
 
-        let num_peers = 100 + rng.random_range(1..10);
+        let num_peers = 100;
         let peers: Vec<PeerInfo> = (0..num_peers).map(|_| rng.random()).collect();
 
         let num_utxos = rng.random_range(1..20);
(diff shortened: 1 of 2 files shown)


neptune-dashboard/src/mock_rpc_client.rs
```diff
@@ -57,7 +57,7 @@ impl MockRpcClient {
     fn init_mock_data() -> MockState {
         let mut rng = StdRng::from_seed(rng().random());
 
-        let num_peers = rng.random_range(1..10);
+        let num_peers = 100 + rng.random_range(1..10);
```
imalfect: maybe we just leave `num_peers` at 100 at this point?

neptune-dashboard/src/peers_screen.rs
```diff
@@ -611,4 +613,44 @@ impl Events {
...
+    /// Get the height of the terminal.
+    fn terminal_height() -> u16 {
+        if let Ok((_cols, rows)) = crossterm::terminal::size() {
+            rows
+        } else {
+            1
+        }
+    }
+
+    /// Jump the selector down by a page.
```
imalfect, line 626: perhaps we should add a comment that in this context a page is half of the visible rows? as per
```rust
let jump_height = usize::max(1, num_rows / 2);
```

aszepieniec:
> maybe we could add pagination to other _list based_ pages as well? seeing todos on the other files about that too? can spin up a PR if alright with you as it's mostly copy-paste

Sure, please do.

imalfect (approved):

PR #774, opened 2025-11-13
upgrade wallet-server

code-pangolin:
keep wallet-server track with master branch


Sword-Smith:
> what do you mean let me control the code, i should control the own branch and dont need to make pr to this repo next time?

I think that would be easier, no? @KaffinPX should be able to add the endpoints you need eventually, so we can get rid of this branch. Alternatively you can look into changing to using his RPC server instead of running your own.

Yes. I rebased. If you have some changes after I've rebased I'd be happy to accept those into the `wallet-server` branch if you prefer that.

PR #777, opened 2025-11-17
feat(json_rpc): Add methods for off-node wallets

KaffinPX:
Adds 3 calls to ``Wallet`` namespace allowing off-node wallets(e.g. browser wallets that doesn't run a node) over RPC.

`GetBlocks` is a specialized call for scanning announcements etc., it includes `proof_leaf` instead of proof as "commitment".
`RestoreMembershipProof` can be used to restore MSMPs (for those wallets that doesn't manage the proofs themselves).
and `SubmitTransaction` for submitting ProofCollection/SingleProof transactions.

- Exposes `BlockAppendix` on BlockKernel-related calls as preparation for hash calculation method on RpcBlock impls.

### DIFF
--- a/neptune-core/src/application/json_rpc/core/model/block/transaction_kernel.rs
+++ b/neptune-core/src/application/json_rpc/core/model/block/transaction_kernel.rs
@@ -158,13 +158,7 @@ impl From<&TransactionKernel> for RpcTransactionKernel {
 
 impl From<RpcTransactionKernel> for TransactionKernel {
     fn from(kernel: RpcTransactionKernel) -> Self {
-        TransactionKernelProxy::from(kernel).into_kernel()
-    }
-}
-
-impl From<RpcTransactionKernel> for TransactionKernelProxy {
-    fn from(kernel: RpcTransactionKernel) -> Self {
-        Self {
+        let kernel_proxy = TransactionKernelProxy {
             inputs: kernel.inputs.into_iter().map(Into::into).collect(),
             outputs: kernel.outputs.into_iter().map(Into::into).collect(),
             announcements: kernel.announcements.into_iter().map(Into::into).collect(),
@@ -173,6 +167,8 @@ impl From<RpcTransactionKernel> for TransactionKernelProxy {
             timestamp: kernel.timestamp,
             mutator_set_hash: kernel.mutator_set_hash,
             merge_bit: kernel.merge_bit,
-        }
+        };
+
+        kernel_proxy.into_kernel()
     }
 }
--- a/neptune-core/src/application/json_rpc/core/model/wallet/transaction.rs
+++ b/neptune-core/src/application/json_rpc/core/model/wallet/transaction.rs
@@ -128,7 +128,9 @@ impl From<RpcTransaction> for Transaction {
 impl From<Transaction> for RpcTransaction {
     fn from(tx: Transaction) -> Self {
         RpcTransaction {
-            kernel: RpcTransactionKernel::from(&tx.kernel), // TODO: Two impls when possible, ref clones and other takes ownership...
+            // TODO: Two impls when possible, ref clones and other takes
+            // ownership...
+            kernel: RpcTransactionKernel::from(&tx.kernel),
             proof: tx.proof.into(),
         }
     }
--- a/neptune-core/src/application/json_rpc/server/service.rs
+++ b/neptune-core/src/application/json_rpc/server/service.rs
@@ -511,12 +511,10 @@ pub mod tests {
     use crate::Block;
 
     pub async fn test_rpc_server() -> RpcServer {
-        let global_state_lock = mock_genesis_global_state(
-            2,
-            WalletEntropy::new_random(),
-            cli_args::Args::default_with_network(Network::Main),
-        )
-        .await;
+        let mut cli = cli_args::Args::default_with_network(Network::Main);
+        cli.tx_proving_capability = Some(TxProvingCapability::ProofCollection);
+        let global_state_lock =
+            mock_genesis_global_state(2, WalletEntropy::new_random(), cli).await;
 
         RpcServer::new(global_state_lock, None)
     }


neptune-core/src/application/json_rpc/core/model/wallet/block.rs
```diff
@@ -0,0 +1,24 @@
...
+
+use crate::application::json_rpc::core::model::block::RpcBlockKernel;
+use crate::protocol::consensus::block::Block;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RpcWalletBlock {
+    pub kernel: RpcBlockKernel,
+    pub proof_leaf: Digest,
+}
```
Sword-Smith, line 15: This struct needs a `hash` method such that the block's hash can be calculated without access to the proof. A good test for this `hash` method would be that it agrees with the `Block` value from this the `RpcWalletBlock` is derived.

neptune-core/src/application/json_rpc/core/model/wallet/mutator_set.rs
```diff
@@ -0,0 +1,50 @@
...
+        )
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct MsMembershipProof {
+    pub aocl_auth_paths: RpcIndexedAoclAuthPaths,
+    pub target_chunks: RpcChunkDictionary,
+}
```
Sword-Smith, line 49: This data structure is not an `MsMembershipProof` but rather a privacy-preserving data structure that allows that caller to extract the real `MsMembershipProof` (name collision) value. Do you get how it works?

The caller only announces the absolute index set of their UTXO to the server, and not the UTXO's AOCL leaf index. The server then responds with a data structure that contains all the required information for the caller to build the real `MsMembershipProof`. That means all the target chunks associated with the absolute index set as well as all possible AOCL authentication paths. The caller can then extact the single AOCL authentication path they care about once they've received the data. Have a look at the `MsMembershipProofPrivacyPreserving` in `neptune-core/src/util_types/mutator_set/archival_mutator_set.rs` where this "extractor" `extract_ms_membership_proof` is defined.

So what does that mean for this PR? 
1) The name of this data structure should communicate that this is not a membership proof but related to a membership proof and related to the preservation of privacy.
2) This structure needs a way to extract the real membership proof, a method like `MsMembershipProofPrivacyPreserving::extract_ms_membership_proof`.

neptune-core/src/application/json_rpc/core/model/wallet/mutator_set.rs
```diff
@@ -4,36 +4,75 @@ use serde::Deserialize;
...
+pub struct RpcMsMembershipProofPrivacyPreserving {
     pub aocl_auth_paths: RpcIndexedAoclAuthPaths,
     pub target_chunks: RpcChunkDictionary,
 }
 
-impl From<MsMembershipProofPrivacyPreserving> for MsMembershipProof {
+impl RpcMsMembershipProofPrivacyPreserving {
+    /// Build the required membership proof by supplying the correct AOCL leaf
+    /// index to extract the right MMR authentication path and the missing
+    /// cryptographic data.
```
Sword-Smith: How about "extract the right mutator set membership proof from the privacy-preserving recovery data. Picks the right AOCL MMR authentication path from the provided list."?

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -754,4 +764,111 @@ pub mod tests {
...
+                mock_amount,
+                Timestamp::now(),
+            )
+            .await
+            .unwrap();
+
+        // Pass transaction into rpc_server network.
+        let devnet_mempool = &devnet_node.lock_guard().await.mempool;
+        let transaction = devnet_mempool.get(artifacts.transaction().txid()).unwrap();
+        let tip =
```
Sword-Smith: Easier if you call this `block_1` IMO. That communicates its height which makes the test a bit easier to read.

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -754,4 +764,111 @@ pub mod tests {
...
+        let tx_creation_config = TxCreationConfig::default()
+            .with_prover_capability(TxProvingCapability::ProofCollection);
+        let test_artifacts = rpc_server
+            .state
+            .api()
+            .tx_initiator_internal()
+            .create_transaction(
+                Default::default(),
+                mock_amount,
+                Timestamp::now(),
```
Sword-Smith: Since you're constructing proofs here, you should allow for the proofs to be reused across test runs. To do that, you need to use a deterministic timestamp here. Probably also for the above timestamp in "artifacts" since there is a dependency between the two timestamps AFAICS. You can e.g. use something like launch_date + 7 months

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -754,4 +764,111 @@ pub mod tests {
...
     }
+
+    #[apply(shared_tokio_runtime)]
+    async fn off_node_wallets_behave_correctly() {
+        let mut rpc_server = test_rpc_server().await;
+
+        // Prepare a transaction to our wallet coming from devnet wallet.
+        let wallet = WalletEntropy::devnet_wallet();
+        let mut devnet_node =
+            mock_genesis_global_state(0, wallet, rpc_server.state.cli().clone()).await;
```
Sword-Smith: So there are two parties involved here? Would it make sense to call them e.g. "alice" and "bob". And the transaction variables (`artifacts`, `transaction`) `tx_from_alice_to_bob` or something like this? I find the tests a bit easier to read if the two parties have human names and not just variable names showing what type they are.

neptune-core/src/application/json_rpc/core/model/wallet/block.rs
KaffinPX: Lets do this on next PR with TODO cleanups?

Sword-Smith: Here's how this was handled on the `wallet-server` branch, with a method on `ExportedBlock`:

```rust
    /// Calculate the block hash without assuming that the proof is valid.
    fn mast_hash(&self) -> Digest {
        let block_header_leaf = Tip5::hash_varlen(&self.kernel.header.mast_hash().encode());
        let body_leaf = Tip5::hash_varlen(&self.kernel.body.mast_hash().encode());
        let appendix_leaf = Tip5::hash_varlen(&self.kernel.appendix.encode());
        let kernel_leafs = [
            block_header_leaf,
            body_leaf,
            appendix_leaf,
            Digest::default(),
        ];
        let kernel_hash = MerkleTree::sequential_frugal_root(&kernel_leafs).unwrap();
        let block_leafs = [Tip5::hash_varlen(&kernel_hash.encode()), self.proof_leaf];

        MerkleTree::sequential_frugal_root(&block_leafs).unwrap()
    }
```

neptune-core/src/application/json_rpc/server/service.rs
KaffinPX: That step is very "isolated" actually, I don't know

neptune-core/src/application/json_rpc/core/model/wallet/block.rs
KaffinPX: Yeah I know, just BFieldCodec and its implementations might take some time (Exposed appendix for preparation)

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -335,6 +341,139 @@ impl RpcApi for RpcServer {
...
     }
+
+    async fn get_blocks_call(&self, request: GetBlocksRequest) -> RpcResult<GetBlocksResponse> {
+        // Reverse get_blocks is not supported yet.
+        // Might be reconsidered after "succinctness" as it might give it a purpose.
+        if request.to_height < request.from_height {
+            return Ok(GetBlocksResponse { blocks: Vec::new() });
+        }
+
+        let max_blocks = if self.unrestricted { usize::MAX } else { 100 };
```
Sword-Smith, line 352: I'm wondering if 100 is a bit low for this number. But let's go with it for now, and then maybe later it could be increased to 200, 500, or even 1000.

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -335,6 +341,139 @@ impl RpcApi for RpcServer {
...
+        }
+
+        Ok(GetBlocksResponse { blocks })
+    }
+
+    async fn restore_membership_proof_call(
+        &self,
+        request: RestoreMembershipProofRequest,
+    ) -> RpcResult<RestoreMembershipProofResponse> {
+        if request.absolute_index_sets.len() > 256 && !self.unrestricted {
```
Sword-Smith, line 384: This is a fine limit, I think. You certainly don't want to create proof collection backed transactions with more than 256 inputs. neptune-core has a hard cap of 200 inputs -- meaning that proof collection backed transactions with more than 200 inputs will never be relayed. If someone has the hardware for it (you need at least 2TB RAM for this), they are of course free to create a single proof backed transaction from a proof collection with more than 200 inputs. But the better way of creating such big transactions is to create smaller single proof backed transactions and then merge those.

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -335,6 +341,139 @@ impl RpcApi for RpcServer {
...
+            return Err(RpcError::RestoreMembershipProof(
+                RestoreMembershipProofError::ExceedsAllowed,
+            ));
+        }
+
+        let state = self.state.lock_guard().await;
```
Sword-Smith, line 390: Read-lock is held till end of function, so returned data is guaranteed to be consistent 👍

neptune-core/src/application/loops/main_loop.rs
```diff
@@ -2066,6 +2066,30 @@ impl MainLoopHandler {
...
+                // Until a decision is made. (As this might get used by api too.)
+                let notification: Option<TransactionNotification> =
+                    transaction.as_ref().try_into().ok();
+
+                {
+                    let mut state = self.global_state_lock.lock_guard_mut().await;
+
+                    state
+                        .mempool_insert(*transaction, UpgradePriority::Critical)
+                        .await;
```
Sword-Smith, line 2080: I wonder if `Critical` is the right value to use here. Let's keep it for now. But it will mean that this node will put in a big effort into upgrading the proof of the transaction. Perhaps we'd eventually want an endpoint to submit transactions where the node calculates its economic interest in a transaction and then uses this to ascribe a priority? An endpoint similar to that of [`incentivized_proof_collection_transaction` on the wallet-server branch](https://github.com/Neptune-Crypto/neptune-core/blob/ef59f1b0b9e4e41f807f8b68c245ae7fd51cc50f/neptune-core/src/application/rest_server/mod.rs#L483C10-L483C51).

neptune-core/src/application/json_rpc/core/model/block/mod.rs
```diff
@@ -12,29 +15,67 @@ pub mod body;
...
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct RpcBlockAppendix(Vec<RpcClaim>);
+
+impl From<&BlockAppendix> for RpcBlockAppendix {
+    fn from(appendix: &BlockAppendix) -> Self {
+        RpcBlockAppendix(
+            appendix
+                ._claims()
```
Sword-Smith, line 45: Yikes. Ugly method name. Not your fault though. Let's just keep it like this and fix in separate commit. Method name should obviously be `claims` without the prepended underscore.

neptune-core/src/application/json_rpc/core/model/block/transaction_kernel.rs
```diff
@@ -109,3 +155,24 @@ impl From<&TransactionKernel> for RpcTransactionKernel {
...
     }
 }
+
+impl From<RpcTransactionKernel> for TransactionKernel {
+    fn from(kernel: RpcTransactionKernel) -> Self {
+        TransactionKernelProxy::from(kernel).into_kernel()
+    }
+}
+
+impl From<RpcTransactionKernel> for TransactionKernelProxy {
```
Sword-Smith: Is this conversion method needed? If you're not using it, please delete. As far as I can see this doesn't make a lot of sense since `TransactionKernelProxy` is an interface for changing values in an already-existing transaction kernel.

neptune-core/src/application/json_rpc/core/model/wallet/transaction.rs
```diff
@@ -0,0 +1,135 @@
...
+            kernel: tx.kernel.into(),
+            proof: tx.proof.into(),
+        }
+    }
+}
+
+impl From<Transaction> for RpcTransaction {
+    fn from(tx: Transaction) -> Self {
+        RpcTransaction {
+            kernel: RpcTransactionKernel::from(&tx.kernel), // TODO: Two impls when possible, ref clones and other takes ownership...
```
Sword-Smith: This comment exceeds max allowed line width. Max line width for code is 100. Max line width for comments is 80. The formatter handles the former correctly. Maybe move comment up to previous line, like so?

```rust
        RpcTransaction {
            // TODO: Two impls when possible, ref clones and other takes
            // ownership...
            kernel: RpcTransactionKernel::from(&tx.kernel),
            proof: tx.proof.into(),
        }
```

neptune-core/src/application/json_rpc/server/service.rs
KaffinPX: If they somehow exceed that limit they can do multiple requests anyway, this limit is actually to prevent OOMs.

neptune-core/src/application/loops/main_loop.rs
KaffinPX: Let's plan it with `Upgrader` namespace?

neptune-core/src/application/json_rpc/core/model/block/transaction_kernel.rs
KaffinPX: It was needed as it was impossible to convert one into a TransactionKernel directly without the proxy (because of readonly macro)

Sword-Smith: What if the convertion method just uses TransactionKernelProxy internally then?

Sword-Smith (approved):

PR #779, opened 2025-11-17
dashboard: cleanup table navigation and implement pagination on the remaining screens 

imalfect:
This PR creates a new `scrollable_table.rs` file which contains utility functions that implement table pagination and navigation. Table navigation logic in screens was replaced with it, which reduces code duplication. On top of that, it also adds the pagination feature (which was recently implemented in the peers screen) to other screens that have interactive tables.

### DIFF
--- a/neptune-dashboard/src/address_screen.rs
+++ b/neptune-dashboard/src/address_screen.rs
@@ -123,11 +123,7 @@ impl AddressScreen {
                         return None;
                     }
 
-                    match key.code {
-                        _ => {
-                            escalate_event = Some(event);
-                        }
-                    }
+                    escalate_event = Some(event);
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


imalfect:
1. 
Screenshot 2025-11-20 at 00 57 36
I see it, same seems to happen on the master branch, could it be because of the label that shows the number of peers connected? It's the only screen with a label like that, and other table-screens render just fine on that size, I'll have a deeper look on the _layout math_ tomorrow

2. I think the original author of that file left it out, certainly not a problem though, I can add that functionality

once solved I'll include both as commits to this PR

aszepieniec:
@imalfect: Are you still working on this PR?

aszepieniec:
Squashed and merged in 3a4d18d4acfb6d873964b71d016ec620363789e1. Two identified issues should not block merger and are factored out as #857 and #856.

PR #783, opened 2025-11-23
Add CLI Tools for Identifying Incoming UTXOs

aszepieniec:
Adds two CLI commands, the second one with RPC and application support:

 1. `neptune-cli index-of <address>` Computes the derivation index from the given address.
 2. `neptune-cli rescan --block 1337 --address <address>` Re-scans block `1337` to look for output UTXOs that can be spent by the secret spending key associated with the given address, and if any are found, adds them to the wallet database and adds their randomness to `incoming_randomness.dat`.

### DIFF
--- a/neptune-core-cli/src/main.rs
+++ b/neptune-core-cli/src/main.rs
@@ -320,7 +320,7 @@ enum Command {
     ///
     /// Usage:
     ///
-    ///  > neptune-cli rescan --block 13 --address <address>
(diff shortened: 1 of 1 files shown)


aszepieniec:
@Sword-Smith: which parts, if any,of this PR are obviated by the new branch for faster wallet updates? If all, then let's close the PR without merging.

PR #784, opened 2025-11-24
Track Mutxo Out of Bounds Bug

aszepieniec:


Sword-Smith:
I suspect the bug you're hunting has to do with a database corruption -- one database was written to but another was not. But these debugging steps cannot hurt.

I think the real solution is going to be some checks on database consistency at startup.

PR #787, opened 2025-11-25
Separate Sync Loop with Parallel Block Downloading

aszepieniec:
## Sync Loop

The sync loop is a separate event loop that is spawned when the node determines that at least one peer is legitimately very far ahead. The sync loop is responsible for managing state related to the syncing process. The sync loop only communicates with the main loop, which is responsible for forwarding messages to peer and from peer loops. The sync loop stores blocks in a temporary directory without validating them. When all blocks are in, the sync loop validates all blocks and feeds them to the main loop in order. At this point, the sync loop is dropped.

## Differences Relative to Prior Architecture

 - The previous sync mode was embedded into the main loop. The model proposed here has a separate event loop that can be (and is) tested in isolation.
 - The previous sync process would select a single peer to download all blocks from. The model proposed here can request blocks from any number of peers.
 - The previous sync process needed the blocks to come in in order. The model proposed here can tolerate blocks coming in in any order, as long as they all come in eventually.
 - The previous sync process would request batches of blocks with height starting at one plus tip height. The model proposed here requests one block at a time by random height. While not supported yet, the randomness is designed to support syncing from peers who themselves are syncing.
 - The previous sync process had no mechanism of resuming a half-completed sync. The model proposed here stores the blocks in a temporary directory which can survive across runs of the client and can be reused for the next run's sync.

## State of this PR:

Ready for review and merger. :rocket:  :checkered_flag: :1st_place_medal: 

 - [x] documentation
 - [x] testing testing testing
 - [x] make test suite work
 - [x] sync from syncing peers
 - [x] tolerate reorgs during syncing
 - [x] find luca before syncing

## `syncing.md`

The following text is copied from `syncing.md` for the reviewer's convenience.

### The Sync Loop

The sync loop (`sync_loop.rs`) is an event loop that manages the sync. It has two responsibilities:
 1. It must download all the blocks.
 2. It must feed them, in the right order, to the main loop.

When sync mode is entered, a new sync loop is spun up. The `MutableMainLoopState` loop holds an optional
`SyncLoopHandle`, which comes with useful wrappers for communication. The node is in sync mode if this option is set.
Likewise, it is in sync mode if the sync anchor (which lives on `GlobalState::net`) is set. The two options are always
set and unset in tandem, and one is accessible to peer loops whereas the other is accessible to the main loop.

The sync loop communicates with the outside world by using the main loop as a proxy. So the main loop is responsible
for forwarding messages from the sync loop to peers if necessary, and from peers to the sync loop (if one is active).

The sync loop stores blocks in a temporary directory. By default, whenever a new sync loop starts it will attempt to
resume the sync with all blocks found in its temporary directory. To override this behavior, and force new sync loops
to start from scratch, run `neptune-core` with the argument `--no-resume-sync`.

The sync loop keeps track of which blocks have been downloaded and which ones not yet through a data structure called
the `SynchronizationBitMask` (and sometimes `coverage` for short). It is a bit mask along with two bounds. All bits
below the lower bound are implicitly 1, and all bits above the upper bound are implicitly 0. In between the bits can be
0 or 1, depending on whether the corresponding block was downloaded. The lower bound coincides with the height of the
highest block processed so far by the state update function of the main loop. The upper bound coincides with the target
block that we are syncing relative to (plus one).

This synchronization bit mask is how the sync loop determines which blocks to request from another node who is also
syncing. Assuming the lower and upper bounds agree, the expression `own_mask & !other_mask` puts zeros in the locations
corresponding to block heights that the peer is capable of serving and which we ourselves have not downloaded yet.

Whenever the fist bit after the lower bound is one, the sync loop spawns a separate task called the tip-successors task
and encapsulated in `SyncLoop::process_successors_of_tip`. This tokio task is responsible for reading the next block
from the temporary directory, validating it, sending it to the main loop, and updating the lower bound accordingly. When
it terminates it asynchronously sends a return code back to the sync loop which processes it accordingly.

Validation of blocks happens relative to an enum `BlockValidator`. In production there is only one variant,
`Production`, which invokes `Block::is_valid`. For tests, `Test` is an alternative that allows testing the sync loop
relative to random (and thus invalid) blocks.

The `SyncProgress` is an object indicating how far the sync is removed from completion relative to very start when the
only block in possession was the genesis block. The `SyncProgress` reports on the downloading step only, and so you
could be "100% synced" and have a large backlog of blocks on disk to update the state with. The reason why the
`SyncProgress` cannot report a progress number relative to the "missing interval" of blocks is because the magnitude of
this missing interval is not known when syncing across a fork. In particular, exact knowledge of that interval implies
knowledge of the latest universal common ancestor (LUCA) block, and it is precisely the sync loop's job to figure out
which block that is. Lastly, it is worth noting that asynchronous messages asking the sync loop to compute the sync
progress or responses to such a request are ranked lowest on the priority ladder and will be eagerly dropped or not sent
to begin with if there is strain on the channels.

### Reorganization While Syncing

If there is a reorganization while syncing, then at some point a block validation will fail. When this happens, the sync
loop terminates and cleans up the temporary directory. At this point the main loop sends a `RequestBlockNotification`
message to all peers. Responses to this request may (and should) trigger a new sync.

### Sync Failure

If the sync loop fails, whether because of a reorganization or another reason, the default behavior is for the temporary
directory to be deleted so that future syncs can start from a clean slate and avoid potentially corrupt data. However,
if the temporary directory could not be deleted, the node logs a `error!` message informing the user to delete them
manually.

### Syncing across a Fork

The lower bound on the synchronization bit mask is initially set to 1, corresponding to (the successor of) the genesis
block. Immediately after spawning the sync loop, the main loop queries the peer that triggered the sync loop about his
block at the height of the current tip. Depending on what the peer returns, there are two possibilities.

 1. The peer returns the same block. In this case the peer is on the same fork but on a higher height. The sync loop can
    be fast-forwarded to set the lower bound to the successor of the tip. All the blocks prior to the tip and including
    the tip have already been downloaded and processed.
 2. The peer returns an unknown block. In this case the peer is on a different fork and LUCA, the latest universal
    common ancestor block, is unknown. In this case we do not fast-forward the sync loop, or at least not immediately.
    The sync loop will ask for uniformly random blocks between 0 and the target height. All incoming blocks are tested:
     - do we know this block already?
     - is it on the canonical chain?
    because if the answer is yes to both then we can (and do) fast-forward the sync loop. If not, the block is treated
    as a regular sync block -- *i.e.*, marked as downloaded and stored in the temporary directory until it can be fed to
    the main loop. The `set_new_tip` function will be triggered on the peer's LUCA-successor once it is downloaded and
    the sync loop starts a tip-successors subtask (which in this very particular context is poorly named because until
    this task is spawned our own tip is still on our own stale branch and the LUCA-successor does not succeed it).


dan-da:
Does this implement headers-first sync?   From the description, I am guessing it does not.

I would suggest it is worth implementing.   Either for this PR, or as a followup.

----------

from LLM:

### 🚀 Speed and Efficiency

Block headers are tiny compared to full blocks, which contain all the transaction data.

1. **Quickly Identify the Heaviest Chain**: By downloading and verifying all block headers first, a new node can quickly determine which chain has the most accumulated Proof-of-Work (PoW), which is the consensus mechanism for finding the "correct" and longest chain. This takes far less bandwidth than downloading all the transaction data for every competing chain.

2. **Parallel Block Download**: Once the full, verified chain of headers is known, the node can request the corresponding full block data from multiple peers simultaneously. In full-block sync, the node often has to download blocks sequentially from a single peer, which is much slower.

### 🛡️ Security and Resilience
The headers-first approach allows for pre-validation and isolation of bad data, making the sync process more secure.

1. **Avoid Downloading Invalid Blocks**: The node can validate the Proof-of-Work (PoW) and the chaining of the headers (each header must correctly reference the previous one) before downloading the much larger, full block data. This prevents a malicious peer from forcing the node to download and store large amounts of data for blocks that are not on the main chain (disk-fill attacks or low-difficulty block spam).

2. **Prevent Orphan Block Storage**: In a full-block sync, a node might receive a block whose parent block has not yet arrived (an "orphan" block). These must be temporarily stored in memory or on disk, which wastes resources. By syncing headers first, the node ensures it only requests blocks that are confirmed to be on the longest known chain and are received in the correct order, avoiding the need to store orphan blocks.

aszepieniec:
Here is a limitation of the current design that I'm not sure how to address so I thought I would put some words to it and invite feedback.

The sync loop is capable of syncing from peers who themselves are syncing. It can be made aware of their synchronization state and query only missing blocks that the peer is capable of serving.

However, in order to enter into sync mode a peer needs to receive a valid sync challenge response. This response can only come from a peer who is fully synced already since it consists of a list of blocks whose heights are random. Therefore, a syncing peer cannot trigger sync mode in another peer. So the situation where two peers are syncing occurs only if they both receive valid sync challenge responses from some other fully synced peer. And they can only share the load if they happen to share a connection; otherwise all blocks will come from fully synced peers anyway.

I think there are a few ways forward.

 1. Modify the sync challenge and response so that it can be relayed. This is dangerous though because there is a (small but nonzero) false success probability. One malicious node can reuse the hard work done for tricking one peer into starting sync mode, to potentially trick many others also.
 2. Replace the sync challenge and response with headers-first sync. Eliminates the false success probability. Lot of work. Also: not succinct: the data transmitted and downloaded for a headers-first sync grows with history. This workload would be made moot by succinctness.
 3. Do nothing. The problem will be solved by succinctness. In the mean time we can tolerate suboptimal bandwidth use.

I gravitate to option 3.

aszepieniec:
Ready for review.

Note to reviewers: probably best to start with `syncing.md`.

neptune-core/src/application/loops/sync_loop/handle.rs
```diff
@@ -0,0 +1,209 @@
...
+use crate::api::export::BlockHeight;
+use crate::application::loops::main_loop::block_validator::BlockValidator;
+use crate::application::loops::sync_loop::channel::MainToSync;
+use crate::application::loops::sync_loop::channel::SyncToMain;
+use crate::application::loops::sync_loop::synchronization_bit_mask::SynchronizationBitMask;
+use crate::application::loops::sync_loop::PeerHandle;
+use crate::application::loops::sync_loop::SyncLoop;
+use crate::protocol::consensus::block::Block;
+
+/// Wraps a [`SyncLoop`] along with channels to an fro.
```
skaunov, line 17: "fro."?

aszepieniec: I misspelled "and". "To and fro" is an expression.

aszepieniec:
> I really like when the first item/comment serves as a (running) summary of the whole thing. I see you was updating it already, 🔥 could you update it again to reflect the current state?

I think there is little to add besides what is already contained in `syncing.md`. But I might as well copy the contents of that file to the description of this PR for the reviewers' convenience and, as a side effect, to eliminate all outdated markers of work-in-progress.

aszepieniec:
> for/in #729 I generalized two like handlers (do I call that right?) you might want to take a look and incorporate this to facilitate the future development
> 
>     * https://github.com/skaunov/neptune-core/blob/libp2p_introduction/neptune-core/src/application/loops/handle_proposal_from_peer.rs
> 
>     * https://github.com/skaunov/neptune-core/blob/libp2p_introduction/neptune-core/src/application/loops/handle_tx_from_peer.rs

Can you motivate this change more thoroughly?

From what I can tell, the functions `handle_proposal_from_peer::the` and handle_tx_from_peer::the` factor out code that used to live in the match statement's case. I understand the value of splitting a very long file into multiple more manageable parts. However, I don't see what is being generalized. Also, I don't see how to apply this pattern to message-handlers related to the sync loop. (Besides simply factoring out case handlers into functions that live in new files, but I'm sure that misses the point of your change.)

skaunov:
> > I really like when the first item/comment serves as a (running) summary of the whole thing. I see you was updating it already, 🔥 could you update it again to reflect the current state?
> 
> I think there is little to add besides what is already contained in `syncing.md`. But I might as well copy the contents of that file to the description of this PR for the reviewers' convenience and, as a side effect, to eliminate all outdated markers of work-in-progress.

just to clarify: I saw "WIP", the thing seemed finished though, and the timing suggested there's something to add / something changed since the last edit
so I asked about this just to clarify; I guess it was only that marker this time --- sorry

Issue #785, opened 2025-11-25
Fast rescan for all onchain-notifications


Sword-Smith:
Following up on #783, I thought it would be good to have a way of rescanning all blocks for incoming transactions.

Wallets that receive a lot of UTXOs, like exchanges and mining pools, need a fast way to rescan all historical blocks for UTXOs for all their addresses. Going through the regular state update `GlobalState::set_new_tip` is too slow.

Such a fast scan can be done if the node manages a lookup table from `block_hash => (flag, receiver_identifier)` for all announcements in each block. This extra lookup table can be implemented as a new database, or as a new table on the existing `block_index_db` on archival state, whichever makes more sense. The extra lookup table should be optional, meaning that it is only generated/managed/maintained whenever the node is started with a flag, e.g. `--announcement-index`.

The rescan function can then be implemented roughly as follows:

```rust
impl GlobalState {
    fn rescan_addresses(&mut self, addresses: Vec<ReceivingAddresses>) {
        
        let addresses: HashMap = addresses.into_iter().map(|addr| ((addr.flag(), addr.receiver_identifier())), addr).collect();
        let tip_height: u64 = self.chain.light_state().kernel.header..height.into();
        let mut incoming_utxos = vec![];
        for i in 0..tip_height {
            let digest = self
                .archival_block_mmr
                .ammr()
                .get_leaf_async(i.into())
                .await;
            
            // get (flag, receiver_id) pairs in block, from the announcement index DB/table
            let announcements: Vec<(BFieldElement, BFieldElement)> = self.announcement_index(digest).await;
            let announcements = announcements.iter().filter(|ann| addresses.contains_key(ann));
            if announcements.is_empty() {
                continue;
            }

            // Find the actual announcement somewhere. Maybe from block on disk?
            let block = self.chain.archival_state().get_block(digest).unwrap().unwrap();
            let keys = announcements.map(|ann| addresses[ann]);
            let incoming_utxos = keys.map(|key| key.scan_for_announced_utxos(&block.kernel.body.transaction_kernel));

            // Add MonitoredUtxo for each incoming utxo / digest pair -- if not already present in wallet. Be sure to filter
            // out duplicates somehow!

        }
    }
}
```
[...]

PR #794, opened 2025-12-01
Perf wallet

### LINKED ISSUE #786: Fast UTXO duplication check in wallet
To ensure that we do not register the same UTXO twice, we currently do some convoluted and slow logic that loops over all monitored UTXOs whenever a new block is processed. I believe this check can be made near instantaneous by adding a new table to `WalletDbTables`:

```rust
#[derive(Debug)]
pub(super) struct WalletDbTables {
...

    // table number: 8
    // Mapping from AOCL leaf index to index into list of `monitored_utxo` for UTXOs
    // managed by this wallet. Value type is list because of potential forks.
    pub(super) aocl_to_mutxo: DbtHashMap<u64, Vec<u64>>,
}
```

This would enable a constant-time lookup from an AOCL leaf index to a monitored UTXO to check if we've already processed this UTXO. This new table would also be required for a fast implementation of #785. 

This suggestion has two complications: 
1. We need an implementation of `DbtHashMap` which we don't currently have. Preferably with an option to loop over all keys, meaning that all keys should be stored in the database.
2. We need to do a wallet database migration from v1 to v2.

Sword-Smith:
Motivation for working on this speedup: nodes managing many UTXOs are reporting unacceptably slow state updates. Closes #786. 

44 % overall speedup on wallet state update. With the unlocked potential of fast rescan as explained in #785. Gets rid of all linear-time searches through all monitored UTXOs by adding two indices for monitored UTXOs: One for the AOCL leaf index and one for the absolute index set. Both of these are known at the time of insertion, so managing these lookup tables is simple enough.

Adds a `DbtMap` implementation so that we can manage hashmap-like data atomically in the database. The `DbtMap` implementation that I went with does not support the deleting of entries since I could not combine that with the capability of looping over all keys. Since our current use cases (the two added indices for monitored UTXOs) don't require the deletion of entries, I find that an acceptable limitation.

Adds a migration case for v1 to v2 for the wallet database.

### DIFF
--- a/neptune-core/src/application/database/storage/storage_schema/mod.rs
+++ b/neptune-core/src/application/database/storage/storage_schema/mod.rs
@@ -660,9 +660,9 @@ mod tests {
         }
         assert_eq!(NUM_INSERTIONS, persisted_map.len().await);
 
-        for (key, value) in hashmap.iter() {
-            assert_eq!(*value, persisted_map.get(&key).await.unwrap());
-            assert!(persisted_map.contains_key(&key).await);
+        for (key, value) in &hashmap {
+            assert_eq!(*value, persisted_map.get(key).await.unwrap());
+            assert!(persisted_map.contains_key(key).await);
         }
         assert_eq!(NUM_INSERTIONS, persisted_map.len().await);
 
@@ -670,9 +670,9 @@ mod tests {
 
         rusty_storage.persist().await;
 
-        for (key, value) in hashmap.iter() {
-            assert_eq!(*value, persisted_map.get(&key).await.unwrap());
-            assert!(persisted_map.contains_key(&key).await);
+        for (key, value) in &hashmap {
+            assert_eq!(*value, persisted_map.get(key).await.unwrap());
+            assert!(persisted_map.contains_key(key).await);
         }
         assert_eq!(NUM_INSERTIONS, persisted_map.len().await);
         assert_eq!(inserted_keys, persisted_map.all_keys().await);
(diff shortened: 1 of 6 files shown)


neptune-core/src/application/database/storage/storage_schema/dbtmap_private.rs
```diff
@@ -0,0 +1,198 @@
...
+use super::SimpleRustyReader;
+use super::WriteOperation;
+use crate::application::database::storage::storage_schema::DbtVec;
+use crate::application::database::storage::storage_vec::traits::StorageVecBase;
+use crate::application::database::storage::storage_vec::Index;
+use crate::application::locks::tokio::AtomicRw;
+
+/// A level-DB backed insertion-only mapping from keys to values.
+///
+/// The reason that remove functionality is not supported is that the
```
dan-da, line 23: I might be missing something, but at first thought it appears to me that DbtVec could efficiently support a delete or remove operation by adding a last_index property.  The delete op would simply delete the key at specified index and reduce the len.  This would leave a missing element, creating a sparse array.  Iteration would iter until last_index, skipping any elements not found.

Though rather than modify DbtVec itself, it may be better to create a new DbtVecSparse or something like that.

It's unclear to me how necessary a delete Op may be for DbtMap.   For this PR it is not necessary, but is it intended for general purpose usage, or just for this particular usage?

Sword-Smith: The `DbtMap` is implemented for our concrete needs. I have not attempted to make it more generic than that. I have, however, attempted to follow principles that you have already established for database qualities, such as thread-safety although that is not strictly necessary for our use since it's all handled by the global state's rw-lock.

PR #798, opened 2025-12-03
Sync performance fixes

aszepieniec:
There might be something useful in these changes. Credit to AllFather team.

### DIFF
--- a/neptune-core/benchmark/benches/wallet_state.rs
+++ b/neptune-core/benchmark/benches/wallet_state.rs
@@ -178,10 +178,7 @@ mod maintain_membership_proofs {
 
                     if update_msmps {
                         global_state
-                            .restore_monitored_utxos_from_archival_mutator_set(
-                                Some(recovery_data),
-                                true,
-                            )
+                            .restore_monitored_utxos_from_archival_mutator_set(None)
                             .await
                     }
                 });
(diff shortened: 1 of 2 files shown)


neptune-core/src/state/wallet/wallet_state.rs
```diff
@@ -709,6 +709,17 @@ impl WalletState {
...
+            .inputs
+            .iter()
+            .map(|rr| rr.absolute_indices)
+            .collect();
+
+        // Early return if no inputs in block - nothing could be spent
+        if confirmed_absolute_index_sets.is_empty() {
+            return HashMap::default();
+        }
+
```
Sword-Smith, line 722: Similar improvement already exists, introduced with the migration to wallet DB v2. Current code: 

```rust
    async fn scan_for_spent_utxos(
        &self,
        transaction_kernel: &TransactionKernel,
    ) -> HashMap<AbsoluteIndexSet, (Utxo, u64)> {
        let mut spent_own_utxos = HashMap::default();

        for index_set in transaction_kernel
            .inputs
            .iter()
            .map(|rr| &rr.absolute_indices)
        {
            if let Some((mutxo, mutxo_list_index)) =
                self.wallet_db.monitored_utxo_by_index_set(index_set).await
            {
                spent_own_utxos.insert(index_set.to_owned(), (mutxo.utxo, mutxo_list_index));
            }
        }

        spent_own_utxos
    }
```

PR #799, opened 2025-12-03
~~proof of reserves,~~ proof of a transfer

skaunov:
# a transfer
current list of review points is tracked & processed in [the following comment](https://github.com/Neptune-Crypto/neptune-core/pull/799#issuecomment-4114747401)
# reserves
here's a good draft of the Triton code
the discussion is below, will be get back to it after the transfer part


neptune-core/src/application/util_proof/reserves/mod.rs
```diff
@@ -0,0 +1,384 @@
+/// Outputting RR not only allows to check that the UTXO wasn't spent immediately after proving, but also allows to check that it isn't used in unrelated proofs (a problem which
+/// occurs on its own and basically leads to the same solution).
+///
+/// Proving per an UTXO makes the job easier for the prover because 1) parrallelization and 2) the smaller proofs. This increase a verifier work which shouldn't be a problem due
+/// to parallelization.
+///
+/// # running
+/// Correct me if I'm wrong: for proving a new Neptune CLI command should be added.
```
aszepieniec: > /// Correct me if I'm wrong: for proving a new Neptune CLI command should be added.

Yes that makes sense.

neptune-core/src/application/util_proof/reserves/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+/// # checks out of Triton
+/// - AOCL must be from the block of interest
```
aszepieniec: > /// - AOCL must be from the block of interest

More precisely: the `AdditionRecord` must live in the AOCL of the block of interest.

neptune-core/src/application/util_proof/reserves/mod.rs
```diff
@@ -0,0 +1,384 @@
+/// Outputting RR not only allows to check that the UTXO wasn't spent immediately after proving, but also allows to check that it isn't used in unrelated proofs (a problem which
+/// occurs on its own and basically leads to the same solution).
+///
+/// Proving per an UTXO makes the job easier for the prover because 1) parrallelization and 2) the smaller proofs. This increase a verifier work which shouldn't be a problem due
+/// to parallelization.
+///
+/// # running
+/// Correct me if I'm wrong: for proving a new Neptune CLI command should be added.
+/// # checks out of Triton
```
aszepieniec: Checks outside of Triton VM, meaning that the verifier verifies the proof and some extra data. Doesn't this extra data not leak privacy? For instance, if you need to include the AdditionRecords, then that leaks exactly when the reserves were obtained.

neptune-core/src/application/util_proof/reserves/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+    input: (Digest, Timestamp),
+    /// the record to check if the reserve is still unspent;
+    /// AOCL digest;
+    /// `lock_postimage` of the address;
+    /// the 'reserve'
+    /// the timelocked 'reserve'
+    output: (
+        crate::util_types::mutator_set::removal_record::RemovalRecord,
+        Digest,
+        Digest,
```
aszepieniec: What do these digests represent?

neptune-core/src/application/util_proof/reserves/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+        NativeCurrencyAmount,
+        NativeCurrencyAmount,
+    ),
+}
+impl crate::protocol::proof_abstractions::tasm::program::ConsensusProgram for PublicData {
```
aszepieniec: Consider naming the program something more specific than `PublicData`. Just about every proof type used in NeptuneCash has some kind of public data.

neptune-core/src/application/util_proof/reserves/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+        let rustfield_coins = rustfield!(Utxo::coins);
+        // let rustfield_typescripthash = rustfield!(Coin::type_script_hash);
+        // let rustfield_state = rustfield!(Coin::state);
+        let audit_preloaded_data = library.import(Box::new(
+            tasm_lib::structure::verify_nd_si_integrity::VerifyNdSiIntegrity::<
+                // MemoryPreload
+                crate::util_types::mutator_set::ms_membership_proof::MsMembershipProof,
+            >::default(),
+        )); // TODO
+        let payload = triton_asm! {
```
aszepieniec: What is the stack signature of this blob of tasm code?

neptune-core/src/application/util_proof/reserves/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+
+            call {lib_hash_absolute_indices}
+            pop 1
+            // *aocl [computed_bloom_indices]
+            write_io 5
+            // *aocl
+
+            /* finish of the things from the RR integrity file
+            ____________________________ */
+
```
aszepieniec: Please add a stack signature.

neptune-core/src/application/util_proof/reserves/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+            swap 2
+            // *aocl *aocl_peaks [num_leafs] [aocl_leaf_index] [canonical_commitment] *auth_path
+
+            call {lib_mmr_verify}
+            assert
+            // *aocl
+
+            /* pasted '8.' from <removal_records_integrity.rs>
+            ============== */
+            /* 8. */
```
aszepieniec: But what does "8." do? Please summarize.

If significant chunks of tasm code is being copied, consider factoring it out into a separate blob or even `BasicSnippet`. In this case, an appropriate name for the blob or snippet will obviate the need for explanatory comment.

Sword-Smith: It's not certain that subsections of `RemovalRecordsIntegrity` can be factored out into a `BasicSnippet` without changing the function signature of `RemovalRecordsIntegrity`, which would be a hard-fork (which we certainly don't want!). 

*Maybe* something can be factored out into a block of code but even that is not certain, as the relevant chunk of code in `RemovalRecordsIntegrity` might depend on a very specific stack setup, or statically-allocated memory addresses.

neptune-core/src/application/util_proof/reserves/mod.rs
skaunov: that's proven and the AOCL digest is output, so a verifier must find the digest in a canonical block

neptune-core/src/application/util_proof/reserves/mod.rs
skaunov: That's why it keeps these private. X) And the prover can prove for the latest block AOCL which is relevant to the task in hand. 

In a sense a good proof of the reserves requires to leak some privacy. Like it makes sense to leak the lock post-image. So the question is where to draw the line. I start the current version with the comment on the choice I present here.

neptune-core/src/application/util_proof/reserves/mod.rs
skaunov: it's 'AOCL digest' and '`lock_postimage` of the address'

neptune-core/src/application/util_proof/reserves/mod.rs
skaunov: Oh, I had this question! How do you denote the empty stack? Jan Ferdinand told me recently the stack is never empty, but I feel like that doesn't impact a _stack signature._

aszepieniec: There are always at least 16 elements on the stack. Any less -- crash!

If the stack contains nothing interesting, you can denote it with `// _`.

neptune-core/src/application/rpc/server.rs
```diff
@@ -1989,6 +1998,13 @@ pub trait RPC {
     /// # Ok(())
     /// # }
     async fn shutdown(token: auth::Token) -> RpcResult<bool>;
+
```
aszepieniec, line 2133: Please add a docstring. The docstring should:
 - Explain how the arguments relate to the claim.
 - List the conditions for success.
 - Indicate how to verify the produced claim and proof.
 - Contain a minimal working example.

aszepieniec: Also, I think the word "transfer" is needlessly generic. Transfer could denote transfer of responsibilities, or transfer of keys. When it's a transfer of funds there is a more specific term: *payment*.

neptune-core/src/application/rpc/server.rs
```diff
@@ -4163,9 +4179,96 @@ impl RPC for NeptuneRPCServer {
             .map(|tx| &tx.kernel)
             .cloned())
     }
+
+    // Crashes if `tx_ix` out of its bound. https://github.com/Neptune-Crypto/neptune-core/issues/816
```
aszepieniec: This crash condition should live on the docstring above, where the function is declared and the trait defined.

neptune-core/src/application/rpc/server.rs
```diff
@@ -4163,9 +4179,96 @@ impl RPC for NeptuneRPCServer {
...
+    async fn prove_transfer(
+        self,
+        context: ::tarpc::context::Context,
+        token: auth::Token,
+        tx_ix: u64,
+        utxo_ix: usize,
+        block: Digest,
+    ) -> RpcResult<(Claim, NeptuneProof)> {
```
aszepieniec, line 4654: Start the RPC call by logging the duration of the call and authenticating the user:

```
        log_slow_scope!(fn_name!());
        token.auth(&self.valid_tokens)?;
```

neptune-core/src/application/rpc/server.rs
```diff
@@ -4163,9 +4179,96 @@ impl RPC for NeptuneRPCServer {
...
+                                // @skaunov guess we don't want to interfere with anything.
+                                api::export::TritonVmJobPriority::Lowest,
+                            )
+                            .build(),
+                    )
+                    .await?,
+                ))
+            } else {
+                Err(todo![])
+            }
```
aszepieniec: A let-else pattern instead of an if-let-else pattern avoids needless indentation and puts the failure actions close to the failure condition.

neptune-core/src/application/rpc/server.rs
```diff
@@ -4163,9 +4179,96 @@ impl RPC for NeptuneRPCServer {
...
+        } else {
+            Err(todo!())
+        }
```
aszepieniec: In fact, you can use the let-else pattern twice here to save 8 characters of indentation.

neptune-core/src/application/rpc/server.rs
```diff
@@ -4163,9 +4179,96 @@ impl RPC for NeptuneRPCServer {
...
+            if let Some(tx_output) = self
+                .state
+                .lock_async(|s| s.wallet_state.wallet_db.sent_transactions().get(tx_ix))
+                .await
+                .tx_outputs
+                .get(utxo_ix)
+            {
+                // let utxo = tx.tx_outputs.utxos_iter().into_iter().skip(utxo_ix).next()?;//.map_err(todo![])?;
+                let utxo = tx_output.utxo();
+                let aocl = block.body().block_mmr_accumulator;
```
aszepieniec: Good news! Thanks to the recent changes to the wallet database (now on master), the field `aocl_leaf_index` now lives on `MonitoredUtxo`. I'm not sure about the exact sequence of dabase lookups to get the right `MonitoredUtxo` but I am sure it's a lot easier with this change than it was before.

neptune-core/src/application/rpc/server.rs
```diff
@@ -4163,9 +4179,96 @@ impl RPC for NeptuneRPCServer {
...
+                let sender_randomness = tx_output.sender_randomness();
+                let sent = sent::new(
+                    sent::claim_outputs(
```
aszepieniec: I cannot parse the `sent::` prefix. Is there a module missing?

aszepieniec: Okay I found the module under `util_proofs/sent/mod.rs`. I wonder why my IDE does not link functions though?

neptune-core/src/application/util_proof/sent/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+    memory::{encode_to_memory, FIRST_NON_DETERMINISTICALLY_INITIALIZED_MEMORY_ADDRESS},
+    prelude::{Digest, Library, TasmObject, TasmStruct},
+    triton_vm::{
+        isa::triton_asm,
+        prelude::{BFieldCodec, LabelledInstruction, Program},
+        proof::Claim,
+        vm::{NonDeterminism, PublicInput},
+    },
+    twenty_first::util_types::mmr::mmr_accumulator::MmrAccumulator,
+};
```
aszepieniec: We use "item" granularity for imports to avoid headaches when solving git merge conflicts.

neptune-core/src/application/util_proof/sent/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+
+use crate::{
+    api::export::{NativeCurrencyAmount, Utxo},
+    protocol::proof_abstractions::{tasm::program::ConsensusProgram, SecretWitness},
+    util_types::mutator_set::ms_membership_proof::MsMembershipProof,
+};
+
+#[derive(TasmObject, tasm_lib::triton_vm::prelude::BFieldCodec, Debug)]
+struct WitnessMemory {
```
aszepieniec: Please avoid overly generic struct names. Almost all consensus programs have some witness living in memory. Name the struct after what sets this one apart from those other structs.

Also, the suffix "Memory" suggests that the struct that lives in VM memory is different from some other witness struct, which contains more information but information that should not go into VM memory. If that's the case, where do I find that more complete struct? If there is no distinction, drop the suffix "Memory".

neptune-core/src/application/util_proof/sent/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+}
+
+pub fn claim_inputs(
+    c: Claim,
+    receiver_digest: Digest,
+    release_date: crate::api::export::Timestamp,
+) -> Claim {
+    c.with_input(
+        [
+            [release_date.0].as_slice(),
```
aszepieniec: What happens if there is no time-lock?

neptune-core/src/application/util_proof/sent/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+        [
+            lock_postimage.values().into(),
+            sender_randomness_digest.values().to_vec(),
+            aocl_digest.values().into(),
+            amount.encode(),
+            // amount_timelocked.encode(),
+        ]
+        .concat(),
+    )
+}
```
aszepieniec, line 68: What's the motivation for splitting these data items across the input and output fields of the claim?

neptune-core/src/util_types/sent/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+            sender_randomness_digest.values().to_vec(),
+            aocl_digest.values().into(),
+            amount.encode(),
+            // amount_timelocked.encode(),
+        ]
+        .concat(),
+    )
+}
+
+fn library_and_code() -> (Library, Vec<LabelledInstruction>) {
```
aszepieniec, line 70: The standard in the rest of the code base is to implement this function (and similar ones, like `hash`) as implementations of trait `ConsensusProgram` on empty structs. So you can do, for instance, `ProofOfPayment::hash()`. Please follow that pattern.

We should consider relaxing the name `ConsensusProgram` since with this addition it also covers programs not related to consensus (that still implement the same trait). I think `NeptuneProgram` might be a good choice.

neptune-core/src/util_types/sent/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+    let rustfield_utxodigest = rustfield!(WitnessMemory::utxo_digest);
+    let rustfield_aoclleafindex = rustfield!(MsMembershipProof::aocl_leaf_index);
+
+    let payload = triton_asm! {
+        /* pasted from <../reserves>
+        ============== */
+        // _
+        push {FIRST_NON_DETERMINISTICALLY_INITIALIZED_MEMORY_ADDRESS}
+        {&rustfield_utxo}
+        {&Utxo::get_field("lock_script_hash")}
```
aszepieniec, line 109: Where do I find the implementation of `get_field`?

aszepieniec: Oh, I get it! It's the one that is declared by `TasmObject`. Why not use the `field` macro instead? That would be consistent with what's used in the rest of the code base.

neptune-core/src/application/util_proof/sent/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+    let payload = triton_asm! {
+        /* pasted from <../reserves>
+        ============== */
+        // _
+        push {FIRST_NON_DETERMINISTICALLY_INITIALIZED_MEMORY_ADDRESS}
+        {&rustfield_utxo}
+        {&Utxo::get_field("lock_script_hash")}
+        read_mem 5
+        pop 1
+        // [lock_script_digest]
```
aszepieniec: This stack signature is incorrect because there are still 16 elements below the lock script digest. Correct is this:

```
// _ [lock_script_digest]
```

aszepieniec: Please include the underscore in the stack signatures below where missing.

neptune-core/src/application/util_proof/sent/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+        write_io 5
+        // _
+
+        push {FIRST_NON_DETERMINISTICALLY_INITIALIZED_MEMORY_ADDRESS}
+        // *w
+        {&WitnessMemory::get_field("aocl")}
+        // *aocl
+
+        // prepare the value for '7.'
```
aszepieniec: Please make the files self-contained. Don't make the reader open another file to figure out what "7." means.

neptune-core/src/application/util_proof/sent/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+        finish of the code pasted from <reserves.rs> */
+
+        pop 1 write_io 4
+        // num_coins num_coins *eof amount timelocked_amount
+    };
+
+    let imports = library.all_imports();
+    let code = triton_asm!(
+        {&payload}
+        {&main}
```
aszepieniec: Does control start in `main` or in `payload`? If the second, what's the purpose of `main`?

neptune-core/src/application/util_proof/sent/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+        hash
+        write_io 5
+
+        push {FIRST_NON_DETERMINISTICALLY_INITIALIZED_MEMORY_ADDRESS}
+        // _ *w
+        {&rustfield_utxo}
+        // *aocl *aocl_peaks [receiver_digest] [sender_randomness] *utxo
+    };
+
+    let main = triton_asm! {
```
aszepieniec: Since control does not start with `main` (right?), the stack is not empty/trivial. What does it look like?

neptune-core/src/util_types/sent/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+        push 0
+        push 0
+        push 0
+        dup 9
+        dup 9
+        dup 9
+        dup 9
+        dup 9
+        hash
+        write_io 5
```
aszepieniec, line 148: Please explain in words what is being written to standard output here.

neptune-core/src/application/util_proof/sent/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+        swap 1
+        // _ *utxo+1 utxo_size
+        call {lib_hash_varlen}
+        hint utxo_hash = stack[0..5]
+        // *aocl *aocl_peaks [receiver_digest] [sender_randomness] [utxo_hash]
+
+        push {FIRST_NON_DETERMINISTICALLY_INITIALIZED_MEMORY_ADDRESS}
+        {&rustfield_utxodigest}
+        read_mem 5
+        assert_vector
```
aszepieniec: Isn't the memory pointer still on top of the stack? If so, that will interfere with the `assert_vector`.

neptune-core/src/application/util_proof/sent/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+        // *aocl *aocl_peaks [receiver_digest] [sender_randomness] [utxo_hash]
+
+        call {lib_ms_commit}
+        // *aocl *aocl_peaks [canonical_commitment]
+
+        /* 7. */
+        dup 6 {&MmrAccumulator::get_field("leaf_count")}
+        // *aocl *aocl_peaks [canonical_commitment] *num_leafs
+
+        addi 1 read_mem {u64_stack_size} pop 1
```
aszepieniec: The 1 in `addi 1` comes from one less than `u64_stack_size`, right? If so, `addi {u64_stack_size - 1}` is clearer.

neptune-core/src/application/util_proof/sent/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+        // *aocl *aocl_peaks [canonical_commitment] [num_leafs]
+
+        push {FIRST_NON_DETERMINISTICALLY_INITIALIZED_MEMORY_ADDRESS}
+        {&rustfield_aoclleafindex} // omits `{&rustfield_membershipproof}` from <../reserves>
+        read_mem {u64_stack_size}
```
aszepieniec: Don't you need to shift the pointer to the end of the data object you want to read first? I think a `addi {u64_stack_size - 1}` might be missing.

neptune-core/src/application/util_proof/sent/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+        pop 1
+        // *aocl *aocl_peaks [canonical_commitment] [num_leafs] [aocl_leaf_index]
+
+        push {FIRST_NON_DETERMINISTICALLY_INITIALIZED_MEMORY_ADDRESS}
+        {&WitnessMemory::get_field("auth_path_aocl")}
+        {&WitnessMemory::get_field("authentication_path")}
+        // *aocl *aocl_peaks [canonical_commitment] [num_leafs] [aocl_leaf_index] *auth_path
+        swap 2
```
aszepieniec: This is not how `swap 2` works.

The stack `_  [num_leafs] [aocl_leaf_index] *auth_path` is short-hand for `_ num_leafs_hi num_leafs_lo aocl_leaf_index_hi aocl_leaf_index_lo *auth_path`.

What you want is `place 4`.

neptune-core/src/application/util_proof/sent/mod.rs
```diff
@@ -0,0 +1,329 @@
...
+}
+
+pub fn hash() -> Digest {
+    static DIGEST: OnceLock<Digest> = OnceLock::new();
+
+    *DIGEST.get_or_init(|| Program::new(&library_and_code().1).hash())
+}
+
+#[derive(Debug)]
+pub struct The(WitnessMemory, Claim);
```
aszepieniec: Please avoid overly generic struct names.

neptune-core/src/application/rpc/server.rs
aszepieniec, line 4654: I'm thinking that this function implementation should be a simple wrapper around another function that lives in the API module. Then that other function can be called also by the JSON-RPC server, as well as tested independently of any RPC servers.

neptune-core/src/application/util_proof/mod.rs
```diff
@@ -0,0 +1,2 @@
+pub mod reserves;
+pub mod sent;
```
aszepieniec: I would suggest that "proof of payment" or even just "payment" is a better description of what this module is about.

neptune-core/src/application/util_proof/reserves/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+    /// AOCL digest;
+    /// the 'reserve'
+    /// the timelocked 'reserve'
+    output: (
+        Digest,
+        crate::util_types::mutator_set::removal_record::RemovalRecord,
+        Digest,
+        NativeCurrencyAmount,
+        NativeCurrencyAmount,
+    ),
```
aszepieniec: These types are too complex. Please separate into many fields and implement convenience functions `input()` and `output()`.

neptune-core/src/application/util_proof/sent/mod.rs
skaunov: Ah, this is an atavism I just copied from another program to get started and was taking around not sure to remove. I'll join these into a single code block when the other suggestions on these are dealt.

neptune-core/src/util_types/sent/mod.rs
skaunov: That macro doesn't work inside a Triton macro. I see the couple advantages of `get_field` like it has clearer name and I don't need to go check some variable and its level of indirection. I mean I feel this line communicates what's going on so well and where it's needed. IIRC I still used the variables when the Rust fields were re-used mainly just to note that those are used in few places of the program.

Also I feel like @jan-ferdinand prefers this kind of syntax too, and I'm not sure if his reasons are different (and even if the feeling is correct :face_with_spiral_eyes:).

neptune-core/src/application/util_proof/sent/mod.rs
skaunov: Now I wrote the following reply it struct me that the question could be about the proving logic, not the code structuring... 😕 Could you clarify this, pls.
> It's two edged. To me this communicates `Claim` parts before serialization much clearer than any docs just by leveraging the types. And on the other side (I should mention this in docs when I get to it!) it mirrors `Claim` methods so, if someone have an experience with it, it would be easy to pick this up.
> 
> Does this question in a way reduces to why `Claim` have these methods splitted.

neptune-core/src/util_types/sent/mod.rs
skaunov: This one is private anyway and it is used in `impl ConsensusProgram` much like you said. Like it's not a problem to add the colors to the module with the empty/trivial structures, just for me it was quite confusing to disentangle this jumping across the docs just to end up on the stub structures. X) To me having all the implementations on the single structure in the module makes it so much more faster to comprehend. I mean would you consider and give this feedback a thought before I convert the code to your likings / the approach?

Yeah, the trait name gave me the few head-scatches before I figured out it actually fits. X) I would like `TritonProgram` more as it needs a Triton VM and probably ends in a Triton job queue. Also it instantly communicates to the reader that here he will need Triton knowledge. Plus it will be beneficial for search queries like 'triton' when someone would like to peek the parts of the code base more related to Triton.

neptune-core/src/application/util_proof/sent/mod.rs
skaunov: ~~Maybe I don't understand how this part works. I thought the proof is w.r.t. to this release date, meaning that after three years since this date the amount you can be sure in the `amount` from this `Claim` `output`, so when time-locking is not relevant/wanted for the proving use-case a verifier can just check that the date in `Claim` is older than three years from the moment he is interested in, making a time-lock irrelevant.~~

I remembered we had a small discussion on that if I'm not confusing this. It's not relevant at all now — just ignored. As a future development sometime it could be added also that time-locked amount.

neptune-core/src/application/util_proof/sent/mod.rs
skaunov: I'm glad to put any name you see fit. Though the structure is intentionally private so I thought it should not be a problem ever.

For the context: I intend the usage of the module here, so all the names are 'up to the module', you already seen [how it looks like](https://github.com/skaunov/neptune-core/blob/4db6468186d9af8114d5185482c01e0682a8c43a/neptune-core/src/application/rpc/server.rs#L4214) - no way to confuse and it reads well I hope. No problem to improve if it's not that good.

neptune-core/src/application/rpc/server.rs
skaunov: I don't know how to avoid `return` with it, so I use it only for procedural/effects things, as `return` takes much more effort/attention to reason about the types compared to identation and code folding. If that's possible, it would be magnificient if you could show/teach me how to employ `let _ else` without `return`.

And you're absolutely right on everything else: just added `?` as a better, more idiomatic, and types-friendly solution to that!

neptune-core/src/application/util_proof/sent/mod.rs
skaunov: Oh, I finally got the question! Isn't it just... more natural? Basically the inputs are the data required for proving, and the outputs is the data which is the result of proving (basically partial disclosure of the witness). Looking from another angle inputs are those which require asserting/checking inside Triton, and the outputs are ok to check/assert outside of Triton (so we even save few lines/opcodes).

neptune-core/src/application/rpc/server.rs
skaunov: While we're here, could you explain why `Claim` is a structure and not a trait? Currently to me a trait would make more sense as each program has its own claim structure.

skaunov:
> Make sure to add tests for the programs. Both positive tests and negative tests -- the latter ones trigger specific assert codes.

For tests it seems I'd need a wallet with an outgoing tx. I glanced over the tests which `wallet` has but all I found were with the empty blocks and some composer txs. I'd appreciate help with finding a test/helper I could base upon; just to not find one **after** I make one from scratch.

neptune-core/src/application/util_proof/sent/mod.rs
skaunov: I mostly covered this in the other comments. 1) Just give it a better name. 2) In the example/RPC it doesn't even come up explicitly. I don't remember if I was already saying this: I was thinking to promote it the module up, but then the data and the logic in the different files which is sad. And here I just want to communicate that it's the only structure of the module which is not really important per se, like "focus on the other items in the module."

skaunov:
> I don't think this is correct. The size-indicator _does_ take into account fields with a statically known size. Are you sure you are not looking at the size-indicator for `Utxo::coins`?

I discovered it with <https://github.com/skaunov/neptune-core/blob/16038b1e79469d1c5180a781af328ee3d0145ea4/neptune-core/src/application/util_proof/sent/tests.rs#L135>: length is obviously 18 but the first element which is SI is 13.

skaunov:
> > I'm out of ideas how to further fix this, so your advice would be very helpful!
> 
> I'm not sure what the problem is; can you describe in more elaborate terms please?

Ah, I forgot to name the elephant in the room! :rofl: `Digest` I get in Triton (via the varlen hash `BasicSnippet`) is different from the one I get from `Tip5::hash` of same `Utxo`.

aszepieniec:
Size-indicator value of 13 suggests to me something like this layout.

 - object of size 13 (field `coins`)
   - length-indicator value 1 (i.e., list of length 1 `Vec` of `Coin`)
     - size indicator value 6 (dynamically-sized field `state`) 
       - length-indicator value of 4 (list of length 4, `state`)
         - 4 `BFieldElement`s encoding the `NativeCurrencyAmount`
     - 1 `Digest` (5 `BFieldElement`s for the statically-sized `type_script_hash`)

So I do think the size indicator you're looking at is the size indicator of the field `coins`. Indeed, the other field is `lock_script_hash` which does not come with a size indicator because it has a statically known length, and it would come after in the `BFieldCodec` layout since it is listed first in the struct definition.

aszepieniec:
The document `neptune-core/src/assertion_error_ids.md` tracks error codes -- or rather, it tracks families of error codes and points to the file where the individual members are defined.

Feel free to define your own error codes and add to this file.

skaunov:
Thank you! An unexpected format tbh. I can't stop thinking why it's not a Rust file which would be included in the docs.

neptune-core/src/application/util_proof/sent/mod.rs
skaunov: I'll get back to this at the very end when grooming before merging since that probably would slip into again until then.

skaunov:
> > I'm afraid I cannot make sense of how the proof-of-reserves is supposed to work. Could you write a draft RPC endpoint? That ought to be enough that ought to be enough to reverse engineer my way through.
> 
> Meanwhile I'll `todo!` the quoted part on proof of reserves; they were basically paused with a transfer in focus, and I'd prefer to finish that before actually switch; but it's a good moment/opportunity to demonstrate my path of thought on reserves while it takes the time for the new review points.

I'm not sure if <https://github.com/skaunov/neptune-core/tree/reserves_sketch> is what you meant -- pls guide if not. But it's to demonstrate the idea: fast verification and fast proving thanks to a proof per an UTXO. The down thing is it discloses the removal record in a claim, but I still argue it's not really an issue since the owner always can choose to restore his funds privacy just by sending it himself. Then the reserves will be not proved anymore -- a verifier would see the removal record on the chain.

For huge numbers of the UTXO this design is actually better then a compression via succinctness as a prover only needs to make the new proofs/arguments only for the spent UTXO and they're so much easier than a proof over few UTXO. And doesn't need to redo the proof(s) for each new block. It only worse for putting such a proof on-chain. But 1) that would be a different story, 2) can be worked around via putting the reserves into a single UTXO.

neptune-core/src/api/wallet/wallet_impl.rs
```diff
@@ -154,30 +158,64 @@ impl<'a> Wallet<'a> {
...
+    /// want to prove from the wallet.
+    ///
+    /// The `block` is needed to prove the UTXO was mined. It's determined by its [`Digest`], the relevant data
+    /// is taken from this node DB. During verification the same data will be pulled from the same block.
+    ///
+    /// A verifier should use [`super::super::super::protocol::proof_abstractions::verifier::verify`] exposed on his API/RPC, or use the [TUI](https://github.com/TritonVM/triton-tui).
+    ///
+    /// # Panics
+    /// The implementation detail is when `tx_ix` is out of its bound it crashes the node until
+    /// <https://github.com/Neptune-Crypto/neptune-core/issues/816> is done.
```
aszepieniec: I appreciate this docstring but it's not perfectly clear to me yet. Let me try to rephrase and then you can correct where I made the wrong guess.

```
Produce a proof of payment.

The result is a [`NeptuneProof`] that certifies that a given transaction output transfers the given
amount of [`NativeCurrency`] to the identified receiving address. The amount, the addition record,
the sender randomness, and the `receiver_digest` and `lock_postimage` (which identify the
receiving address) are disclosed by the claim. Besides verifying this proof against the claim, the
verifier must also test membership of the given addition record in the AOCL.

It is a zero-knowledge proof of knowledge revolving around a [`Utxo`] (not disclosed) that hashes
to the given addition record using the sender randomness  and receiver digest. The UTXO in
question has the given `lock_script_hash` and k > 1 [`Coin`]s of type script hash [`NativeCurrency`]
whose states are valid positive amounts and sum to the given amount.
```

Also:

It is not clear to me from this description whether the UTXO is disclosed or not. It strikes me that there might be a problem if the UTXO is not disclosed: I produce a proof of payment for a UTXO that contains, besides the `NativeCurrency` coin of the right amount, some other coin. The recipient has no way of spending this UTXO. One way to get around this obstacle would be to prove instead that the UTXO has *only one coin, which is the `NativeCurrency` coin with the given value*.

I don't get this sentence:

 > The UTXO is binded by hashing `sender_randomness` which serves for identification between the similar transfers.

If I read correctly, this function does 3 things that I probably would have written as separate functions:

 1. Gather the data that the prover needs.
 2. Produce the proof.
 3. Produce the claim.

neptune-core/src/api/wallet/wallet_impl.rs
```diff
@@ -197,10 +235,97 @@ mod worker {
...
     pub async fn spendable_inputs(gs: &GlobalState, timestamp: Timestamp) -> TxInputList {
-        // sadly we have to collect here because we can't hold ref after lock guard is dropped.
+        // Sadly we have to collect here because we can't hold ref after lock guard is dropped.
         gs.wallet_spendable_inputs(timestamp)
             .await
             .into_iter()
             .into()
     }
+
+    pub async fn prove_transfer(
```
aszepieniec: If it is `pub`, please add a docstring, even if it is only referring to the other `prove_transfer`.

neptune-core/src/api/wallet/wallet_impl.rs
aszepieniec: Also, please add a doctest example how to use prove in conjunction with verify.

neptune-core/src/api/wallet/wallet_impl.rs
```diff
@@ -197,10 +235,97 @@ mod worker {
...
+            .get_block(block)
+            .await
+            .map_err(|e| WalletError::Failed(e.to_string()))?
+            .ok_or_else(|| WalletError::Failed("`block` not found".to_string()))?;
+
+        let tx_sent =
+            &crate::application::database::storage::storage_vec::traits::StorageVecBase::get(
+                gs.wallet_state.wallet_db.sent_transactions(),
+                tx_ix,
+            )
```
aszepieniec: (Why) doesn't `gs.wallet_state.wallet_db.sent_transactions().get(tx_ix)` work?

neptune-core/src/api/wallet/wallet_impl.rs
```diff
@@ -197,10 +235,97 @@ mod worker {
...
+                // aocl_digest,
+                utxo.lock_script_hash(),
+                tx_output.native_currency_amount(),
+            ),
+            block_aocl,
+            sender_randomness,
+            aocl_leaf_ix,
+            utxo,
+            aocl_membership_proof,
+        );
```
aszepieniec: This code block gives me no clue about the type of the variable `sent`. Furthermore, since "sent" is a very generic word, I have no idea what it does or is supposed to do.

aszepieniec: Reading ahead slightly, I guess that the type of `sent` is a struct that implements `ConsensusProgram`. Is that right? If so, why not call `StructName::new` explicitly?

neptune-core/src/application/rpc/server.rs
aszepieniec: A `Claim` holds data. It consists of a program digest, a program input, and a program output, and maybe some other fields that I can't think of off the top of my head. I just don't see how a trait is a better fit.

neptune-core/src/application/util_proof/sent/mod.rs
aszepieniec: `ProofOfPaymentWitness`

neptune-core/src/util_types/sent/mod.rs
aszepieniec: I think `TritonProgram` is a great name! It fits much better than `ConsensusProgram`. Let's use `TritonProgram` going forward and feel free to search-and-replace.

aszepieniec: > I mean would you consider and give this feedback a thought before I convert the code to your likings / the approach?

That's a fair request but I'm afraid I don't understand the feedback. Please elaborate (if it is still relevant).

neptune-core/src/util_types/sent/mod.rs
aszepieniec: > I mean I feel this line communicates what's going on so well and where it's needed.

Agreed. I stumbled over the novelty, but after thinking about it I think this improves readability.

neptune-core/src/application/util_proof/sent/mod.rs
aszepieniec: `ProofOfPayment`

neptune-core/src/util_types/sent/spec.rs
```diff
@@ -0,0 +1,106 @@
...
+        },
+    },
+};
+
+impl crate::protocol::proof_abstractions::tasm::program::tests::ConsensusProgramSpecification
+    for super::The
+{
+    fn source(&self) {
+        // get in the current program's hash digest
+        // let self_digest: Digest = tasm::builtins::own_program_digest();
```
aszepieniec, line 26: We only need this for recursion :-)

neptune-core/src/application/util_proof/sent/spec.rs
```diff
@@ -0,0 +1,106 @@
...
+    fn source(&self) {
+        // get in the current program's hash digest
+        // let self_digest: Digest = tasm::builtins::own_program_digest();
+
+        // read standard input
+        let publicinput_releasedate = tasm::builtins::tasmlib_io_read_stdin___bfe();
+        let publicinput_receiverdigest: Digest = tasm::builtins::tasmlib_io_read_stdin___digest();
+
+        // divine witness from memory
+        let ram: super::WitnessMemory = tasm::builtins::decode_from_memory(
```
aszepieniec: I think `witness` is a better variable name.

neptune-core/src/util_types/sent/spec.rs
```diff
@@ -0,0 +1,106 @@
...
+        // receiver: expose the address lock script part
+        tasmlib_io_write_to_stdout___digest(ram.utxo.lock_script_hash());
+
+        // fill `Claim::output` with `sender_randomness_digest`
+        tasmlib_io_write_to_stdout___digest(ram.sender_randomness.hash());
+
+        let utxo_digest = Tip5::hash_varlen(ram.utxo.encode().as_slice());
+
+        // constraint consistency for two parts of the witness
+        // assert_eq!(ram.utxo_digest, utxo_digest);
```
aszepieniec, line 47: Very much necessary. Why is it commented out?

neptune-core/src/application/util_proof/sent/tests.rs
```diff
@@ -0,0 +1,311 @@
...
+        aocl_leaf_index,
+    )
+}
+
+// #[test]
+// fn smoke() {super::library_and_code();}
+
+// TODO tune `cases` but the single speeds up the development
+#[test_strategy::proptest(async = "tokio", cases = 1, rng_seed = RngSeed::Fixed(0))]
+async fn property_test_happy_path(
```
aszepieniec: Please describe the property that you are testing in the name of the test. The motivation is so that you can write

```
cargo test test_name
```

and run only one test without triggering a whole bunch of pattern matches.

neptune-core/src/application/util_proof/sent/tests.rs
```diff
@@ -0,0 +1,311 @@
...
+//     }
+// }
+
+// Consolidated negative test: AOCL proof verification failure.
+#[test_strategy::proptest(async = "tokio", cases = 1, rng_seed = RngSeed::Fixed(0))]
+async fn aocl_proof_verification_failed(
+    #[strategy(0.0..=1.0)] spend_percent: f64,
+    #[strategy(0.0..=1.0)] fee_percent: f64,
+    #[strategy(proptest_arbitrary_interop::arb())] rness: Randomness<0, 2>,
+    #[strategy(proptest_arbitrary_interop::arb())] aocl_mp_bad: MmrMembershipProof,
```
aszepieniec: I think there may be edge cases in which the "bad" `MmrMembershipProof` happens to be valid. You can use `#[filter()]` to require it to be invalid. Or you take the valid `MmrMembershipProof` and modify it in a way that is guaranteed to make it invalid.

neptune-core/src/application/util_proof/sent/tests.rs
```diff
@@ -0,0 +1,311 @@
...
+            assert!(
+                format!("{instruction_error:?}").contains(
+                    super::ERROR_AOCL_PROOF_VERIFICATION_FAILED
+                        .to_string()
+                        .as_str()
+                ),
+                "Expected Triton VM error id {}, got: {instruction_error}; err: {err}",
+                super::ERROR_AOCL_PROOF_VERIFICATION_FAILED
+            );
+        }
```
aszepieniec: There is an optional id on the `AssertionError` which you should be able to read out instead of digging into the rendered string.

neptune-core/src/application/util_proof/sent/tests.rs
```diff
@@ -0,0 +1,311 @@
...
+//                 "Expected Triton VM error id {ERROR_UTXO_DIGEST_NEQ}, got: {instruction_error}; err: {err}"
+//             );
+//         }
+//         Err(other) => panic!("Expected TritonVMPanic, got: {other:?}"),
+//     }
+// }
+
+// Consolidated negative test: AOCL proof verification failure.
+#[test_strategy::proptest(async = "tokio", cases = 1, rng_seed = RngSeed::Fixed(0))]
+async fn aocl_proof_verification_failed(
```
aszepieniec: Good job! This test function has the right template. Please add one for every assert statement in the specification / tasm code of the `TritonProgram`.

neptune-core/src/application/util_proof/sent/spec.rs
```diff
@@ -0,0 +1,106 @@
...
+                    utxo_digest,
+                    ram.sender_randomness,
+                    publicinput_receiverdigest
+                )
+                .canonical_commitment
+            )
+        );
+
+        // output the AOCL digest used for the proof
+        tasmlib_io_write_to_stdout___digest(ram.aocl.bag_peaks());
```
aszepieniec: "AOCL digest" is not a good term if you refer to the result of bagging the peaks in the AOCL. Because the AOCL contains commitments which are also digests.

neptune-core/src/api/wallet/wallet_impl.rs
skaunov: # a doctest
I spent some time thinking about this when working on the earlier an example suggestion. As it needs a mined spent tx, two addresses, and the wallet --- could you show me anything similar in the code, I'd build off something existing, especially the testing/mocking utils.

# another `Coin`
I remember we had the discussion on this like two months ago after you explained me the mechanism in general on the Discourse. (I don't remember where was the discussion - on IRC, or Tg, or there, or here. :face_with_spiral_eyes:) I left with the feeling I raised this concern and we concluded it's not an issue at this stage. Like a thing which should be added when this simpler approach/iteration works (in the wild). I think adding an issue on it when merged is the best choice.

# The rephrase preformat...

...seems to be more like a summary of the code, but that's just me. `Coin` can be single. But generally no objects for it instead the docs I wrote. 

Two items mentioned are not disclosed! 
- The sender randomness **digest** is disclosed which serves just to be able to distinguish the proofs/arguments with the same parameters. Does that helps with the quoted sentence? 
- The addition record is only proved AFAIR.

# The three things
are there indeed. I'd make it a post-merge issue just because I'd separate it not without fine types as all the interactions are not that general/reusable, but introducing the types now could restrict the next proof features designs. I mean it would make sense to do that basing on the needs and hopefully more similar modules to structure.

neptune-core/src/api/wallet/wallet_impl.rs
skaunov: I guess I just tend to use an item directly if it's needed only once in the module. Gives so much more context, takes less space/lines/volume. I mean both should work the same.

neptune-core/src/api/wallet/wallet_impl.rs
skaunov: Because I'm bad at naming entities. I'm glad to put a better name on this. 

I found it more natural to use the module name, but am not insisting at all.

neptune-core/src/application/rpc/server.rs
skaunov: (also the point below can be a step to associated types in this part of the type system; I noted in a code comment how it would fit/benefit)

`struct Claim` holds just field elements which to me is bad for the type system; from the fact that it allows to feed a claim from one `ConsensusProgram` to another, to that it's not possible to check if some bunch of elements are valid structured via deserialization trials. I'd prefer a well-defined claim type per a program and the trait which gets the data serialized from the structure.

In other words claims for each program are generally different by their internal type, and I see benefits in reflecting those differences in the type system instead of currently losing those when instantiating `Claim`.

Third attempt to say the same in clearer words. Each program requires its own claim and then particular trait(s) needs to look at a claim in an uniform way (basically serialized) --- that's a very suitable for `trait` situation.

neptune-core/src/util_types/sent/spec.rs
skaunov: Because I removed this check from the Tasm program. It's a very recent removal. Before asking for the new review iteration I was going through the program and was trying to understand/remember how it got into. My best reconstruction is that I thought the value is part of another structure that witness loads into `ram` and I wanted to save some computing by constraining it outside Triton (maybe it still will be a case for reserves btw). But as it ended up to be fully computed inside Triton then after meticulous search I couldn't find a reason to have it the public part at all.

neptune-core/src/application/util_proof/sent/tests.rs
skaunov: It's the only one left. :shrug: Other potential assertions were basically pushed out to be the public outputs to be checked outside of Triton. (There was another one which left the yet commented out error code, but it has its own thread here.)

neptune-core/src/application/rpc/server.rs
aszepieniec: Thanks, that's certainly some nutritious food for thought :-)

I see how a lot of things can still work if `Claim` is a trait instead of a struct -- and indeed, work *better* (arguably). But in some cases I can't complete the picture. For instance, what happens in the block appendix? It is a list of claims that can be extended in the future, with the salient feature that the block proof establishes the truth of all claims in the list. If `Claim` isn't a struct, what do we store here instead?

neptune-core/src/application/util_proof/sent/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+    let rustfield_utxo = rustfield!(ProofOfTransferWitness::utxo);
+    let rustfield_utxo_digest = rustfield!(ProofOfTransferWitness::utxo_digest);
+    let rustfield_aoclleafindex = rustfield!(ProofOfTransferWitness::aocl_leaf_index);
+
+    let main = triton_asm! {
+        // _
+        push {FIRST_NON_DETERMINISTICALLY_INITIALIZED_MEMORY_ADDRESS}
+        {&rustfield_utxo}
+        {&Utxo::get_field("lock_script_hash")}
+        addi 4 read_mem 5 pop 1
```
aszepieniec:
```suggestion
        addi {Digest::LENGTH-1} read_mem {Digest::LENGTH} pop 1
```

Please use the const `Digest::LENGTH` to indicate where the magic numbers 4 and 5 come from: `addi {Digest::LENGTH - 1} read_mem {Digest::LENGTH} pop 1`. You do use this pattern other times, but if you don't use it consistently it raises the impression that the other cases are intentional -- and leaves the reader guessing about that intention.

neptune-core/src/application/util_proof/sent/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+        dup 0
+        {&MmrAccumulator::get_field("peaks")}
+        // _ *aocl *aocl_peaks
+
+        // read `receiver_digest`
+        read_io 5
+        // _ *aocl *aocl_peaks [receiver_digest]
+
+        push {FIRST_NON_DETERMINISTICALLY_INITIALIZED_MEMORY_ADDRESS}
+        // _ *w
```
aszepieniec:
```suggestion
        // _ *aocl *aocl_peaks [receiver_digest] *w
```

neptune-core/src/util_types/sent/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+        write_mem 1
+        pop 1
+        // _
+
+        // prepare the stack and call the `BasicSnippet` to get the UTXO amount
+        push {FIRST_NON_DETERMINISTICALLY_INITIALIZED_MEMORY_ADDRESS}
+        {&rustfield_utxo}
+        {&Utxo::get_field("coins")}
+        // _ *coins
+        addi 0 read_mem 1 addi 1
```
aszepieniec, line 208: `addi 0` is a nop. If this is the intention, please explain in a comment.

neptune-core/src/util_types/sent/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+        addi 1
+        // _ SI_coins *coins[0]_si
```
aszepieniec, line 210: Unless I am mistaken, memory is laid out as

```
size_indicator length_indicator [element 0] [element 1] ...
```

where every element itself has a variable length and therefore comes prepended with a size indicator of its own.

And so the stack evolves as

| stack | instruction |
|--|--|
| `_ *coins` | |
| | `addi 0` |
| `_ *coins` | |
| | `read_mem 1 addi 1` |
| `_ coins_si *coins` | |
| | `addi 1` |
| `_ coins_si *coins_si` | |

which disagrees with the comment. I might be wrong -- please double-check.

Also note that you can collapse two consecutive `addi` instructions.

neptune-core/src/application/util_proof/reserves/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+            call {lib_ms_commit}
+            // *aocl *aocl_peaks [canonical_commitment]
+
+            /* pasted '7.' from <removal_records_integrity.rs>
+            ============== */
+            /* 7. */
+            dup 6 {&MmrAccumulator::get_field("leaf_count")}
+            // *aocl *aocl_peaks [canonical_commitment] *num_leafs
+
+            addi 1 read_mem {u64_stack_size} pop 1
```
aszepieniec:
```suggestion
            addi {u64_stack_size - 1} read_mem {u64_stack_size} pop 1
```

neptune-core/src/application/util_proof/reserves/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+            // *aocl *aocl_peaks [canonical_commitment] [num_leafs]
+
+            push {FIRST_NON_DETERMINISTICALLY_INITIALIZED_MEMORY_ADDRESS}
+            {&rustfield_membershipproof}
+            {&rustfield_aoclleafindex}
+            read_mem {u64_stack_size}
```
aszepieniec:
```suggestion
            addi {u64_stack_size - 1} read_mem {u64_stack_size}
```

neptune-core/src/application/util_proof/reserves/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+            {&WitnessMemory::get_field("authentication_path")}
+            // *aocl *aocl_peaks [canonical_commitment] [num_leafs] [aocl_leaf_index] *auth_path
+            swap 2
+            // *aocl *aocl_peaks [canonical_commitment] *auth_path [aocl_leaf_index] [num_leafs]
+            swap 3
+            // *aocl *aocl_peaks [num_leafs] *auth_path [aocl_leaf_index] [canonical_commitment]
+            swap 1
+            // *aocl *aocl_peaks [num_leafs] *auth_path [canonical_commitment] [aocl_leaf_index]
+            swap 2
+            // *aocl *aocl_peaks [num_leafs] [aocl_leaf_index] [canonical_commitment] *auth_path
```
aszepieniec: I don't think this will work. Instruction `swap n` swaps the BFieldElement on top of the stack with the one at index `n`. The objects `[num_leafs]` and `[aocl_leaf_index]` are pairs of BFieldElements, representing u64s.

neptune-core/src/application/util_proof/reserves/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+            {&rustfield_membershipproof}
+            {&rustfield_senderrandomness}
+            read_mem {Digest::LEN}
+            pop 1
+            // *aocl [aocl_leaf_index] [receiver_preimage] [sender_randomness]
+
+            push {FIRST_NON_DETERMINISTICALLY_INITIALIZED_MEMORY_ADDRESS}
+            {&rustfield_utxodigest}
+            read_mem {Digest::LEN}
+            pop 1
```
aszepieniec: Note that the `field` helpers set the pointer to the memory object's first element in memory. In memory, objects are laid out forwards meaning that the rest of the object is located at higher addresses. However, `read_mem` *decreases* the memory pointer so that after the read operation the same memory object is laid out forwards on stack, with the first word at stack index 0 and later words at higher indices. Right now you're reading in the wrong direction: you're reading the first word of the targeted object and the last few words of the object that came before it.

neptune-core/src/application/util_proof/reserves/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+            // *coins
+            read_mem 1 addi 2
+            // SI_coins *coins[0]
+            push 0 swap
+            // SI_coins 0 *coins[0]
+            push 0
+            push 0
+            push 0
+            push 0
+            // SI_coins 0 *coins[0] amount timelocked_amount utxo_amount utxo_is_timelocked
```
aszepieniec: There are a bunch of instructions missing here, right? Because I think the stack at this point has this shape:

```
_ SI_coins 0 *coins[0] [amount]
```

neptune-core/src/util_types/sent/mod.rs
skaunov: I'm ok to just trust you! X) Would be nice to have a piece of documentation on `length_indicator` anyway to start learning about it; I could not find any yet.

What do you mean by `coins_si`? I thought that `*coins` and `*coins_si` is the same thing, so if I used that previously I took an effort to unify it to `*coins` IIRC.

Is `length_indicator` takes a single field element, so should last `addi` be just one bigger?

neptune-core/src/api/wallet/wallet_impl.rs
KaffinPX: > I guess I just tend to use an item directly if it's needed only once in the module. Gives so much more context, takes less space/lines/volume. I mean both should work the same.

Translation: "AI told me to."

skaunov: If you call `rust-analyzer` so...

Can you keep the discussions constructive?

KaffinPX: Pretty sure `rust-analyzer` sticks to autocomplete and imports, let me know how you got it to take prompts 🤓

Sure but you denying your tools will make you a target of Skynet later. :)

skaunov: I'm not seeing AI/LLM tools as something bad per se. The tests in this branch I 85% generated. (With that funny parsing @aszepieniec suggestion straightened.) I guess @aszepieniec used some while adding `libp2p`.

I'm not compelled to a discussion in the format you are sticking too; and I don't appreciate your humor. :clown_face: So I rather won't go into details or/and your assumptions.

KaffinPX: I never said it was bad. LLMs are tools like anything else, and their output is only as good as the prompts...

Anyway im not making assumptions about you, only reviewing the code, the comments, and the impressions they create

neptune-core/src/util_types/sent/mod.rs
aszepieniec: The trait and derive macro for `BFieldCodec` are defined in [`twenty-first`](https://github.com/Neptune-Crypto/twenty-first). I'm afraid there is no documentation but there are examples. And in the worst case you can `BFieldCodec`-encode an object and inspect the resulting words.

The `length_indicator` is only used for lists where the number of elements is not known at compile time. It is a single word -- you guessed it --indicating  the number of elements in the list.

The difference between `*coins_si` and `*coins` is subtle because of nested data structures. They both contain the same address. The second one contains the address of the first word in the data object `coins`. This first word happens to coincide with the size indicator for that object. That size indicator is itself a data object called `coins_si` and `*coins_si` points to its first word. Obviously the two pointers' values are identical. The difference between `*coins_si` and `coins_si` is the difference between the pointer and the value pointed to.

aszepieniec:
> Have you shared those comments? It's a complicated topic...

I'm afraid I did not save them.

neptune-core/src/application/rpc/server.rs
skaunov: Done, *except the verification part.* An easy approach for me would be just slap `verify_` function(s) near `prove_` and call it a day, but I put some effort to stop there. =( I do see the proving part in `neptune-core` as we don't really have a competing wallet (and a setup VM with a queue taking the jobs). But verification needs so less things. 
- Triton verifier (in Wasm will do I guess)
- the corresponding block - can be taken from an explorer
- `Claim` parser/formatter to check the facts

So I would argue verification doesn't belong much to `neptune-core` as it doesn't really depend on it. It's possible to bring it there (easily!), but I feel like it only will encourage some bad design of needing the whole `...core` for a light/specialized task.

PR #802, opened 2025-12-04
sec: Limit bincode frame length

aszepieniec:
Countermeasure against bombs (short messages that cause large allocations).

Credit to AllFather team.


aszepieniec:
Closed in favor of #804.

PR #809, opened 2025-12-09
feat: Compute and show max, circulating, and burned supply

aszepieniec:
Add functions for computing:
 - The amount mined between the genesis block and a given block height.
 - The circulating supply.
 - The max supply.
 - The burned supply.

The last method relies on another, `known_burns` which lists hints about where to find burned coins. These hints are validated against the archival state, so even if there is a deep reorganization invalidating some of these hints, there is no possibility for showing the wrong number.

The function `known_burns` currently only contains valid information for the one big burn at 16999/3 where 1'526'640 coins were burned. The information for validating the other two known burns is, for now, missing.

### DIFF
--- a/neptune-core/src/application/json_rpc/core/model/message.rs
+++ b/neptune-core/src/application/json_rpc/core/model/message.rs
@@ -262,7 +262,7 @@ pub struct FindUtxoOriginResponse {
     pub block: Option<Digest>,
 }
 
-#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
+#[derive(Clone, Copy, Debug, Serialize_tuple, Deserialize_tuple)]
 #[serde(rename_all = "camelCase")]
 pub struct CirculatingSupplyRequest {}
 
@@ -272,7 +272,7 @@ pub struct CirculatingSupplyResponse {
     pub amount: RpcNativeCurrencyAmount,
 }
 
-#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
+#[derive(Clone, Copy, Debug, Serialize_tuple, Deserialize_tuple)]
 #[serde(rename_all = "camelCase")]
 pub struct MaxSupplyRequest {}
 
@@ -282,7 +282,7 @@ pub struct MaxSupplyResponse {
     pub amount: RpcNativeCurrencyAmount,
 }
 
-#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
+#[derive(Clone, Copy, Debug, Serialize_tuple, Deserialize_tuple)]
 #[serde(rename_all = "camelCase")]
 pub struct BurnedSupplyRequest {}
 
--- a/neptune-core/src/application/json_rpc/README.md
+++ b/neptune-core/src/application/json_rpc/README.md
@@ -11,7 +11,8 @@ Let’s say you want to add a new method called `submit_block`.
 2. **Create request and response types**
 
    * In `json_rpc::core::model::message`, define a matching pair of structs:
-     `SubmitBlockRequest` and `SubmitBlockResponse`.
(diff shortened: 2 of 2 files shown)


neptune-core/src/state/archival_state.rs
```diff
@@ -1407,6 +1419,249 @@ impl ArchivalState {
...
+                // Since we don't know the lock script hash of this UTXO, we
+                // actually can't authenticate the amount in this burn.
+                Digest::try_from_hex("00000000000000000000000000000000000000000000000000000000000000000000000000000000").unwrap(),
+                Digest::try_from_hex("00000000000000000000000000000000000000000000000000000000000000000000000000000000").unwrap(),
+                Digest::try_from_hex("00000000000000000000000000000000000000000000000000000000000000000000000000000000").unwrap(),
+            ),
+            // TODO: add data for UTXO with AOCL leaf index 78179. Apparently
+            // 0.1 NPT was burned here (receiver digest set to zero and
+            // lock script "burn"). However, unless we know the block we have no
+            // way of verifying this claim.
```
Sword-Smith: https://explorer.neptune.cash/announcement/digest/85ba160eab02c8e39bae40e68e001d1220902f31cb869e320b52afab9a5badf6855b020000000000/1

aszepieniec: Thanks!

neptune-core/src/application/json_rpc/core/api/ops.rs
```diff
@@ -96,6 +96,15 @@ pub enum RpcMethods {
     #[namespace(Namespace::Archival)]
     FindUtxoOrigin,
 
+    #[namespace(Namespace::Archival)]
+    GetCirculatingSupply,
```
KaffinPX: If RPC method is a `getter`(no parameter intake), avoid get prefix

neptune-core/src/application/json_rpc/core/api/rpc.rs
```diff
@@ -268,6 +268,21 @@ pub trait RpcApi: Sync + Send {
         request: FindUtxoOriginRequest,
     ) -> RpcResult<FindUtxoOriginResponse>;
 
+    async fn get_circulating_supply_call(
```
KaffinPX: Standard non _call functions are needed for better UX on clients, please add corresponding fns

neptune-core/src/application/json_rpc/core/model/message.rs
```diff
@@ -262,6 +262,36 @@ pub struct FindUtxoOriginResponse {
     pub block: Option<Digest>,
 }
 
+#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
```
KaffinPX: Request params are supposed to be a tuple, use Serialize_tuple and Deserialize_tuple macros

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -1008,4 +1066,31 @@ pub mod tests {
             RpcError::SubmitBlock(SubmitBlockError::InvalidBlock)
         );
     }
+
+    #[apply(shared_tokio_runtime)]
+    async fn supply_methods_return_reasonable_results() {
+        let rpc_server = test_rpc_server().await;
+        let circulating_supply = rpc_server
+            .get_circulating_supply_call(GetCirculatingSupplyRequest)
```
KaffinPX: After adding standard non _call client calls, you can simplify these Fns

neptune-core/src/application/json_rpc/core/model/message.rs
aszepieniec: I fixed it in line with this recommendation just now.

I wonder if there is a test that could catch the derivation of the wrong serializer/deserializer? Or if that test already exists, why did it not fail?

PR #821, opened 2025-12-17
Utxo index / rescan

Sword-Smith:
Add a new database to the archival state: UTXO index whose purpose it is to allow for fast rescanning of blocks for balance-affecting UTXOs.

Focus for these "rescanning" methods is performance, they should be able to complete fast.

A goal for the "rescan" methods `GlobalState` is to allow for the rebuilding of the list `MonitoredUtxo` from an empty wallet database, so announced UTXOs, expected UTXOs, guesser rewards, and expenditures (the spending of UTXOs) must all be covered. Some of these can be handled without the new UTXO index. In fact UTXO index is only needed for the rescanning of announced UTXOs and expenditures. 

Missing:

- [x] Endpoints -- how should the user interface with these methods? (tarps + json RPC)
- [x] More tests
  - [x] idempotency of rescan
  - [x] balance equality before clear and after clear and rescan
  - [x] correct handling of guesser rewards
  - [x] correct balance history restoration through rescan methods
- [x] Benchmarks
- [x] Description of UTXO index in docs

**Tests**
- Two unit tests in `rusty_utxo_index.rs` of basic UTXO index behavior.
- `monitored_utxos_from_rescan_of_incoming_utxos` verifying that incoming UTXOs with associated announcements are correctly discovered through rescan.
- `rescan_guesser_rewards` -- verifying guesser rewards can be discovered through rescan.
- `wallet_history_from_rescan` -- verifying rescanning recovers complicated UTXO history with balance changes, for UTXOs with and without on-chain announcements.
- I successfully recovered wallet history for a wallet with ~300 UTXOs on main net.

**Disk usage when UTXO index is chosen**
- 40 bytes per block
- 16 bytes per announcement in each block
- 40 bytes per input in each block

Activating this on main net on a running node of mine shows that the entire index is only 11MB.

**Benchmarking results**
Scanning 110 blocks, of which 10 contain UTXOs for the wallet, and 100 are empty:
```text
╰─ rescan                                              │               │               │               │         │
   ├─ rescan_110_blocks_announced_bench  119.8 ms      │ 168.2 ms      │ 131.6 ms      │ 134.6 ms      │ 10      │ 10
   ├─ rescan_110_blocks_expected_bench   37.08 ms      │ 41.9 ms       │ 37.98 ms      │ 38.37 ms      │ 10      │ 10
   ╰─ rescan_110_blocks_outgoing_bench   843.9 µs      │ 1.552 ms      │ 889.6 µs      │ 1.037 ms      │ 10      │ 10
```

doing a full rescan of ~22.000 blocks on main net on a wallet that has received ~330 UTXOs takes less than two seconds on one of my machines (with an admittedly very fast SSD). I'm very pleased with these results.


Sword-Smith:
Missing:

- [ ] Endpoints -- how should the user interface with these methods?
- [ ] More tests
  - [ ] idempotency of rescan
  - [ ] balance equality before clear and after clear and rescan
  - [ ] correct handling of guesser rewards
- [ ] Benchmarks?

docs/src/SUMMARY.md
```diff
@@ -23,6 +23,7 @@
   - [Generating Addresses](./user-guides/generating-addresses.md)
   - [Scan Mode](./user-guides/scan-mode.md)
   - [Shamir Secret Sharing](./user-guides/shamir-secret-sharing.md)
+  - [Wallet Rescan](./user-guides/wallet-rescan.md)
```
aszepieniec, line 26: :+1:

docs/src/neptune-core/overview.md
```diff
@@ -28,7 +28,7 @@ The global state has five fields and they each follow some rules:
...
+  `archival_state` consists of data stored both in a database and on disk. The blocks themselves are stored on disk, and meta-information about the blocks are stored in the `block_index` database. `archival_state` also contains the `archival_mutator_set` which can be used to recover unsynced membership proofs for the mutator set. `archival_state` also contains an optional UTXO index that can be maintained by starting the node with the `--utxo-index`. The UTXO index contains a digest for each historical transaction input, and two b-field elements for each historical announcement. Exchanges and mining pools should run the node with the UTXO index, as this makes the rescanning of UTXOs for once's own wallet possible. The reason that all transaction outputs are not part of the UTXO index is that the transaction outputs are already present in another database, the `archival_mutator_set`.
```
aszepieniec: - Why is it optional? Is it a bad idea to force everyone to do this?
 - An index is an integer pointing into a list. However, if I read right then the UTXO index is a container of either digest or BFieldElement pairs. That's confusing. Have you considered other names?
 - A pair of BFieldElements sounds rather arbitrary. What's the relation between the announcement and the pair of BFieldElements?
 - The last sentence reads rather clunky and I think it is because of the inclusion of the word "all", which suggests that *some* transaction outputs *are* present -- which, if so, begs the question: why?

neptune-core-cli/src/main.rs
```diff
@@ -259,6 +259,38 @@ enum Command {
         max_search_depth: Option<u64>,
     },
 
+    /// Rescan the selected (inclusive) range of blocks for announced, incoming
```
aszepieniec, line 262: In `neptune-core` (the crate) there is the helper function `parse_range` which allows the argument to be a `RangeInclusive<u64>`. It is used, for instance, in the argument `scan_blocks`. Consider using a single consistent format for CLI ranges.

neptune-core-cli/src/main.rs
```diff
@@ -259,6 +259,38 @@ enum Command {
...
+    /// the UTXO index activated.
+    RescanOutgoing {
+        first: u64,
+        last: u64,
+    },
+
+    /// Rescan the selected (inclusive) range of blocks for guesser rewards.
+    /// Useful if the client's seed has been used to guess on correct proof-of-
+    /// work solutions in the past but wallet state was somehow lost. Works
+    /// regardless of UTXO index status.
```
aszepieniec, line 288: I think it's fairly common for miners to use the same seed on one machine that does the guessing, and another machine that produces the transactions. In this case, the wallet is not *lost* but merely not transmitted to the other machine.

neptune-core/src/application/config/cli_args.rs
```diff
@@ -543,6 +543,14 @@ pub struct Args {
     #[clap(long)]
     pub(crate) scan_keys: Option<usize>,
 
+    /// Construct an maintain a UTXO index
```
aszepieniec:
```suggestion
    /// Construct and maintain a UTXO index
```

neptune-core/src/application/config/cli_args.rs
```diff
@@ -543,6 +543,14 @@ pub struct Args {
...
+    ///
+    /// If set, all announcements and outputs in all processed blocks will be
+    /// indexed in a database that enables a fast rescan for the discovery of
+    /// all balance-affecting inputs and outputs of blocks.
+    #[clap(long)]
+    pub utxo_index: bool,
```
aszepieniec: What happens to blocks already processed before the node was restarted with this flag enabled? My guess is that the UTXOs in those blocks are not indexed in the database. Please clarify.

If such UTXOs are not indexed in the database, what can I do to get populated database if my node was already running without this flag?

neptune-core/src/application/json_rpc/core/api/ops.rs
```diff
@@ -114,6 +114,18 @@ pub enum RpcMethods {
...
+
+    #[namespace(Namespace::Wallet)]
+    RescanExpected,
+
+    #[namespace(Namespace::Wallet)]
+    RescanOutgoing,
+
+    #[namespace(Namespace::Wallet)]
+    RescanGuesserRewards,
+
```
aszepieniec, line 129: I think the "Archival" namespace makes more sense because these commands do not make any sense for non-archival nodes. We foresee having non-archival nodes in the not-too-far future.

They do make sense for non-wallet nodes when they serve wallet nodes through a foreign interface. In that case the argument list would have to be expanded with a set of addresses or something like that.

neptune-core/src/application/rpc/server.rs
```diff
@@ -1628,6 +1628,39 @@ pub trait RPC {
         tx_artifacts: TxCreationArtifacts,
     ) -> RpcResult<()>;
 
+    /// Rescan the specified inclusive range for incoming UTXOS that were sent
+    /// with associated on-chain announements.
+    async fn rescan_announced(
+        token: auth::Token,
+        first: BlockHeight,
+        last: BlockHeight,
+    ) -> RpcResult<()>;
```
aszepieniec, line 1656: Why no `keys: Vec<SpendingKey>`?

neptune-core/src/state/archival_state.rs
```diff
@@ -93,6 +99,11 @@ pub struct ArchivalState {
     /// Archival-MMR of the block digests belonging to the canonical chain.
     pub(crate) archival_block_mmr: RustyArchivalBlockMmr,
 
+    /// Mapping from block digest to a list of (flag, receiver_id) pairs for all
+    /// announcement in the block. This index is only maintained if the node has
+    /// been started with the
```
aszepieniec:
```suggestion
    /// been started with the CLI flag `--utxo-index`.
```

neptune-core/src/state/archival_state.rs
```diff
@@ -310,6 +348,18 @@ impl ArchivalState {
             .await
             .expect("Must be able to initialize block index database");
         debug!("Got block index database");
+
+        // UTXO index is always initialized. But only populated with blocks if
+        // this index is activated. It's always populated with the genesis block
+        // though.
```
aszepieniec: This comment can be improved.

aszepieniec: Perhaps like so (suggestion from AI):

```
 // The UTXO index is always initialized, but it is only populated with blocks
 // when the index feature is enabled. The genesis block is always included by default.
```

neptune-core/src/state/archival_state.rs
```diff
@@ -310,6 +348,18 @@ impl ArchivalState {
...
+        let mut utxo_index = ArchivalState::initialize_utxo_index(&data_dir)
+            .await
+            .expect("Must be able to initialize utxo index database");
```
aszepieniec, line 357: By *why* must we be able to initialize the UTXO index database? What's the relevant context that makes the conditions for failure of the called function not apply?

Why not fail inside `initialize_utxo_index` and turn that function into something infallible?

neptune-core/src/state/archival_state/rusty_utxo_index.rs
```diff
@@ -0,0 +1,294 @@
...
+        let hash = block.hash();
+
+        let tx_kernel = &block.body().transaction_kernel;
+
+        let announcements = tx_kernel
+            .announcements
+            .iter()
+            .filter_map(|ann| ann.try_into().ok())
+            .collect_vec();
+        self.tables.announcements.insert(hash, announcements).await;
```
aszepieniec: Since `tables.announcements` does not contain a container of `Announcement`s, the type cast hidden by the `try_into()` is actually rather non-trivial. Please expand. For instance, `AnnouncementFlags::try_from(ann)`. (Right?)

neptune-core/src/state/archival_state/rusty_utxo_index.rs
```diff
@@ -0,0 +1,294 @@
...
+    }
+}
+
+impl StorageWriter for RustyUtxoIndex {
+    async fn persist(&mut self) {
+        self.storage.persist().await;
+    }
+
+    async fn drop_unpersisted(&mut self) {
+        unimplemented!("announcement index does not need it")
```
aszepieniec, line 243: How about removal records then?

neptune-core/src/state/mod.rs
```diff
@@ -743,6 +749,486 @@ impl GlobalState {
...
     }
 
+    /// Rescan the specified (inclusive) range of blocks for spends of UTXOs.
+    /// Only works for nodes that maintain a UTXO index.
+    ///
+    /// Never loads the entire block from disk, so performance is good.
+    ///
+    /// # Panics
+    /// - If start block height is greater than end block height
+    pub async fn rescan_outgoing(&mut self, first: BlockHeight, last: BlockHeight) -> Result<()> {
```
aszepieniec, line 759: It is not clear from this function name whether "outgoing" is the object of the command or some supplementary context. Suggestion: `rescan_blocks_for_spent_utxos`.

neptune-core/src/state/mod.rs
```diff
@@ -743,6 +749,486 @@ impl GlobalState {
...
+
+        Ok(())
+    }
+
+    /// Rescan the specified (inclusive) range of blocks for guesser rewards to
+    /// this node's wallet.
+    ///
+    /// # Panics
+    /// - If start block height is greater than end block height
+    pub(crate) async fn rescan_guesser_rewards(&mut self, first: BlockHeight, last: BlockHeight) {
```
aszepieniec, line 843: The next two functions feature "incoming" in the name and this one does not. That discrepancy suggests there is a way in which those UTXOs are incoming and these ones are not. Is that correct? If not, please synchronize the function name. If it is, please anticipate confusion through explanatory comments.

neptune-core/src/state/wallet/wallet_state.rs
```diff
@@ -788,19 +788,22 @@ impl WalletState {
     /// Scan the given list of addition records for items that match with list
     /// of expected incoming UTXOs, and returns expected UTXOs that are present.
     ///
-    /// note: this algorithm is o(n) + o(m) where:
-    ///   n = number of ExpectedUtxo in database. (all-time)
-    ///   m = number of transaction outputs.
+    /// note: this algorithm is o(n) where:
```
aszepieniec: $O(n)$ ? (Big O versus small o.)

aszepieniec (commented):
The engineering is on-point; it does exactly what it should be doing.

I do have some stylistic comments though:
 - Using the term "UTXO index" to denote a *database* is confusing. That phrase might also refer to a) the index of the monitored UTXO in the wallet database, or b) the AOCL leaf index. In some cases you use the phrase "UTXO index database" and that's a massive improvement. Still, let's ponder alternative names even further apart from "UTXO index". The term "registry" comes to mind. Perhaps I'm also uncomfortable with including the term "UTXO" in a description of what this database does -- because it tracks information that's only tangentially related to UTXOs. (In particular, it does not store UTXOs.) 
 - Please expand the RPC interfaces to take optional lists of `ReceivingAddress`s or hashes of `AbsoluteIndexSet`s. In future, archival non-wallet nodes may be used to support non-archival wallet nodes remotely. The RPC interface will be difficult to change, so it pays to get it right from the start.
 - Please clarify how to use (and how not to use) the `--utxo-index` flag. Consider adding a procedure at startup to test if the UTXO index databae is present already and if it is not (but the flag *is* set), then to populate the database with information from all historical blocks.

neptune-core/src/state/archival_state.rs
Sword-Smith: Because I follow the pattern of the other `initialize` functions. I try to pattern match on existing code. I guess that principle can be criticized. If I was the only developer on this project, I'd have fewer functions that returns `Result<T>` and simply return `T`, but I'm not, so I try to reuse common patterns in the code that I write.

When it comes to these database initialization functions, they generally only fail if a node is already running on the same network on the same machine, or if there is some permission problem with the data directory (e.g. the user does not own the specified data directory). These errors should, in my opinion just panic immediately when encountered but in practice we use some mix of panics and errors. As long as the errors are quickly (close in the call graph) turned into panics, I think that's acceptable though. This `expect` achieves that.

neptune-core/src/application/rpc/server.rs
Sword-Smith: Because this endpoint scans all keys known to the wallet.

neptune-core/src/application/config/cli_args.rs
Sword-Smith: Your guess is wrong but you're right that this behavior should be specified in the docstring. New docstring:

```rust
    /// Construct and maintain a UTXO index
    ///
    /// If set, all announcements and inputs in all processed blocks will be
    /// indexed in a database that enables a fast rescan for the discovery of
    /// all balance-affecting inputs and outputs of blocks.
    ///
    /// If blocks have already been processed without this flag active, and the
    /// flag is later activated, all blocks up to the current tip will be
    /// indexed. This process might take some time (probably a few minutes).
    #[clap(long)]
    pub utxo_index: bool,
```

neptune-core/src/application/json_rpc/core/api/ops.rs
Sword-Smith: These endpoints are endpoints that affect the wallet. They *do* require a UTXO index and an archive but they concern the wallet. I also asked the main RPC developer @KaffinPX about how to categorize this, and he agreed with me that `Wallet` was preferred.

I've added a new category in a later commit though `Namespace::UtxoIndex` for 3rd party requests that rely on this index, namely `block_hashes_by_flags` which given a list of announcement flags returns a list of block hashes in which the specified announcements are present.

docs/src/neptune-core/overview.md
Sword-Smith: It's optional because it's a very specific index that most users, even those that store all blocks, don't need.

The "index" part of the name is inspired by database lingo, where "adding an index to a row" means that you get much faster (usually logarithmic) lookup times when selecting entries in a database based on the values in that row. It's also used in Bitcoin Core for [something very similar](https://bitcoin.stackexchange.com/a/35708/38458).

I elaborated on the "pair of b-field elements".

Agreed with you last point: "all" should be removed.

New text:
>`archival_state`, persistent.
  `archival_state` consists of data stored both in a database and on disk. The blocks themselves are stored on disk, and meta-information about the blocks are stored in the `block_index` database. `archival_state` also contains the `archival_mutator_set` which can be used to recover unsynced membership proofs for the mutator set. `archival_state` also contains an optional UTXO index that can be maintained by starting the node with the `--utxo-index`. The UTXO index contains a digest for each historical transaction input, and two b-field elements and a digest for each historical announcement. These two b-field elements are interpreted as `AnnouncementFlag`s and can be used to quickly identify announcements that are relevant to a wallet. They are just the two first elements of each announcement in the block. Exchanges and mining pools should run the node with the UTXO index, as this makes the rescanning of UTXOs for once's own wallet possible. The reason that transaction outputs are not part of the UTXO index is that the transaction outputs are already present in another database, the `archival_mutator_set`.

Sword-Smith:
> Please expand the RPC interfaces to take optional lists of ReceivingAddresss or hashes of AbsoluteIndexSets. In future, archival non-wallet nodes may be used to support non-archival wallet nodes remotely. The RPC interface will be difficult to change, so it pays to get it right from the start.

These can be added as new RPC endpoints in the future such that the caller doesn't have to derive their announcement flags themselves but can instead rely on the server to do that.

Issue #362, opened 2025-02-04
Multi-Coin Hardware Wallet Support


aszepieniec:
If you want to use an external, multi-coin hardware wallet to safeguard your Neptune coins, then you quickly run into problems. Here is a probably incomplete and possibly inaccurate summary.

 - Basically all wallets support [BIP-32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki) in some or other capacity and this simply does not apply to Neptune because we don't use elliptic curves.
 - Users may want to use the same 24 words they have already backed up for other cryptocurrencies to backup the Neptune Cash holdings also. However, `neptune-cli` only supports 18-word seed phrases, and it will likely fail if you omit the last 6 words. Or alternatively, users who are appropriately distrustful of their machines may want to *generate* the seed on a secure hardware device and then copy it over by first exporting it as 24 BIP-39 words. These hardware devices rarely support the non-standard option of 18 words.
 - [BIP-39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) has *optional passphrases*, which we do not support. In fact, supporting optional passphrases would mean either importing SHA2-512 or breaking the standard and use Tip5 instead. I tend to prefer the former option as the relevant mechanics take place *before* conversion to `BFieldElement`s.
 - There is a standard for deriving a coin-specific master seed from a multi-coin übermaster seed: [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki)/[SLIP-44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md). Multi-coin hardware wallets generally support this standard and supply the coin-specific master seed to the wallet software.

If that summary is accurate, I derive the following tasks:
 - [ ] support importing 24-word BIP-39 seed phrases
 - [ ] support BIP-39 passwords
 - [ ] register a SLIP-44 derivation number
 - [ ] support import-and-derive-key via SLIP-44 derivation path

dan-da:
is this a mainnet blocker?

It doesn't seem like we'd want to be modifying our mnemonic phrases shortly after launch....

aszepieniec:
> is this a mainnet blocker?

No.

We do not need to modify existing functions and commands. We just need to add new ones.

PR #822, opened 2025-12-18
Feat: Accept 18 and 24 words mnemonic

hduoc2003:


aszepieniec:
I anticipate that a) compatibility with hardware wallets, and b) compatibility with other blockchains (for users (including exchanges) who need to interface with multiple cryptocurrencies) are important motivations. @hduoc2003: would you mind expanding on these motivations and adding any that I am missing?

As I see it, this PR is addressing two separate, but related, concerns.

 1. Support for 24-word mnemonic seed phrases (import + export).
 2. Optionally use 256 bits to represent the secret key material.

It is quite obviously moot to do (2) without doing (1): where would the remaining 64 bits of entropy come from or go to? However, it is not unthinkable to do (1) without doing (2). Why did you decide against this strategy?

I confess I am opposed to using more than 192 bits for the secret key material because 192 is already more than enough. Every derivation path that involves the secret key material passes it through a hash function first. Any attack on the secret key material involves attacking one-wayness or second pre-image resistance of that hash function. The complexity of this task is linear in the size of the output space, or 192 bits. In particular, we need not worry about the square root complexity associated with finding collisions. However, I am willing to hear a counter-argument.

Regardless of whether we decide to do just (1) or (1) and (2), I should like to use this opportunity to address the concerns raised in #362. Would you mind commenting on the correctness and completeness of that summary, and on the feasibility of expanding the scope of this PR to include the tasks described therein?

Furthermore, in case it should be mentioned: a very important constraint for this and similar PRs is that *existing seed phrases and secret-sharings continue to work*. Could you add unit tests showing that this is indeed the case?

PR #835, opened 2026-01-15
Nix packages and reproducible dev shell using flake

MathisWellmann:
[Nix] is a declarative approach to creating reproducible packages and artifacts and makes it easy for users to run the `neptune` software.

Basically this PR allows a few things:
- Easily run the reproducible package directly from github with `nix run github:Neptune-Crypto/neptune-core`
  or a specific binary with `nix run github:Neptune-Crypto/neptune-core#neptune-cli` for example.
- Locally with the git repo cloned the above simplifies to `nix run` or `nix run .#neptune-cli`
- The `flake.nix` provide a declarative and reproducible development shell with some rust tooling which can be entered using `nix develop`, drastically simplifying dependency management between different developers (no more "works on my machine" type stuff).
- And yes, this flake provides packages for different target systems, by default `["x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin"]`

## Changes include:
- Adds `flake.nix` which defines the input dependencies, packages, apps and devShell.
- The `flake.lock` file pins the system dependencies, guaranteeing reproducibility for packages, devShells etc.
- some README.md changes to highlight the nix path, which requires much fewer steps and does not suffer from non-reproducibility.
- Include `.envrc` which is useful with `direnv` to automatically enter the flake dev shell when `cd`ing into the directory.

## Future outlook
In the future I'll try to upstream the packages to [`nixpkgs`] so it can join the over 126000 packages and be more widely distributed. This will allow people to run `nix-shell -p neptune-core` for example as well to test it out quickly.

I'll work on creating NixOS options for it so it becomes trivial for people to run the software on NixOS servers, similar to how the monero node currently works, e.g:
```nix
  services.monero = {
    enable = true;
    rpc = {
      address = "0.0.0.0";
      port = 18081;
      restricted = true;
    };
  };
```
So we eventually have something similar:
```nix
services.neptune.enable = true;
```

Anyways, thanks for the great project. Looks very promising and I can't wait to dig into it more.

[Nix]: https://nixos.org/
[`nixpkgs`]: https://github.com/NixOS/nixpkgs

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Neptune-core is the reference implementation for the [Neptune Cash](https://nept
 
  -  Make sure you have `nix` installed or install it with:
     ```shell
-    sh <(curl --proto '=https' --tlsv1.2 -L https://nixos.org/nix/install) --daemon`
+    sh <(curl --proto '=https' --tlsv1.2 -L https://nixos.org/nix/install) --daemon
     ```
  -  Now simply run it with:
     - `nix run github:Neptune-Crypto/neptune-core` will run the `neptune-core` binary, fetching the git repo automatically.


README.md
```diff
@@ -18,6 +18,21 @@ Neptune-core is the reference implementation for the [Neptune Cash](https://nept
 
 ## Installing
 
+### Build with [Nix](https://nixos.org/)
+
+ -  Make sure you have `nix` installed or install it with:
+    ```shell
+    sh <(curl --proto '=https' --tlsv1.2 -L https://nixos.org/nix/install) --daemon`
```
Sword-Smith: Last backtick should be removed here, I think.

Sword-Smith (changes requested):

README.md
MathisWellmann: Good catch!
Corrected now

Sword-Smith (approved):

PR #873, opened 2026-02-28
refactor!(wallet): Add RPC support, explicit unlocking step

aszepieniec:
1. Refactor `WalletStatus`

 1.a) Remove membership proofs from WalletStatus.

Motivation: performance. Accelerates assembling and doing operations with the `WalletStatus`. Note that the archival node does not need the membership proofs to live on `WalletStatus`; they are fetched from the archival mutator set just in time for initiating transactions. The light node does not have the archival mutator set and does need membership proofs to initiate transactions but can afford to get them from the monitored UTXOs database because, after all, the `StrongUtxoKey`, or the data from which it can be derived, is available.

 1.b) Disable getting balances from `WalletState` some of the time and
      from `WalletStatus` other times.

Motivation: confusing and error-prone distribution of responsibility. The new one-way cascade `WalletState` -> `WalletStatus` -> balances concentrates responsibility and logic into `WalletStatus`.

 1.c) Expand `WalletStatus` with Mempool UTXOs.

Motivation: without knowledge of mempool UTXOs, the `WalletStatus` cannot answer questions about unconfirmed balances -- which is why some of the time the user had to get that info from `WalletState`.

 1.d) Allow the user to specify block height to classify
      balance-updates as confirmed versus unconfirmed.

Motivation: the RPC endpoint specification states that the user must be able to specify a number of confirmations relative to which the balance is calculated. In practice, the RPC handler fetches the block height, subtracts the number-of-confirmations, and uses that threshold parameter.

 2. Add RPC support for getting balance with n confirmations.

Motivation: specified by RPC endpoints for exchange integration. Exchanges have an incentive to a) avoid crediting users before n confirmations; and b) to avoid repeating the bookkeeping done when initating a withdrawal transaction or when
consolidating-to-cold-wallet UTXOs that may be reorganized away.

 3. Rename `mempool_spent_utxos` and `mempool_unspent_utxos` (both fields and functions living on `WalletState`) into `mempool_outgoing_utxos` and `mempool_incoming_utxos`.

Motivation: the term "spent UTXOs" is ambiguous but certainly one of its two senses succeeds to denote own UTXOs that are about to be spent by transactions that now live in the mempool but are about to be confirmed by a block. However, the term "unspent UTXOs" is the opposite but on the wrong dimension. The UTXOs this term is supposed to describe *are* unspent but *all mempool UTXOs are unspent* (and "spent UTXOs" too). The thing denoted by this variable and function is UTXOs that the wallet recognizes as its own, and that are generated by transactions that now live in the mempool but are about to be confirmed. The correct dimension for oppositional naming is not spent versus unspent, but incoming versus outgoing.

 4. Change type of `mempool_unspent_utxos_iter()` (now `mempool_incoming_utxos_iter()`) from `(Utxo, AdditionRecord)` to `(Utxo, Digest, Digest)` where the two digests are the sender randomness and the receiver preimage, respectively.

Motivation: one of the responsibilities of the `WalletStatus` is to hold UTXOs for sampling inputs for initiating payments. In a future world where you can spend incoming mempool UTXOs (such as in a chaining transaction), you will certainly need to demonstrate knowledge of the preimage of the addition record. So the (item, sender randomness, receiver postimage) should be considered information needed to spend the UTXO. The jury is still out on whether the receiver postimage suffices or whether the receiver preimage is needed, so the choice for the latter option errs on the safe side. Note that the addition record can be computed from the available data with very little hashing.

 5. Rename `InputSelectionPolicy` to `InputSelectionPriority`.

Motivation: "policy" is more general and will be used to include other things, such as which inputs to filter out. The enum in question is really only about how to sort inputs that are already established to be admissible candidates.

 6. Add `InputSelectionPriority` variant `ByAge`.

Motivation: sometimes, like when consolidating UTXOs, you want to spend the oldest UTXOs first. The comment (now dropped) suggested it was difficult to do based on block height. And that might be true, but you do have access to the AOCL leaf index, which is used instead.

 7. Add and use struct `InputSelectionPolicy` over `InputSelectionPriority`.

Motivation: "policy" is a strict superset of "priority". Right now it also contains a threshold number of confirmations below which to disqualify inputs.

 8. Separate input selection from unlocking.

Previously, the list of all spendable inputs were obtained, sent to the transaction builder API, which proceeded to make a selection of inputs and build a transaction based on that selection. The where and how of unlocking those inputs was a little muddy and under-performant. This refactor changes the data types and data flow to the following.

`WalletState`  |  User calls `get_wallet_status` to generate a wallet
      |        |  status.
      v
`WalletStatus` |  `WalletStatus` contains a list of all UTXOs the
      |        |  wallet knows about, both synced and unsynced fetched
      |        |  from the database and unconfirmed fetched from the
      |        |  mempool.
      |        |  The user calls `spendable_inputs` to get a list of
      |        |  candidate inputs for building a transaction. These
      |        |  candidate inputs are not unlocked.
      v
`InputCandidate` | (Renamed from `TxInput`.) The difference between a
      |          |  `InputCandidate` and a raw UTXO known to be under
      |          |  management by the wallet is metadata for enforcing
      |          |  a transaction input selection policy.
      v
`InputSelector` | (Renamed from `TxInputListBuilder`.) The
      |         | `InputSelector` makes a selection of the given
      |         | inputs suitable for the intended transaction.
      v
`GlobalState`  |  The selected inputs need to be unlocked: extended
      |        |  membership proofs and witness data that enable the
      |        |  production of the proofs in the proof collection
      |        |  relative to these inputs. The reason why this
      |        |  process involves `GlobalState` is because the
      |        |  `GlobalState` knows whether it is archival or light.
      |        |  If it is a light node, then the membership proofs
      |        |  are fetched from the wallet. If it is an archival
      |        |  node, then the membership proofs are fetched from
      |        |  the archival mutator set. In either case, the
      |        |  remaining witness data are fetched from the wallet.
      v
`UnlockedUtxo` |  An `UnlockedUtxo` is ready for the proof collection
      |        |  production step. But as a convenience layer it is
      |        |  wrapped in a ...
      v
`TxInputs`     |  (Renamed from `TxInputList`.) ... vector new type.
               |  The `TxInputs` is what lives on the
               |  `TransactionDetails` object.

This sequence cleanly separates *input selection* from the *input unlocking*. All inputs that reach the selection stage are assumed to be unlockable. Unlocking is relatively expensive, so you want to do it on as few inputs as possible. By separating this step out and putting it after input selection, we shrink the workload.

A note on filtering out inputs based on mempool expenditures: the `WalletStatus` method `spendable_inputs` filters out synced inputs based on whether their AOCL leaf indices are contained in a hash set of AOCL leaf indices generated from mempool inputs. Previously this filtering used the absolute index sets. However, the view of known inputs to mempool transactions is already generated using absolute index sets (so matching based on absolute index sets constitutes repetition of work) and their AOCL leaf indices are available at that point. This change just keeps those AOCL leaf indices handy until they are used by the `WalletStatus`.

Also, rename:
 - `TxInput` to `InputCandidate`;
 - `TxInputList` to `TxInputs`;
 - `TxInputListBuilder` to `InputSelector`.

 9. Enforce required number of confirmations.

Motivation: required by RPC endpoint specification. You want to be able to avoid redoing bookkeeping (in tems of allocating inputs to transactions) when a reorganization happens and you need to re-initiate.

Struct `InputCandidate` is extended with the meta-data, the number of confirmations specifically, necessary to perform this filter. The `InputSelector` now possesses the required information and applies the filter.


Sword-Smith:
`make happy` reports a few cosmetic problems. 

Suggested unit test, for testing the "confirmations" parameter to the balance getter:
- have devnet wallet
- receive and send transactions over 7 blocks
- Verify tip balance after each block
- After final tip, verify balance matches previous tips when "confirmations" parameter is set to another number than 1.

PR #876, opened 2026-03-06
Ta/exchange integration

Sword-Smith:


neptune-core/src/application/json_rpc/server/rpc.rs
```diff
@@ -11,6 +12,15 @@ use crate::state::GlobalStateLock;
 pub struct RpcServer {
     pub(crate) state: GlobalStateLock,
     pub(crate) to_main_tx: mpsc::Sender<RPCServerToMain>,
+
+    /// If set, means that 3rd parties should not have access to the RPC server,
+    /// as they can do harm.
```
aszepieniec: What are 3rd parties exactly?
 - Another process run on the same machine by the same user?
 - Different user?
 - A trusted machine on the same LAN?
 - Untrusted machine on the same LAN?

Does this shoulding have any causal effect, or is it just a note of warning to be kept in the back of the mind of the programmer / devops guy?

aszepieniec: Here is a suggestion.

```
Whether to allow otherwise-restricted commands. With this boolean set to true, the querier can:
 - make queries that induce large workloads;
 - effect changes to network topology;
 - (assuming the "Personal" namespace is also enabled) observe and spend balance.
```

neptune-core/src/application/json_rpc/server/rpc.rs
```diff
@@ -11,6 +12,15 @@ use crate::state::GlobalStateLock;
...
+    ///
+    /// DOS protection: If set to true, the client may make requests that
+    /// require a lot of work or a long time to answer.
+    ///
+    /// Own wallet state protection: If set to true, access to the node's own
+    /// wallet is permitted through the RPC.
```
aszepieniec:
```suggestion
    /// Own wallet state protection: If set to true, and if the `Personal` namespace
    /// is active, access to the node's wallet is permitted through the RPC.
```

neptune-core/src/application/loops/channel.rs
```diff
@@ -119,7 +123,11 @@ pub(crate) enum RPCServerToMain {
     SetTipToStoredBlock(Digest),
 
     // Used by JSON-RPC
-    SubmitTx(Box<Transaction>),
+    BroadcastThirdPartyTx(Box<Transaction>),
+
+    /// Request that the wallet registers a UTXO claim.
```
aszepieniec, line 128: Redemption claim or off-chain notification claim?

PR #894, opened 2026-03-25
Out-of-Process Tests and Feature Flag "spec"

aszepieniec:
This PR expands the coverage of the test (now tests, plural) in `external_proving_process.rs`. The test that previously merely computed a Fibonacci number now reads from all three sources of non-determinism first. A new test reads two input symbols from standard input. Another new test invokes the out-of-process prover on an actual `KernelToOutputs` computation. In all cases the correct verification of the resulting proof is asserted.

The last case, the `KernelToOutputs` test is a simple but adequate excerpt from the transaction initiation pipeline. Not too complicated but indicative of which things the out-of-process prover ought to support. In order to invoke and test this Triton program, the `TritonProgramSpecification` must be exposed to the integration test, which means that it cannot live behind a test flag alone as it did up until now. To get around this obstacle, this PR introduces a new feature flag "spec". Now, `TritonProgramSpecification` is declared if either the test flag is on, or if compilation happens with feature flag `spec`.

### DIFF
--- a/neptune-core/Cargo.toml
+++ b/neptune-core/Cargo.toml
@@ -165,6 +165,10 @@ libp2p = { version = "0.56.0", features = [
 whoami = "2.0.0"
 
 [dev-dependencies]
+# Integration test `can_prove_out_of_process_kernel_to_outputs` needs both
+# arbitrary implementations and specfications. The next line says, for
+# integration tests we want to rely on whatever is public in crate neptune cash
+# *when feature flags "arbitrary-impls" and "spec" are enabled*.
 neptune-cash = { path = "../neptune-core", features = ["arbitrary-impls", "spec"] }
 
 # note: arbitrary, proptest, proptest-arbitrary-interop are duplicated in [dev-dependencies]
--- a/neptune-core/Cargo.toml
+++ b/neptune-core/Cargo.toml
@@ -62,6 +62,9 @@ tokio-console = ["dep:console-subscriber"]
 # the client side.
 mock-rpc = []
 
+# defines and implements TritonProgramSpecification for various programs
+spec = []
+
 [dependencies]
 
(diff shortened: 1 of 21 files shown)


neptune-core/src/protocol/consensus/transaction/validity/proof_collection.rs
```diff
@@ -307,7 +307,7 @@ impl ProofCollection {
                 [self.salted_inputs_hash, self.salted_outputs_hash]
                     .into_iter()
                     .flat_map(|d| d.reversed().values())
-                    .collect_vec(),
+                    .collect::<Vec<BFieldElement>>(),
```
Sword-Smith: I think you can use `collect::<Vec<_>>()` here for shorter expression.

neptune-core/src/protocol/proof_abstractions/tasm/program.rs
```diff
@@ -354,6 +288,82 @@ pub mod tests {
...
+                ..Default::default()
+            };
+            Self {
+                job_settings,
+                ..Default::default()
+            }
+        }
+    }
+
+    impl From<(TritonVmJobPriority, Option<u8>)> for TritonVmProofJobOptions {
```
Sword-Smith, line 351: These conversion implementations are terrible. But at least they're hidden behind the `test` flag, and also they don't come from this PR. Just wanted to document my discontent.

neptune-core/Cargo.toml
```diff
@@ -162,6 +165,7 @@ libp2p = { version = "0.56.0", features = [
 whoami = "2.0.0"
 
 [dev-dependencies]
+neptune-cash = { path = "../neptune-core", features = ["arbitrary-impls", "spec"] }
```
Sword-Smith, line 172: This looks a little funky. A comment justifying it would be in order.

Sword-Smith (changes requested):

Issue #816, opened 2025-12-13
Change `assert` to `debug_assert` -- It'll panic anyway


skaunov:
The title is by Thor. I was asking in it "why does it panic instead of returning an `Option`?"

https://github.com/Neptune-Crypto/neptune-core/blob/313932dff4bdab12686f4a533434469368a9c7a8/neptune-core/src/application/database/storage/storage_schema/dbtvec_private.rs#L50

aszepieniec:
This should be `debug_assert!` instead of `assert!`. We do not actually want to spend cycles on this assert in production. However, we do want to spend those cycles in tests so we can return useful information to the user in case their logic is wrong.

skaunov:
I bet I created the issue when I saw `assert` which in my mind is not a way to deal with this (thank you for expressing the logic again), and I did not look further much. Reading your fresh analysis/note I noticed that [it actually switch `Option` which is already there for `panic`](https://github.com/Neptune-Crypto/neptune-core/blob/313932dff4bdab12686f4a533434469368a9c7a8/neptune-core/src/application/database/storage/storage_schema/dbtvec_private.rs#L68) which is further eye-popping to me. I mean I know very little about the DB intricacies solved here and next to nothing about Level.

But my first instinct is to rid of both `assert` and `panic` and basically handover `Option` from the referenced line.

PR #898, opened 2026-04-01
ci: Ensure panic without debug assertions

Sword-Smith:
A number of tests must panic, also when debug assertions are not activated. This should act as a tripwire test against the bad logic I reference here:
https://github.com/Neptune-Crypto/neptune-core/issues/816#issuecomment-4170429369

Please don't bother considering this before we have a result from CI.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,7 +62,6 @@ jobs:
       - name: Check Benches
         run: cargo check --benches
 
-      # Nextest uses a filter expression (-E) for skipping tests.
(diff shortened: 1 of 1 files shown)


Sword-Smith:
This new run of tests takes up to 30 minutes, at least without caching. That's up to 50 % of the total time CI workflow time. Maybe it's too high a price to pay? If we instead run these tests through separated crates, by placing them in `neptune-core/tests/` instead, then the price would probably be a lot lower. Try that!

Issue #861, opened 2026-02-19
trimmed/specialized binaries?


skaunov:
I see compilation time concerns, and remember couple of thought on crates separation (#775, #843). After reflecting on the verification function which the user will need for/in #799 I realized that I want to offer another angle to view the separation.

I want to question why does so many things are in `neptune-core` which different users/use-cases might never need? I guess the answer is historical as all of that grew organically. What do you think about separating the things from usage stand-point?

very roughly/illustrational
- an explorer: blocks DB
- an user: the wallet
- a miner: Triton programs

The nice thing is the code is already structured this way. So it's more not to overwhelm/confuse with the whole thing those who want to make/integrate an explorer or wallet.

Correct me if I'm wrong but this won't need even to separate for example networking, as the unused things from a dependency doesn't affect the binary, only takes some space & bandwidth on a dev machine. So maybe having everything in one crate (or one crate per a maintainer) is actually good  --- and all that needed is just small restructuring & development of the code to produce the binaries of different thickness/flavors from the same crate which now yields only a single "size".

PR #903, opened 2026-04-09
feat: add first version of the prover of sent UTXO

skaunov:
this version is limited to native coin without the time-lock

it's the first Non-consensus Triton program in the system available to an user

it was already [discussed in depth since I did not know with which part to start (the other part is the reserves proving)](https://github.com/Neptune-Crypto/neptune-core/pull/799)


aszepieniec:
I wonder if there's a integration test that also verifies a successfully generated proof-of-transfer? I don't see it but I might be missing something.

skaunov:
@aszepieniec 
An example, pls? Each time I get to this topic I receive no answer. :upside_down_face: https://github.com/Neptune-Crypto/neptune-core/pull/799#issuecomment-4114747401

aszepieniec:
Have a look at `neptune-core/tests/send_and_receive_basic.rs`. I guess it doesn't cover the RPC layer but it does cover everything underneath that has to do with sending and receiving.

aszepieniec (changes requested):
I should like to try to send you a proof-of-transfer of a payment that never happened.
 - If I fail, then that helps to convince me that the construction is sound.
 - If I succeed, then that ought to convince you that the construction is not sound.

However, as far as I can see, all changes are in `neptune-core/`. I don't see how you would process that proof-of-transfer, even if I managed to generate one. I would have expected a `neptune-cli` command, or a new option in `neptune-dashboard`. Just exposing a new RPC interface is not enough. It doesn't make any sense to have proofs-of-anything if no-one is capable of verifying them.

skaunov:
@aszepieniec

Let me add this there. **Is there any way to see sent txs?** I have not found last time I checked. An answer would help to make this smoother, as without that I could offer only what will come to my snall mind.

P.s. Probably it's because I struggle a bit to wrap my head around those starting from why it's a different crate instead of maybe a feature of the core. I guess I was trying to get it at #861 .

aszepieniec:
> Is there any way to see sent txs?

Not presently, no.

I guess this is the down side of a privacy preserving ledger. There is no link between the individual user-initiated transactions and the aggregate transactions that appear in blocks on the blockchain. So a minimalistic client that deals just with blockchain data doesn't know what to show the user.

I recall articulating a desire once for a `UiTransaction` which has no mechanical effect except to track transactions initiated by the user and display them properly in the user interface. However, I can't find that term mentioned in the issues.

skaunov (commented):
Added both commands (`prove-an-transfer`, `verify-transfer`, sorry for the names - just change any after merging). I started to document it too but stopped myself: probably something will change again, and you will be reviewing the code anyway which seems to be self-explaining. And I'm glad to provide docs/explanations if/where those would help. In short proving command takes optional args and without these it just proves all the coins (except change) for last sent via the tip of the node.

Let's try this out like you said, pls!

PR #904, opened 2026-04-09
Regtest Mine Loop Fixes

aszepieniec:


neptune-core/src/application/loops/mine_loop.rs
```diff
@@ -80,6 +81,43 @@ pub(crate) async fn compose_block_helper(
...
+            latest_block,
+            composer_parameters.clone(),
+            guesser_address,
+            coinbase_timestamp,
+            rand::random(),
+            txs,
+            global_state_lock.cli().network,
+        );
+        return Ok((block, composer_parameters.extract_expected_utxos(composer_txos)));
+    }
```
Sword-Smith: I'd prefer not to inline this. Can you factor this out?

This function `compose_block_helper` is part of a vital codepath, that of composing real blocks, and I don't want it contaminated with many lines of test-related code. This is better:

```rust
if global_state_lock.cli().network.use_mock_proof() {
    return new_name_function().await;
}
```

Sword-Smith (changes requested):

Sword-Smith (approved):

PR #905, opened 2026-04-11
fix lockscript proving paths and faster sync bitmask canonize

mooncitydev:
## summary

- **lockscript + \TransactionProofType\**: \TxProvingCapability::LockScript\ used \unimplemented!()\ in \From\ for \TransactionProofType\, so building \ProverJobSettings\ from CLI (e.g. \--tx-proving-capability=lockscript\) could panic at startup. It now maps to \PrimitiveWitness\, which matches how \can_prove\ treats capability vs proof type and matches the absence of a separate on-chain lock-script-only proof variant.

- **proof upgrader**: \LockScript\ hit \	odo!()\ in \UpgradeJob::from_primitive_witness\, which could panic when broadcasting a primitive-witness-backed tx. Behavior now mirrors \PrimitiveWitness\: mock networks upgrade toward single-proof; otherwise a clear panic message tells the user to use \proofcollection\ or \singleproof\.

- **\SynchronizationBitMask::canonize\**: replaced the one-bit-at-a-time loop with batched trailing-ones steps so long runs of set bits do not scan linearly (this was marked as very slow in a TODO).

## testing

Rust toolchain was not available in this environment; please run \cargo test -p neptune-cash\ before merge.

made by mooncitydev

### DIFF
--- a/neptune-core/src/application/loops/main_loop/proof_upgrader.rs
+++ b/neptune-core/src/application/loops/main_loop/proof_upgrader.rs
@@ -255,7 +255,17 @@ impl UpgradeJob {
             TxProvingCapability::PrimitiveWitness => {
                 panic!("Client cannot have primitive witness capability only")
             }
-            TxProvingCapability::LockScript => todo!("TODO: Add support for this"),
+            TxProvingCapability::LockScript if network.use_mock_proof() => {
+                UpgradeJob::PrimitiveWitnessToSingleProof(PrimitiveWitnessToSingleProof {
+                    primitive_witness,
+                })
+            }
+            TxProvingCapability::LockScript => {
+                panic!(
+                    "Client cannot upgrade primitive witness with lock-script-only proving capability; use \
+                     --tx-proving-capability=proofcollection or singleproof to broadcast transactions"
+                )
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


aszepieniec:
The text does not render properly. That makes it difficult to read (for humans).

mooncitydev:
## Summary

### LockScript + `TransactionProofType`

`TxProvingCapability::LockScript` previously used `unimplemented!()` inside `From<TransactionProofType>`, which could cause a panic during startup when building `ProverJobSettings` from CLI arguments such as:

`--tx-proving-capability=lockscript`

This has now been fixed by mapping it to `PrimitiveWitness`.

The new behavior is consistent with:

* how `can_prove` handles capability vs proof type
* the fact that there is no dedicated on-chain “lock-script-only” proof variant

---

### Proof Upgrader

`LockScript` previously hit a `todo!()` inside `UpgradeJob::from_primitive_witness`, which could panic while broadcasting a transaction backed by a primitive witness.

Behavior now matches `PrimitiveWitness` handling:

* mock networks automatically upgrade toward single-proof
* otherwise, a clear panic message instructs the user to use:

  * `proofcollection`
  * or `singleproof`

---

### `SynchronizationBitMask::canonize`

Optimized the implementation by replacing the previous one-bit-at-a-time loop with batched trailing-ones processing.

This avoids linear scanning for long runs of set bits and resolves the existing TODO that marked the implementation as very slow.

---

## Testing

Rust toolchain was not available in this environment.

Please run:

`cargo test -p neptune-cash`

before merging.

---

Made by mooncitydev

PR #916, opened 2026-05-08
Thv/private addresses

Sword-Smith:
Missing tests:

- [x] loop over all key/address types, send to an address of this format, and verify that this new UTXO can be spent and included in a block. Also verify that balances are updated this way.
- [x] Verify that `PrivateAddressKey` can unlock `PrivateAddress`'s lock script. Preferably by going through TVM.
- [x] Test with new key type in existing tests.

Missing:
- [x] documentation of new address type
- [x] rename to secret address per @aszepieniec's request.

More:
- [x] Add address format originally proposed by @KaffinPX: Secret addresses where anyone who knows the address can read the entire (onchain-announced) transaction history.


Sword-Smith:
@aszepieniec please review cryptography part. Especially `SecretAddress::encrypt` in `neptune-core/src/state/wallet/address/secret_address.rs`. The rest is pretty much just boiler plate and book keeping code.

aszepieniec (commented):
About the name ... "secret address" is a small step up from "private address" but both are terrible.

What's the point of calling them "secret addresses" if they can be published and retain security assuming quantum computers don't come? The need to keep them *secret* arises from the assumption that quantum computers are either already there or will be in the relevant future. However, under this assumption, there is no added benefit to the layer of quantumly-insecure encryption. So this name reflects assumptions that make its features obsolete.

So for starters, I would suggest a name that reflects the mechanics and leave judgment about the risk of the quantum computer to the user. In this line, try "pre-quantum" or "ECDH".

That said, I think I want to dwell a bit on this topic because I think there more is at play than mere semiotics.

As far as I can tell, this address format is perfectly secure for broadcasting in a world inhabited only by classical attackers. In a world with quantum attackers, it *still* serves a use case, namely as a medium for the sender to learn the receiver digest and ID, and for both sender and receiver to produce the same symmetric key; but with the caveat that the attacker who learns the address can decrypt corresponding ciphertexts and eventually deanonymize payments made to it. With respect to this use case, this address format is suboptimal in terms of length and operational cost -- meaning, the same can be achieved with a shorter and cheaper address format. If/when quantum computers come, we probably want to switch to this more efficient format for covering that use case.

The use case that makes the address format of this PR the optimal choice is when users do not care about quantum readiness and just want the shortest possible addresses and (on-chain) announcements.

So there are two use cases.

 1. not-for-sharing
 2. publishable

I guess the point I am converging towards is that you cannot simultaneously optimize both targets.

The address format of this PR represents a trade-off on both dimensions. It is not the best in the "publishable" category because of the quantum risk; and it is not the best in the "not-for-sharing" category either because of the superfluous public key encryption layer.

I realize that right now this optimal "not-for-sharing" address format does not exist, but getting there is not an infeasible engineering task. Moreover, address formats supported once must be supported forever, so it pays to be careful and deliberate. Therefore, let's consider the merit of this PR assuming that that optimal "not-for-sharing" alternative does already exist -- would this PR still be worthwhile?

Well, *maybe*, but only on the basis of a motivation that sets it apart from alternatives, which is that it is *the shortest publishable address if you don't care about quantum-readiness*.

The reason why that maybe is a maybe is not technical but more related to marketing than anything else. Supporting this address format (with the correct motivation!) means watering down with context and fine text one of the main selling points of Neptune Cash, which is that promises to resist quantum attacks, period.

Let me try to turn all that into something constructive.

The optimal "not-for-sharing" address should be on the table as an option -- even if we decide to discard it later on, it pays to have something concrete to talk about. At that stage I see four options:
 1. Keep the optimal "not-for-sharing" address format; drop the ECDH address format.
 2. Keep both.
 3. Drop the optimal "not-for-sharing" address format; keep the ECDH address format.
 4. Drop both.

Right now my preference aligns with exactly that order, but I welcome counter-arguments.

Suggested next steps:
 - Rename "secret address" to "ECDH address"
 - Implement optimal "not-for-sharing" addresses (these may properly be called secret addresses)
 - Resume discussion

docs/src/neptune-core/keys.md
```diff
@@ -55,6 +53,26 @@ correct, it would be safe to put funds in a paper or metal wallet and ignore
 them for decades, perhaps until they are transferred to the original owner's
 children or grand-children.
 
+### `Secret` keys and addresses
+
+Secret keys are implemented with aes-256-gcm and EC Diffie-Hellman key exchange.
+
+Like `Generation` addresses, `Secret` addresses are post-quantum secure but only when the addresses
+are kept secret. In other words, an attacker that possesses a quantum computer *and* knows the address
```
aszepieniec: "post-quantum secure but only when the addresses are kept secret" does not describe `Generation` addresses.

Sword-Smith:
Suggested way forward:
- I'll rename this address format to "elliptic curve hybrid", "ech" for short.
- I'll add the "secret" address format per @KaffinPX's original suggestion: The address contains the AES key directly, meaning that everyone who gets the address can see the entire transaction history of it, without a quantum computer. This address format will be called "secret", maybe "sec" for short.

Sword-Smith:
OK. Both address formats are added now:
- elliptic curve hybrid: Requires that attacker knows address *and* has quantum computer in order to decrypt onchain announcements.
- secret: Only requires that attacker knows the address in order to decrypt onchain announcements.

For both address format, funds are safe from theft even if attacker possesses quantum computer and knowledge of address.

## Size analysis

| Address Type | Address Length (Bech32 characters) | Ciphertext Length (b-field elements) |
| --- | --- | --- |
| EC hybrid address | 168 | 19 |
| Secret address | 115 | 15 |
| Generation address | 3482 | 335 |

**EC hybrid address**
```
nechm1qtmcrts70kl4hzjej5te74pk67aja8wmz6aawx995c0cnzl2sxuxdv37p70sgxl7z4dqnmrvqcy2lgsy3xu8gc6xpgaps8x6ppxulpzqj8vqnm5p3n4cyjrzpr2pjr6zwn49x0ejs2p3vyausep5pcftn2dq200dz7
```

**Secret address**
```
nsecm10243hqsugd3nvmck0f3f49y75dku9lq4ym37lxx6vzl0p2l4kvthjcyzkkpy6f9v9gh8w6hxhaa4fexl4zhnh2qq7ulyrdmttjwukzs3nwpls
```

docs/src/neptune-core/keys.md
```diff
@@ -1,6 +1,6 @@
 # Keys and Addresses
 
-`neptune-core` uses an extensible system of keys and addresses.  This is accomplished via an abstract type for each.  At present two types of keys are supported: `Generation` and `Symmetric`.
+`neptune-core` uses an extensible system of keys and addresses.  This is accomplished via an abstract type for each.  At present three types of keys are supported: `Generation`, `EcHybrid`, and `Symmetric`.
```
aszepieniec: + Secret

docs/src/neptune-core/keys.md
```diff
@@ -55,6 +51,28 @@ correct, it would be safe to put funds in a paper or metal wallet and ignore
 them for decades, perhaps until they are transferred to the original owner's
 children or grand-children.
 
+### `EcHybrid` keys and addresses
+
+Elliptic curve hybrid keys are implemented with aes-256-gcm and EC Diffie-Hellman key exchange.
+
+Like `Generation` addresses, EC hybrid addresses are post-quantum secure if used correctly. However,
```
aszepieniec, line 84: > Like `Generation` addresses, EC hybrid addresses are post-quantum secure if used correctly.

This sentence is true only under a non-standard definition of security and for the correct value of "correct use". Both are difficult to glean from the subsequent text.

Suggestion: define the features up front, and then say which ones ECH addresses offer.

 - Theft-prevention: the adversary who sees the address cannot spend funds spent to it.
 - Privacy (address-destination unlinkability): the adversary cannot link addresses to payments to them.
 - Privacy (confidentiality): the adversary who sees an address and a payment to that address, cannot determine the type or amount of funds transferred.
 - Privacy (input-output unlinkability): the adversary who sees the address and a payment to that address, cannot determine whether the transferred funds were spent again in the time since.
 
We can obviously continue defining variants of privacy but I don't think any of those variants discriminate against ECH addresses.

What I think you want to say is that ECH addresses offer post-quantum theft-prevention, but confidentiality only against classical adversaries.

docs/src/neptune-core/keys.md
```diff
@@ -55,6 +51,28 @@ correct, it would be safe to put funds in a paper or metal wallet and ignore
...
+unlike `Generation` addresses they stop being post-quantum secure if the address is published, or
+shared between more parties than merely the sender and the receiver. In other words, an attacker that
+possesses a quantum computer *and* knows an `EcHybrid` address can read the UTXO notifications for
+that specific specific address and thus decrypt all amounts and UTXOs that were announced on-chain
+for this address. This attacker still cannot steal funds from the address.
+
+`EcHybrid` addresses are intended to be seen by two parties: the sender and the receiver. If it is
```
aszepieniec:
```suggestion
`EcHybrid` addresses are intended to be seen by only two parties: the sender and the receiver. If it is
```

neptune-core/src/state/wallet/address/addressable_key.rs
```diff
@@ -31,13 +43,29 @@ pub enum KeyType {
 
     /// [symmetric_key] built on aes-256-gcm
     Symmetric = symmetric_key::SYMMETRIC_KEY_FLAG_U8,
+
+    /// Elliptic curve hybrid address. Only post-quantum secure if address is
+    /// kept private between sender and receiver.
```
aszepieniec: The sentence "Only post-quantum secure if address is kept private between sender and receiver." invites more questions than it answers. Drop or replace with something that assumes less knowledge on the part of the reader.

neptune-core/src/state/wallet/address/addressable_key.rs
```diff
@@ -31,13 +43,29 @@ pub enum KeyType {
...
+    ///
+    /// If an attacker has a quantum computer *and* knows the address, they can
+    /// see the entire transaction history of that address.
+    EcHybrid = elliptic_curve_hybrid::ELLIPTIC_CURVE_HYBRID_ADDRESS_FLAG_U8,
+
+    /// An address format that leaks transaction information to anyone who
+    /// knows the address, if on-chain announcements are used.
```
aszepieniec, line 57: This text describes a consequence of the definition of a Secret Address, and is not the definition itself. That makes it rather generic, as it also captures other address formats that share this property but are mechanically different.

To capture the definition, I would suggest something like "the notification encryption key is derived from the address through symmetric crypto only".

neptune-core/src/state/wallet/address/viewing_address.rs
```diff
@@ -0,0 +1,502 @@
...
+        let digest = Tip5::hash_pair(
+            self.receiver_digest,
+            Digest([e0, e1, e2, SECRET_ADDRESS_FLAG, bfe!(0)]),
+        );
+        let digest: [u8; 40] = digest.into();
+
+        digest.into_iter().take(32).collect_array().unwrap()
+    }
+
+    /// Manually serialize to exactly 64 bytes to avoid bincode overhead.
```
aszepieniec, line 154: Huh, I thought bincode had zero overhead.

Anyhow, I might be wrong or right but either way there is confusion about exactly what bincode does under the hood, which is why this function is good to have regardless.

aszepieniec (approved):
I have reviewed the cryptography and find it adequate.

I do have one technical objection, which is the definition of the AES key for the elliptic curve hybrid address in terms of XOR of a receiver share and a sender share. While it does not undermine security (at least, not that I can see), there is no reason for this convolution. The motivation for this XOR is to ensure that the quantum attacker must also possess the address in order to launch the attack. However, the same is true for the obvious and simpler alternative, where $K_{\sf AES} = \mathsf{H}(g^{ab})$.

In either case, the quantum attacker can obtain $b$ from $g^b$ but then needs to iterate over all known addresses to find a matching $\mathsf{H}(g^ab)$. In the current scheme, the attack is more complicated but only by a constant number of steps per iteration.

Non-technical comment: the documentation can use some work. Consider using an LLM.

PR #920, opened 2026-05-18
fix(startup): recover gracefully when AMS sync label disagrees with tip

mastercyb:
## Problem

On startup, `restore_monitored_utxos_from_recovery_data` asserts that the archival mutator set sync label equals the stored chain tip:

```rust
assert_eq!(tip_hash, asm_sync_label, "Error: sync label in archival mutator set database disagrees with block tip...");
```

If a previous run crashed mid-`update_mutator_set` — for example due to SIGKILL, OOM, or power loss during a long sync session — the AMS lands in a valid-but-stale state: it is internally consistent up to block N, while the stored tip is block N+k. The hard assert then panics on the next startup with no recovery path short of deleting the entire `databases/` directory and re-syncing from scratch.

This is a significant usability problem for long-running sync sessions (15 000+ blocks). Any interruption permanently bricks the node, and re-syncing from scratch currently takes many hours due to the random block download strategy.

The error seen on startup:
```
thread 'main' panicked at neptune-core/src/state/mod.rs:2174:9:
assertion `left == right` failed: Error: sync label in archival mutator set database disagrees with block tip.
Possible solution: try deleting the database at `DATA_DIR/databases/`.
```

## Fix

Convert the hard `assert_eq!` into a `warn!` + early return:

```rust
if tip_hash != asm_sync_label {
    warn!(
        "AMS sync label disagrees with tip (ams={asm_sync_label:x} tip={tip_hash:x}). \
        Skipping UTXO recovery; AMS will self-heal as new blocks arrive."
    );
    return Ok(());
}
```

**Why this is safe:**
- The AMS at the sync label height is internally correct — only the path from sync-label to tip is missing
- `update_mutator_set`'s `find_path` already handles non-trivial rewalks (the backwards/forwards path logic). When the next `set_new_tip` is called for a new block, it will correctly walk the AMS forward from the sync label to the new block's predecessor, catching up automatically
- UTXO recovery is skipped in the interim, which is acceptable: the wallet has its own recovery data (`read_utxo_ms_recovery_data`) that is independent of the AMS state, and any missing membership proofs will be reconstructed once the AMS is back in sync

Note: this fix is most useful in combination with the companion PR that makes `update_mutator_set` idempotent, which prevents the AMS from drifting out of sync in the first place.

### DIFF
--- a/neptune-core/src/state/mod.rs
+++ b/neptune-core/src/state/mod.rs
@@ -2171,17 +2171,16 @@ impl GlobalState {
         let ams_ref = &self.chain.archival_state().archival_mutator_set;
 
         let asm_sync_label = ams_ref.get_sync_label();
-        assert_eq!(
-            tip_hash, asm_sync_label,
-            "Error: sync label in archival mutator set database disagrees with \
-            block tip. Archival mutator set must be synced to tip for successful \
-            MUTXO recovery.\n\
-            Possible causes: different or new genesis block; corrupt file system.\n\
-            Possible solution: try deleting the database at `DATA_DIR/databases/`. \
-            Get the value of `DATA_DIR` from the first message in the log, and \
-            *do not* delete the wallet file or directory.\n\n\
-            Tip:\n{tip_hash:x};\nsync label:\n{asm_sync_label:x}"
-        );
+        if tip_hash != asm_sync_label {
+            // AMS is behind tip — likely a crash during update_mutator_set.
+            // Skip UTXO recovery; AMS will catch up via update_mutator_set on
+            // the next set_new_tip call (find_path handles non-trivial rewalk).
+            warn!(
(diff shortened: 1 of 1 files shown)


Sword-Smith:
This makes a lot of sense, and I think it's correct. CI also gives you the thumbs up (apart from a flaky test that's not your fault). I was wondering though if it would be possible to add a test that this achieves what you hope it achieves? Or maybe it would make sense to just update the state with latest `tip` block if this mutator set discrepancy is discovered?

Sword-Smith:
Thinking about this a bit more, I'm a bit uncomfortable proceeding with the node booting up when the archival mutator set is not synced to the tip. So if this check fails at startup, would there maybe be a way to fix it immediately?

Sword-Smith:
Superseded by #924 

Instead of accepting a broken archival mutator set state, we prefer to fix it at startup.

PR #922, opened 2026-05-18
fix(sync): start successors subtask eagerly on FastForward; tolerate legacy block format

mastercyb:
### Problem

When neptune-core resumes a previously interrupted sync session, the coverage bitmask is pre-populated with blocks downloaded before the interruption. The `FastForward` handler advances the tip and updates the bitmask floor, but the successors subtask is only spawned when a *new* block arrives at exactly `tip + 1`.

If that block was already downloaded in the prior session, no peer will send it again. The successors subtask is never spawned, and the sync loop stalls indefinitely — it has all the data it needs but no code path to act on it.

This makes interrupted sync sessions unrecoverable without deleting the temp directory and starting over.

### Fix

After updating the tip in the `FastForward` handler, check whether the next height (`tip + 1`) is already present in the coverage bitmask. If it is, and no successors subtask is currently running, spawn one immediately:

```rust
if maybe_successors_subtask.is_none()
    && self.download_state.coverage().contains(
        self.tip.header().height.next().value(),
    )
{
    // spawn process_successors_of_tip immediately
}
```

This restores the invariant that a successors subtask is always running whenever there is contiguous covered work ready to commit, even across session boundaries.

---

### Problem

`mutator_set_update()` in `block_kernel.rs` calls `RemovalRecordList::try_unpack` on the raw inputs field. For blocks minted before a serialization format change (roughly heights < 20 000 on mainnet), this call returns `Err`, causing `mutator_set_update` to `bail!` and preventing those blocks from being applied to the archival mutator set.

```rust
let inputs = RemovalRecordList::try_unpack(self.body.transaction_kernel.inputs.clone())
    .map_err(BlockValidationError::from)?;  // ← bails for old-format blocks
```

The result: AMS application fails silently for older blocks, leaving the AMS desynchronized from the chain.

### Fix

Fall back to the raw field when `try_unpack` fails:

```rust
let inputs = RemovalRecordList::try_unpack(self.body.transaction_kernel.inputs.clone())
    .unwrap_or_else(|_| self.body.transaction_kernel.inputs.clone());
```

The raw field contains the correct removal records in the older encoding, so the resulting `MutatorSetUpdate` is accurate.

### DIFF
--- a/neptune-core/src/application/loops/sync_loop.rs
+++ b/neptune-core/src/application/loops/sync_loop.rs
@@ -374,6 +374,40 @@ impl SyncLoop {
                             if new_tip.header().height > self.tip.header().height {
                                 self.download_state.fast_forward(new_tip.header().height);
                                 self.tip = *new_tip;
+
+                                // If the next block after the new tip is already
+                                // in the coverage bitmask (e.g., recovered from a
+                                // previous sync session), start the subtask now.
+                                // Without this, the subtask would only start when
+                                // a peer sends a block at exactly tip+1, which
+                                // never happens for already-covered heights.
+                                if maybe_successors_subtask.is_none()
+                                    && self.download_state.coverage().contains(
+                                        self.tip.header().height.next().value(),
+                                    )
+                                {
+                                    tracing::info!(
+                                        "Sync loop: next block {} is already in coverage; \
+                                        starting successors subtask immediately.",
+                                        self.tip.header().height.next()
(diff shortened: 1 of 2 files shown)


Sword-Smith:
I really don't think the serialization format changed at block height 20,000. It changed as a result of the (balance preserving) reboot. So since it changed at the reboot, none of the old blocks (without input packing) are seen by active nodes today.

PR #931, opened 2026-05-29
test: harden flaky update_mutator_set integration tests

### LINKED ISSUE #928: Flaky tests
`update_mutator_set_on_own_transaction_two_new_blocks_from_peer`
```
stderr ───

    thread 'update_mutator_set_on_own_transaction_two_new_blocks_from_peer' (23454) panicked at neptune-core/tests/update_mutator_set_own_tx.rs:121:18:
    called `Result::unwrap()` on an `Err` value: Failed("block not in blockchain after 5 seconds")
```

`update_mutator_set_on_own_transaction_one_new_block_self_mined`
```
  stderr ───

    thread 'update_mutator_set_on_own_transaction_one_new_block_self_mined' (5720) panicked at neptune-core\tests\update_mutator_set_own_tx.rs:147:10:
    called `Result::unwrap()` on an `Err` value: tx not confirmable after 30 seconds
```

zeokin:
Reduces flakiness of the two tests tracked in #928 by removing their deterministic and test-infrastructure failure modes:
fix #928 
- Fix the "block not in blockchain" error string, which used a non-interpolated `{timeout_in_secs}` literal.
- `wait_until_tx_in_mempool_confirmable` no longer panics when the tx is evicted by a failed mutator-set update; it reports a clear error on timeout instead.
- Cluster nodes now get random base ports instead of ports derived deterministically from the call site, which collided across repeated/concurrent runs.
- Add a regression test for the cluster port invariants.

`cargo fmt`, `cargo clippy -- -D warnings`, and both tests (serial and parallel) pass locally.

### DIFF
--- a/neptune-core/tests/update_mutator_set_own_tx.rs
+++ b/neptune-core/tests/update_mutator_set_own_tx.rs
@@ -113,6 +113,11 @@ async fn worker(mut alice: GenesisNode, new_block_source: SourceOfNewBlocks) {
             num_new_blocks,
             peer: mut bob,
         } => {
+            // Bob must have synced Alice's two blocks before he starts mining,
+            // so that he composes on top of the shared tip.
+            bob.wait_until_block_height(2u64, timeout_secs)
+                .await
+                .unwrap();
             bob.gsl
                 .api_mut()
                 .regtest_mut()
--- a/neptune-core/src/api/regtest/regtest_impl.rs
+++ b/neptune-core/src/api/regtest/regtest_impl.rs
@@ -256,7 +256,7 @@ impl RegTestPrivate {
         state.net.freeze = false;
     }
 
-    // waits (polls) until block is found in canonical chain or 5 second timeout occurs.
+    // waits (polls) until block is found in canonical chain or a timeout occurs.
     //
     // note: temporary until listener approach is implemented.
     async fn wait_until_block_in_chain(
@@ -278,9 +278,9 @@ impl RegTestPrivate {
                 {
                     return Ok(());
                 }
-                return Err(RegTestError::Failed(
-                    "block not in blockchain after {timeout_in_secs} seconds".into(),
-                ));
+                return Err(RegTestError::Failed(format!(
+                    "block not in blockchain after {timeout_in_secs} seconds"
+                )));
             }
             tokio::time::sleep(std::time::Duration::from_millis(10)).await;
         }


zeokin:
Hi @Sword-Smith please review this PR when you're free.

neptune-core/tests/common/genesis_node.rs
```diff
@@ -174,24 +175,36 @@ impl GenesisNode {
...
+        // node `i` gets peer port `base + i` and rpc port `base + num_nodes + i`.
+        let n = u16::from(num_nodes);
+        let base_port = rand::rng().random_range((1u16 << 10)..(u16::MAX - 2 * n));
+        base_args.peer_port = base_port;
+        base_args.rpc_port = base_port + n;
+        tracing::debug!(
+            "assigning ports for cluster {cluster_id}: peer base {}, rpc base {}",
+            base_args.peer_port,
+            base_args.rpc_port
+        );
```
Sword-Smith: Doesn't this make it harder to run tests in parallel? With the `hash_string_to_port_range` function, the port assignments are deterministic, they stay the same for every test run, whereas with this random selection of ports it seems to me that the flakyness of testing will increase, as the port selection will overlap maybe 5 % of all test runs.

neptune-core/tests/common/genesis_node.rs
```diff
@@ -392,34 +405,47 @@ impl GenesisNode {
...
-                    .gsl
-                    .lock_guard()
-                    .await
+            // `None` means the tx is (currently) not in the mempool. During a
+            // normal mutator-set update the transaction is updated in place and
+            // stays present, so a missing tx indicates it was evicted (e.g. by
+            // a failed update). We keep polling regardless and only report the
+            // eviction if we time out, so that a transient absence never fails
+            // the wait spuriously (and never panics, as the previous `.expect`
+            // did).
```
Sword-Smith: I don't like this comment because it assumes behavior of the mempool and the transaction-upgrading process. It may or may not be correct (I honestly don't know) but the reason it shouldn't be there is that it assumes internals about other code modules that are far removed from this test.

neptune-core/tests/common/genesis_node.rs
```diff
@@ -392,34 +405,47 @@ impl GenesisNode {
...
+                state
                     .mempool
                     .get(txid)
-                    .expect("Referenced tx must be in mempool")
-                    .is_confirmable_relative_to(mutator_set)
-                {
-                    break;
-                }
+                    .map(|tx| tx.is_confirmable_relative_to(mutator_set))
+            };
```
Sword-Smith: This change actually makes sense to me. And on the surface it looks good, but I don't think I've ever seen this `expect` panic. So are you actually fixing anything through this change? I am skeptical.

neptune-core/tests/update_mutator_set_own_tx.rs
```diff
@@ -231,3 +232,46 @@ pub async fn update_mutator_set_on_own_transaction_two_new_blocks_from_peer() {
...
+/// The two-node `..._from_peer` test starts a cluster, and earlier the cluster
+/// helper derived its ports deterministically from the (fixed) call site, which
+/// made repeated or concurrent runs collide on the same sockets - a source of
+/// flaky failures. The ports are now drawn from a random base. This test pins
+/// the invariants that prevent the collision: every node gets a distinct peer
+/// port and a distinct rpc port, no number is reused as both a peer and an rpc
+/// port (which would make a node fail to bind), and every node is told to
+/// connect to each of the others.
+#[test]
+fn cluster_nodes_get_distinct_non_overlapping_ports() {
```
Sword-Smith: Will this test not just fail once every time the RNG gives two test runs an overlapping port range? It seems to me that the more of these integration tests we add, the more likely one of these tests (that assumes specific ports are free) will fail, and what's worse is that they will do so in a flaky manner.

Sword-Smith (changes requested):
I'm not convinced that this addresses the problem of the flaky tests. This PR achieves three things:
1. Fixes an incorrectly interpolated error message.
2. Makes port distribution RNG-based
3. Allows for the transaction to be dropped from the mempool while we are waiting for the transaction to be updated to a new mutator set.

I'm not sure 2 and 3 actually will fix anything. The errors we are getting from the flaky test failures are not because of ports being occupied by other tests. Instead they are about blocks not being present in the state at the timeout[^1]. I definitely want 1) but the two others I don't see a use for TBH.

[^1]: See for example the latest CI failure: https://github.com/Neptune-Crypto/neptune-core/actions/runs/26633530487/job/78488319172 `Failed("block not in blockchain after {timeout_in_secs} seconds")`.

Sword-Smith (approved):
LGTM. Can you sign your commit, so that we can get this merged? It's a security policy of `neptune-core`. It shouldn't take more than 10 minutes to set up automatic commit signing if you follow a guide online.

zeokin:
> LGTM. Can you sign your commit, so that we can get this merged? It's a security policy of `neptune-core`. It shouldn't take more than 10 minutes to set up automatic commit signing if you follow a guide online.

Hi, I've signed. Thanks for the review.

PR #932, opened 2026-05-31
feat(json_rpc): add archival batch_are_bloom_indices_set endpoint

### LINKED ISSUE #929: New RPC endpoint: archival::batch_are_bloom_indices_set
Allow the user to ask about many absolute indices, in one request. This is done instead of issuing one request per absolute index set.

zeokin:
## Add RPC endpoint: `archival::batch_are_bloom_indices_set`

Closes #929.

Batched version of `are_bloom_indices_set`: checks many absolute index sets in one request instead of one request per set.

- **In:** `Vec<AbsoluteIndexSet>` → **Out:** `Vec<bool>` (one per input, in order)
- Locks state once for the whole batch
- Empty input → empty response

Includes a test covering ordering, duplicates, empty input, and parity with the single-set endpoint.

### DIFF
--- a/neptune-core/src/application/json_rpc/core/api/rpc.rs
+++ b/neptune-core/src/application/json_rpc/core/api/rpc.rs
@@ -89,6 +89,9 @@ pub enum RpcError {
     #[error("Filtering conditions may not be empty")]
     EmptyFilteringConditions,
 
+    #[error("Too many absolute index sets requested: maximum is {max}, got {got}")]
+    TooManyAbsoluteIndexSets { max: usize, got: usize },
+
     // Common case errors
     #[error("Invalid address provided in arguments")]
     InvalidAddress,
@@ -125,6 +128,8 @@ pub enum RpcError {
 }
 
 pub type RpcResult<T> = Result<T, RpcError>;
+
+pub const MAX_BATCH_ARE_BLOOM_INDICES_SET_INDEX_SETS: usize = 1000;
 
 #[async_trait]
 pub trait RpcApi: Sync + Send {
@@ -336,6 +341,29 @@ pub trait RpcApi: Sync + Send {
         &self,
         request: AreBloomIndicesSetRequest,
     ) -> RpcResult<AreBloomIndicesSetResponse>;
+
+    /// Batched form of [`Self::are_bloom_indices_set`]: for each absolute index
+    /// set, check whether all of its indices are set in the node's archival
+    /// mutator set, in a single request instead of one request per index set.
+    ///
+    /// The returned `are_set` vector corresponds element-by-element, and in the
+    /// same order, to the `absolute_index_sets` argument (duplicate inputs are
+    /// preserved). An empty input yields an empty response. At most 1000
+    /// absolute index sets may be requested at once.
+    async fn batch_are_bloom_indices_set(
+        &self,
+        absolute_index_sets: Vec<RpcAbsoluteIndexSet>,
+    ) -> RpcResult<BatchAreBloomIndicesSetResponse> {
+        self.batch_are_bloom_indices_set_call(BatchAreBloomIndicesSetRequest {
+            absolute_index_sets,
+        })
+        .await
+    }
+
+    async fn batch_are_bloom_indices_set_call(
+        &self,
+        request: BatchAreBloomIndicesSetRequest,
+    ) -> RpcResult<BatchAreBloomIndicesSetResponse>;
 
     async fn circulating_supply(&self) -> RpcResult<CirculatingSupplyResponse> {
         self.circulating_supply_call(CirculatingSupplyRequest {})
--- a/neptune-core/src/application/json_rpc/core/model/message.rs
+++ b/neptune-core/src/application/json_rpc/core/model/message.rs
@@ -284,6 +284,20 @@ pub struct AreBloomIndicesSetResponse {
     pub are_set: bool,
 }
 
+#[derive(Clone, Debug, Serialize_tuple, Deserialize_tuple)]
+#[serde(rename_all = "camelCase")]
+pub struct BatchAreBloomIndicesSetRequest {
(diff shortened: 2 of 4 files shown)


zeokin:
Hi @Sword-Smith how are you? Could you review this PR?

neptune-core/src/application/json_rpc/core/model/message.rs
```diff
@@ -284,6 +284,23 @@ pub struct AreBloomIndicesSetResponse {
     pub are_set: bool,
 }
 
+// Note: cannot derive `Copy` (and the response cannot be `Serialize_tuple`)
+// because of the `Vec` fields; mirrors the other `Vec<RpcAbsoluteIndexSet>`
+// batch requests such as `BlockHeightsByAbsoluteIndexSetsRequest`.
```
Sword-Smith: This comment works better as a commit messages instead of a comment.

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -404,6 +404,24 @@ impl RpcApi for RpcServer {
...
+        // index set against the same archival mutator set snapshot.
+        let state = self.state.lock_guard().await;
+        let ams = state.chain.archival_state().archival_mutator_set.ams();
+
+        let mut are_set = Vec::with_capacity(request.absolute_index_sets.len());
+        for absolute_index_set in request.absolute_index_sets {
+            are_set.push(ams.absolute_index_set_was_applied(absolute_index_set).await);
+        }
+
+        Ok(BatchAreBloomIndicesSetResponse { are_set })
```
Sword-Smith, line 430: Looks good! Let's cap the input size to 1000 though. That way, we keep response times down. So: It shouldn't be allowed to request for more than 1000 `AbsoluteIndexSet`s per request. Downstream dependencies will just have to deal with this complication IMO. What do you think?

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -1827,6 +1845,121 @@ pub mod tests {
         );
     }
 
+    /// Regression test for the `archival::batchAreBloomIndicesSet` endpoint
+    /// (issue #929): a batch query returns one result per input set, in the same
+    /// order (duplicates preserved), and agrees with the single-set
+    /// `are_bloom_indices_set` endpoint element-by-element.
```
Sword-Smith: A test of a new feature is not a regression test. Regression tests are for when you fix bugs, not when you add new features :)

neptune-core/src/application/json_rpc/core/api/rpc.rs
```diff
@@ -337,6 +337,28 @@ pub trait RpcApi: Sync + Send {
         request: AreBloomIndicesSetRequest,
     ) -> RpcResult<AreBloomIndicesSetResponse>;
 
+    /// Batched form of [`Self::are_bloom_indices_set`]: for each absolute index
+    /// set, check whether all of its indices are set in the node's archival
+    /// mutator set, in a single request instead of one request per index set.
+    ///
+    /// The returned `are_set` vector corresponds element-by-element, and in the
+    /// same order, to the `absolute_index_sets` argument (duplicate inputs are
+    /// preserved). An empty input yields an empty response.
```
Sword-Smith: Well specified, that duplicates are preserved!

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -1827,6 +1845,121 @@ pub mod tests {
...
+            .get_wallet_status_for_tip()
+            .await;
+        let SyncedUtxo {
+            aocl_leaf_index,
+            utxo,
+            sender_randomness,
+            receiver_preimage,
+            ..
+        } = &wallet_status.synced_unspent(None).next().unwrap();
+        let unapplied_set = AbsoluteIndexSet::compute(
```
Sword-Smith: I prefer the name `not_applied_set` or maybe just `not_applied` here. Unapplied sounds to me like it was first applied and then rolled back.

neptune-core/src/application/json_rpc/server/service.rs
```diff
@@ -1827,6 +1845,121 @@ pub mod tests {
         );
     }
 
+    /// Regression test for the `archival::batchAreBloomIndicesSet` endpoint
+    /// (issue #929): a batch query returns one result per input set, in the same
+    /// order (duplicates preserved), and agrees with the single-set
+    /// `are_bloom_indices_set` endpoint element-by-element.
+    #[apply(shared_tokio_runtime)]
+    async fn batch_are_bloom_indices_set_matches_single_set() {
```
Sword-Smith, line 1863: Overall, an excellent test.

Sword-Smith (changes requested):
Small stylistic changes required. Then I'm happy to merge :)

zeokin:
Hi @Sword-Smith I've fixed. Please review again when you're free.

Sword-Smith (approved):
Very happy to merge this. Good job. And congratulations on your 2nd commit to `neptune-core`!

I have one stylistic observation that we should keep in mind: Let's watch out that we don't grow the codebase too much. The production code is excellent in this PR, but the test might be a little verbose -- it adds a lot of lines of code. In the future, we should try to either make better test helper functions that would make similar tests take up fewer lines of code, or we should focus on testing the underlying logic (in this case `ArchivalMutatorSet::absolute_index_set_was_applied`) thoroughly and then leave the thin RPC layer untested.

To be clear: Don't change anything in this PR. Let's just try to be mindful about lines of codes going forward.

Sword-Smith:
One more thing: This PR should be one commit, not two. I prefer that non-default branches (all branches except `master`) are rebased on top of `master`. In this PR, there is instead an unnecessary merge commit. Can you rebase fix that? If not, I can fix it and then make the commit, and add you as a co-author. Whatever you prefer.

Sword-Smith:
Never mind. I just cherry-picked your commit and pushed it to `master`. So consider this PR merged.

PR #935, opened 2026-06-06
Add an endpoint that returns all the pairs

zeokin:
Add an endpoint that returns all the (block hash, BlockHeader) pairs that could be the origin of the UTXO referenced by the AbsoluteIndexSet


zeokin:
Hi @Sword-Smith I've done. Please check.

neptune-core/src/state/archival_state.rs
```diff
@@ -906,6 +907,75 @@ impl ArchivalState {
...
+                .await?
+            else {
+                continue;
+            };
+            let Some(end_block_hash) = self
+                .canonical_block_digest_of_aocl_index(capped_range_end)
+                .await?
+            else {
+                continue;
+            };
```
Sword-Smith, line 956: ~~Those absolute index sets that the node cannot find the origin off are just skipped here. So the returned result might not have the same length as the request. How does the caller then map the absolute indices to the returned block header/block digests? I think the return value must have the same length as the request. So maybe we return `Vec<Option<(Digest, BlockHeader)>>` from this function, and change `UtxoOriginFromAbsoluteIndicesResponse` equivalently?~~

Edit: The above reasoning doesn't make sense. Let me reason from first principles:
- The caller knows the AOCL leaf index
- He doesn't want to reveal it to the neptune-core server, so the caller sends the absolute index set instead
- If the caller gets (block hash, block header) pairs, he needs to match those with the relevant AOCL leaf index. So the returned type needs to include the AOCL range of the block in question.

So the caller needs to filter out the relevant elements from the returned list. To do this, the caller needs to more fields in `BlockInfo` which could be renamed to `BlockAoclInfo`.

Suggested type of returned elements:
```rust
pub struct BlockAoclInfo {
    header: BlockHeader,
    block_hash: Digest
    min_aocl_index: u64,
    max_aocl_index: u64
}
``` 

Does this make more sense? The you can use the same logic with the `continue` as you do above. As it's now the caller's responsibility to find the relevant entries in the response.

Sword-Smith (changes requested):

zeokin:
@Sword-Smith Thanks for your explanation.
I've fixed. Please check again.

neptune-core/src/state/archival_state.rs
```diff
@@ -906,6 +907,95 @@ impl ArchivalState {
...
+                .expect("Number of genesis outputs must fit in u64");
+            return (0, num_genesis_outputs.saturating_sub(1));
+        }
+
+        let record = self
+            .get_block_record(block_digest)
+            .await
+            .expect("Canonical block digest must have a known block record");
+        (record.min_aocl_index, record.max_aocl_index())
+    }
```
Sword-Smith: It's enough if this function takes `block_digest`. It shouldn't take `block_height` also. It should take one or the other, not both. And it looks to me like `block_digest` is the best fit. To check if the requested block is the genesis block use `self.genesis_block.hash()`. This way, this function's signature becomes 

```
async fn block_aocl_index_range(
        &self,
        block_digest: Digest,
    ) -> (u64, u64)
```

Also: Let's add a docstring to the function. Something like "Returns the inclusive range of AOCL leaf indices added by the specified block.". And if the function can panic, that should be declared: "panics if block digest is not known."

Sword-Smith (changes requested):
Logic seems fine. Looks like we're down to stylistic issues now.

neptune-core/src/state/archival_state.rs
```diff
@@ -906,6 +907,94 @@ impl ArchivalState {
...
+
+    pub(crate) async fn utxo_origin_blocks_from_absolute_index_sets(
```
Sword-Smith, line 926: The docstring on this method should show that only canonical blocks are scanned.

neptune-core/src/state/archival_state.rs
```diff
@@ -906,6 +907,94 @@ impl ArchivalState {
...
+                .await?
+            else {
+                continue;
+            };
+            let Some(end_block_hash) = self
+                .canonical_block_digest_of_aocl_index(capped_range_end)
+                .await?
+            else {
+                continue;
+            };
```
Sword-Smith: If you rewrote `canonical_block_digest_of_aocl_index` to return both the digest *and* the height of the block, you could save a lot of work.

neptune-core/src/state/archival_state.rs
```diff
@@ -906,6 +907,94 @@ impl ArchivalState {
...
+                .ammr()
+                .try_get_leaf(height.value())
+                .await
+                .expect("Canonical block height must have a block digest");
+            let header = self
+                .get_block_header(block_hash)
+                .await
+                .expect("Canonical block digest must have a known block header");
+
+            let (min_aocl_index, max_aocl_index) = self.block_aocl_index_range(block_hash).await;
```
Sword-Smith: I think you can make this a bit simpler:

```
let (min_aocl_index, max_aocl_index) = if block_hash == self.genesis.hash() {
    ... 
} else {
    let record = self.get_block_record(block_hash).await.expect("Block record of canonical hash must exist");
    let (min_aocl_index, max_aocl_index) = (record.min_aocl_index, record.max_aocl_index());
}
```

Sword-Smith (changes requested):

neptune-core/src/state/archival_state.rs
zeokin: Perfect. Thanks for letting me know.

Sword-Smith:
Thanks. Looks really good. I still have a few stylistic wishes though. And your git management makes it hard for me to address that since you merge-in `master` instead of ensuring that your PR is already built on top of `master`. A frequent use of `git fetch origin && git rebase origin/master` will make it easier for me to merge your PRs.

Because of the messy git state of this PR, I'll just be merging this through the CLI.

Sword-Smith:
Merged in CLI as c0c3bade25db5af319061f75a7c47670845df4a1.

PR #936, opened 2026-06-06
fix(`proof_collection.rs`): Check zip operands' lengths

aszepieniec:
Transactions initiated by malicious parties can have a mismatching number of lock scripts versus lock script hashes; or type scripts versus type script hashes. In this case, the zip will yield zipped pairs matching the lenth of the shortest operand. If zero, no lock scripts (or type scripts) are verified.

Such malicious transactions can only exist as `ProofCollection` transactions and cannot be upgraded to `SingleProof` because the tasm code iterates over the list of hashes and fetches the matching script proof. If no such script proof is present, validation will simply fail. So the tasm code does the right thing. This commit makes the host machine do the right thing as well.

That said, malicious parties can still broadcast malicious `ProofCollection` transactions, and they will land in the mempool. They will never be confirmed, but as long as they are in the mempool they might cause the wrong amount to be shown in 'unconfirmed balance'.

### DIFF
--- a/neptune-core/src/protocol/consensus/transaction/validity/proof_collection.rs
+++ b/neptune-core/src/protocol/consensus/transaction/validity/proof_collection.rs
@@ -281,6 +281,19 @@ impl ProofCollection {
             return false;
         }
 
+        // There must be exactly one halting proof per collected script hash.
+        // The verification loops below use `zip`, which silently truncates to the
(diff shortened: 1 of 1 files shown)


Sword-Smith (approved):

Sword-Smith:
LGTM

PR #938, opened 2026-06-10
Hard fork gamma

Sword-Smith:
This hardfork is active on `testnet-0` now. Activated at block height 4650.
http://51.158.203.7/block/height/4650

First subtraction from lustration counter:
http://51.158.203.7/block/height/4652

Revalidated all historical blocks on `testnet-0` with `neptune-cli revalidate-history`:
```
2026-06-10T00:45:03.231285426Z  INFO ThreadId(24) neptune_cash::state: All blocks in range 1..=4652 are valid
```

I'll keep mining on testnet-0 to get more blocks on hardfork-gamma. Watch the above block explorer for progress.

Missing

- [x] Add integration test of HF gamma activation on testnet-0
- [x] Decide on rollback height


aszepieniec (commented):
I'm confused about e55bcf6. You write:

> The only alternative I see would be to hardcode a listof block hashes and just treat  those as valid.

That's exactly what we did following the vulnerability in January. That's also exactly what 2559425 does. If that's the strategy we are going with, why neuter this commit? If that's not the strategy we are going with, why accept it? And more importantly, what did we miss back in January/February?

If I understand correctly, the counterpart to this 'alternative' is the change in this commit, described as:

> Allow for the validation of old blocks, without dragging around old versions of SingleProof and BlockProgram.

Allowing the validation of unsound proofs strikes me as a bad idea. What is to stop a powerful attacker from reorganizing deeper than the threshold so as to get back in the regime where unsound proofs are accepted, and then initiate inflationary transactions there?

You write:

> that solution [of hardcoding true block claims of historical blocks] disallows the user from checking the validity of historical blocks in other regards than just the block proof.

I'm not sure what you mean. What are the other regards? The other regards that are not the block proof that I can come up with do not involve the block proof or Triton VM at all, so they can still be validated.

neptune-core/src/bin/neptune-blk-info.rs
```diff
@@ -0,0 +1,56 @@
...
+
+use anyhow::bail;
+use anyhow::Result;
+use clap::Parser;
+use neptune_cash::state::archival_state::ArchivalState;
+
+#[derive(Parser, Debug, Clone)]
+#[clap()]
+struct CliArg {
+    path: PathBuf,
```
aszepieniec, line 16: Why not use the default path and wrap this field in an option? Also, is this data-dir/ or data-dir/blocks/?

aszepieniec (approved):

neptune-core/src/bin/neptune-blk-info.rs
Sword-Smith: You point it to a `blk` file. So: `data-dir/blocks/blk12.dat`.

PR #947, opened 2026-07-11
refactor: Split archival mutator set into its own crate

zeokin:
Extract the database-backed archival mutator set out of `neptune-mutator-set` into a new `neptune-archival-mutator-set` crate, so that consumers of `neptune-consensus` that only need the in-memory accumulator no longer transitively depend on `neptune-database`.

The pure-data membership-proof types (IndexedAoclAuthPath, MsMembershipProofPrivacyPreserving) stay in neptune-mutator-set so the RPC API contract also remains database-free. Tests that exercise the accumulator against a database-backed archival set move to the new crate (avoiding a dev-dependency cycle).


Sword-Smith:
Looks good. The meat of this PR is that `neptune-mutator-set` no longer depends on `neptune-database` which means that `neptune-consensus` also doesn't depend on `neptune-database`.


image

Sword-Smith:
Also looks like all tests are preserved. Very nice!! Good job.

I reserved the name `neptune-archival-mutator-set` on crates.io just now such that we can publish everything once we're ready to.

I just pushed a import-reordering. But apart from that, everything looks green. LGTM!

Sword-Smith (approved):

Sword-Smith:
I benchmarked compilation times for this split. 

```
  ┌───────────────────────────┬──────────┬──────────┬──────────────┐
  │         Scenario          │  master  │  branch  │      Δ       │
  ├───────────────────────────┼──────────┼──────────┼──────────────┤
  │ Wall clock, -j (24 cores) │ ~35.9 s  │ ~35.3 s  │ ~0 (noise)   │
  ├───────────────────────────┼──────────┼──────────┼──────────────┤
  │ Total CPU work (user+sys) │ ~311.5 s │ ~286.2 s │ −25 s (~8%)  │
  ├───────────────────────────┼──────────┼──────────┼──────────────┤
  │ Wall clock, -j2           │ 108.4 s  │ 102.7 s  │ −5.6 s (~5%) │
  └───────────────────────────┴──────────┴──────────┴──────────────┘
```

PR #949, opened 2026-07-13
keep and upgrade unsynced ProofCollection transactions

zeokin:
Changes
1. It makes the transaction flicker in-and-out of the mempool giving a bad user experience
2. The update part is cheap. Anyone performing a "raise" operation (`ProofCollection`->`SingleProof`) can just follow it up by an "update" operation. So the proof upgraders should keep outdated `ProofCollection` transactions in their mempool.


Sword-Smith:
Excellent tests, a bit too verbose comments. I separated the two fixes from one into two commits and rebased against `master`.

PR #964, opened 2026-08-10
Transaction Chaining

aszepieniec:
# Transaction Chaining

Adds a second transaction pipeline alongside the single-proof one. *Link transactions* (`LinkTx`) spend outputs that are not yet confirmed. Merging such transactions enables *cut-through*, which deletes the transaction output-and-input altogether, for each such put.

A `LinkKernel` is a `TransactionKernel` plus one extra MAST leaf, `thruputs`. It contains addition records that are an unconfirmed output of some predecessor transaction and an input to this one. A `LinkTx` with a non-empty thruput set cannot enter a block. A `LinkTx` with an empty thruput set can be made into a regular `Transaction` which can enter into a block.

## Consensus change

`LinkProof` is a new consensus program with four branches:

| branch | signature | what it does |
|---|---|---|
| `Forge` | witness → `LinkTx` | initiation; inlines `RemovalRecordsIntegrity`, `KernelToOutputs`, `CollectLockScripts`, `CollectTypeScripts` non-recursively |
| `Chain` | `LinkTx * LinkTx → LinkTx` | combine two links, cancelling every matching (output, thruput) pair on both sides |
| `Advance` | `LinkTx → LinkTx` | re-target to a newer mutator set, promoting thruputs it has since confirmed into ordinary inputs |
| `Cast` | `Transaction → LinkTx` | pull a single-proof transaction into a chain |

`SingleProof` gains two branches:

- `Fix: LinkTx → Transaction` -- recursively verifies the link proof; the derived kernel MAST hash is well-formed only when the thruput set is empty.
- `Weld: Transaction * LinkTx → Transaction` -- what `Fix(Chain(Cast(A), B))` computes in three proofs and four recursive verifications, in one proof and two.

Block format is unchanged: a block holds exactly one `SingleProof`-backed `Transaction`, and a `LinkTx` never enters one.

New rule set `ConsensusRuleSet::HardforkDelta`, activating at height 55 000 (mainnet) / 5 400 (testnet). `SingleProof` becomes a family of programs indexed by the rule set. This family contains two members, so there are two distinct `SingleProof` hashes . The pre-delta program runs unchanged until activation.

The `SingleProof` digest `D` is a parameter of the `LinkProof` claim rather than a constant inside it. This indirection is necessary to break the `Fix`/`Cast` definitional cycle. (Otherwise `Fix` needs a hardcoded `Cast` program hash; and `Cast` needs a hardcoded `Fix` program hash.)

## Why

- A block admits at most one interaction with a given contract UTXO, capping *e.g.* a DEX at one trade per block. Chaining collapses many incremental updates into one `Transaction` fit for a `Block`.
- Dependent payments, change-splitting, and consolidation no longer need a confirmation in between.
- Without promotion, a successor whose predecessor is mined alone is stranded permanently, recoverable only by re-forging. `Advance` moves confirmed thruputs into the confirmed inputs instead.
- `Forge → Fix` benchmarks ~30% cheaper than `Collect` + `Raise`; chained-versus-separate throughput is ~36% better at N = 2 (`cargo bench -p neptune-consensus --bench chaintx`).

## Follow-ups (none of which block this merge)

- [x] Documentation
- Consensus change and hard fork logic
- Mempool policy

See open todo-items sprinkled across this thread.


aszepieniec:
## Transaction chaining — benchmark results

`neptune-consensus/benches/chaintx.rs`; `cargo bench -p neptune-consensus --bench chaintx`.

Measured on my i9-12900K (16 cores / 24 threads, 123 GB RAM) with `TVM_LDE_TRACE=no_cache`.

### Claim 1 — `Forge + Fix` beats `Collect + Raise`

One transaction, 2 in / 2 out / 1 announcement, same primitive witness down both routes.

| | legacy | chaining |
|---|---|---|
| Σ padded height | 2 183 168 | 1 572 864 |
| prover time | 858 s | **599 s** |

<details><summary>breakdown</summary>

```
  chained: Forge -> Fix
    Forge[0]  height  1048576     422.1 s
    Fix       height   524288     177.1 s
    TOTAL     height  1572864     599.2 s

  legacy: Collect -> Raise
    RemovalRecordsIntegrity  height    65536      12.1 s
    CollectLockScripts       height     4096       0.5 s
    KernelToOutputs          height     8192       0.8 s
    CollectTypeScripts       height     8192       0.7 s
    Raise                    height  2097152     844.1 s
    TOTAL                    height  2183168     858.2 s
```
</details>

**≈30% cheaper.**

### Claim 2 — chaining N interactions beats N separate transactions

Not like-for-like: the chained fixture's successor spends its predecessors' *unconfirmed* outputs, which the legacy pipeline cannot express at all. The closest counterpart is N independent transactions raised and merged.

| N = 2 | legacy | chaining |
|---|---|---|
| Σ padded height | 5 414 912 | 3 670 016 |
| prover time | 2 119 s | **1 361 s** |

<details><summary>breakdown</summary>

```
  chained: 2 interactions
    Forge[0]  height  1048576     392.5 s
    Forge[1]  height  1048576     391.7 s
    Chain[0]  height  1048576     397.5 s
    Fix       height   524288     178.9 s
    TOTAL     height  3670016    1360.6 s

  legacy: 2 separate transactions
    RemovalRecordsIntegrity[0]  height    65536      11.5 s
    CollectLockScripts[0]       height     4096       0.5 s
    KernelToOutputs[0]          height     8192       0.9 s
    CollectTypeScripts[0]       height     8192       0.8 s
    Raise[0]                    height  2097152     839.9 s
    RemovalRecordsIntegrity[1]  height    65536      11.5 s
    CollectLockScripts[1]       height     4096       0.5 s
    KernelToOutputs[1]          height     8192       0.7 s
    CollectTypeScripts[1]       height     8192       0.8 s
    Raise[1]                    height  2097152     850.4 s
    Merge[0]                    height  1048576     401.7 s
    TOTAL                       height  5414912    2119.2 s
```
</details>

**≈36% cheaper.** At the margin: one more chained interaction is `Forge + Chain` ≈790 s, against ≈1265 s for one more legacy transaction.

### Why

Not the RRI inlining: proving RRI alone is 12.1 s, 1.4% of the legacy total. It is *recursive verification* which dominates all of these programs. `Forge+Fix` nets 3 fewer recursive verifications than `Raise`.

`N = 4` compiles but has not been run. `no_cache` makes both columns pessimistic in absolute terms; the ratio is unaffected.

Sword-Smith:
Thanks for the benchmark. Very cool that this is faster than "collect" + "raise"! 

As you mention, all that really matter is the number of in-vm verifications you have to do. Since "collect" alone (without the "raise" step) is still much faster than the "chaining" approach, I think that "collect" should stay as the default proving method that downstream wallets use. That being said, that's of course up to each wallet implementation and outside of our control.

Sword-Smith:
The auditing of the entire witness might be a bit excessive. I wonder what you pay for that in clock cycles. I'm thinking about this TASM:

```rust
let audit_preloaded_data =
            library.import(Box::new(VerifyNdSiIntegrity::<ChainWitness>::default()));
```

Maybe it's OK? I think it's worth investigating what the witness auditing costs. We don't want that to become dominant in the future, for any transactions sizes. Consider using more "WitnessMemory" structures if auditing is too expensive.

Sword-Smith:
Let me suggest that we don't call the existing transaction pipeline the "legacy" pipeline, since that communicates that it will be retired, which is not the case. Let's instead call it the "single-proof" pipeline.

Just posting a stream of consciousness here. Maybe it'll be useful for others.

## Vocabulary
- **confirmed input** An input to a `LinkTx` that comes from the mutator set. So a UTXO that was previously mined, and that is being spent in this `LinkTx`.
- **thruputs** Inputs to a `LinkTx` that comes from outputs from another `LinkTx`.

The reason that "update" on a transaction with no inputs was disallowed, was that the "update" of a "nop" transaction would grant composers a fast way to mine empty blocks. With the current ban on updating a transaction with no inputs, composers wanting to mine an empty block are required to create a "nop" transaction from scratch for each new block proposal, going through "raise" for both their coinbase transaction, and their "nop" transaction. Including a transaction from the mempool in their block proposal requires the composer only to go through one "raise" and an "update", as opposed to two "raise".

Let's ensure that composers keep this incentive, to include mempool transactions in their block proposals.

## New Transaction Pipeline
image

Sword-Smith:
So we need to ban "update" on `LinkTx` when the `LinkTx` has no confirmed inputs *and* no thruputs.

And we should ban negative fees throughout the `LinkTx` pipeline. Those are only required in single proof's "merge" operation.

Also: The `SingleProof` program digest could be printed to std out when it is used to construct the `SingleProof` claim inside of the `LinkTxProof` program.

edit:
Alan agrees on:
- ban "update" in chain pipeline when `confirmed_inputs.len() + thruputs.len() == 0`.
- Ban negative fees throughout the chain pipeline, transitively banned through entrypoints "cast" and "forge"
- Agree on the security-in-depth/redundant check that prints the single proof program digest in all the link proof executions. This changes the claim for the `LinkProof` program.

Sword-Smith:
Updated to latest `triton-vm` and `tasm-lib` version: `v8.0.0`.

Sword-Smith:
Missing internal audits:

**LinkProof**
- [x] Forge
  - [x] a. coinbase is `None`
  - [x] b. merge bit is `false`
  - [x] c. fee is non-negative
  - [x] d. All confirmed inputs live in the AOCL and match the absolute indices in the witness data -- AOCL membership / absolute index set correspondence
  - [x] thruputs ++ confirmed inputs == input UTXO witness data
  - [x] outputs == output UTXO witness data
  - [x] All output elements match UTXOs from witness field `outputs`
  - [x] `NativeCurrency` typescript is always present. Also if inputs, thruputs, and outputs are empty.
  - [x] Type script claims list constructed correctly
  - [x] All type scripts halt
  - [x] Lock script claims list constructed correctly
  - [x] All lock scripts halt
  - [x] no bfe-wrap when summing number of thruputs and number of (confirmed) inputs.
  - [x] All fields MAST verified before use
      - [x] coinbase (a)
      - [x] merge bit (b)
      - [x] AOCL MMR from witness (intial verification)
      - [x] confirmed inputs
      - [x] thruputs
      - [x] outputs
      - [x] coinbase
      - [x] merge bit
      - [x] fee
  - [x] Non-MAST authenticated fields don't need to be authenticated
    - [x] announcements
    - [x] timestamp
- [x] Chain
  - [x] Verify proof of left and right link-transaction
  - [x] left inputs ++ right inputs == new inputs
  - [x] Cut-through cancels outputs, and cancels thruputs simultaneously
  - [x] fee is summed and bounded
  - [x] all mutator set hashes agree: left, right, new
  - [x] announcements are concatenated
- [x] Advance
  - [x] a. Previous `LinkProof` is valid for the relevant link-kernel MH
  - [x] b. no coinbase
  - [x] c. merge-bit not set
  - [x] d. `new_timestamp >= old_timestamp`
  - [x] e. disallow empty thruput + empty inputs
  - [x] f. Fee unchanged
  - [x] g. announcements unchanged
  - [x] h. outputs unchanged
  - [x] i. mutator set moved forward
  - [x] j. All promoted thruputs are members of the new AOCL
  - [x] k. Promoted thruputs inputs are calculated from the same witness data as the verified AOCL leafs
  - [x] l. All promotions have strictly increasing AOCL leaf indices
  - [x] m. new thruputs are old throughputs minus promotions <=> set(old thruputs) = set(new thruputs) ++ set(promotions)
  - [x] n. new inputs are old inputs plus promotions
  - [x] o. promotion-witness loop iteration only uses bounded jumps.
  - [x] All fields MAST verified before use
    - [x] old mutator set hash / old AOCL <- old lkmh (during i)
    - [x] new mutator set hash new AOCL -< new lkmh (i)
    - [x] old inputs (j/k/l/m)
    - [x] new inputs (j/k/l/m)
    - [x] old thruputs (j/k/l/m)
    - [x] new thruputs (j/k/l/m)
    - [x] old outputs (h)
    - [x] new outputs (h)
    - [x] old announcements (g)
    - [x] new announcements (g)
    - [x] new fee (f)
    - [x] old fee (f)
    - [x] old timestamp (d)
    - [x] new timestamp (d)
    - [x] new coinbase (b)
    - [x] new merge bit (c)
  - [x] Non-MAST authenticated fields don't need to be authenticated
    - [x] old coinbase (`Forge` and `Cast` block coinbase from being anything but `None`)
    - [x] old merge-bit (as above)
- [x] Cast
  - [x] disallow set coinbase field
  - [x] disallow set merge bit
  - [x] force fee to be non-negative
  - [x] claimed tx-kernel MH matches link-kernel MH in claim (with empty thruputs set)
  - [x] `SingleProof` for tx-kernel is valid

**SingleProof**
- [x] Fix
  - [x] Construct claim correctly
  - [x] Verify claim

Everything else is window dressing.

aszepieniec:
After updating TODO.md to reflect recent progress and decisions, the following items remain open. I write them here because most of them are up for discussion -- a discussion that is easier to have here than in that file.

## Documentation

 - [x] Extend `transaction.md` with an accurate description of the dual
      (introduced in this development streak) pipeline and graph.
 - [x] Glossary: `thruput` = an `AdditionRecord` that is simultaneously (a) an
      unconfirmed input to this tx and (b) an output of a predecessor in the
      chain of transactions.
 - [x] Glossary: the **promotion equations** = the coupled pair `Advance`
      enforces over the witness-supplied promotion set `P`, both compared as
      multisets:
```
          multiset(old.thruputs) == multiset(new.thruputs) ⊎ records(P)
          indexsets(new.inputs)  == indexsets(old.inputs) ⊎ indexsets(P)
```
      Name them in `transaction.md` and write them out there verbatim. The
      term already earns its keep in the code -- `assert_promotion_equations`
      is the tasm subroutine that checks exactly this pair -- so the spec
      should use the same words. Say why they are one thing and not two: an
      addition record leaves the thruput list if and only if a removal record
      carrying its index set joins the input list. Say what `P = []` means:
      thruputs and index sets both unchanged, which is the update that
      promotes nothing.
 - [x] Write §Breaking the `Fix`/`Cast` cycle into `transaction.md` too — the
      claim shape `[lkmh, D]` is consensus-visible, so it belongs in the spec,
      not only in this tracker.
 - [x] Security argument for cut-through: no value creation, no double-spend
      across a chain; and the argument that RRI proven once in Forge and carried
      by recursion through `Chain`/`Fix` is as sound as a recursive `Raise`.
      Ensure the test suite has a test catching every claim here.

## Consensus
  - [ ] `ConsensusRuleSet::default()` is still gamma, deliberately: it is the
        rule set in force, and a caller that forgets to thread the real one
        should fall back to what today's peers accept. Move it with the
        activation.
 - [x] soundness audit: `SingleProof` `Fix` branch, and `Forge`'s inlined RRI
       + the recursion in `Chain`/`Advance`/`Cast`, including the verbatim `D`
       pass-through
 - [x] Regenerate/store proof artifacts for the new program versions

## Integration
 - [ ] Announcement payload for self-perpetuating UTXOs:
      `{ utxo, sender_randomness, receiver_preimage }` — `UtxoTriple`
      (`transaction/utxo_triple.rs:18`) with the preimage in place of the
      digest, so it yields a `UtxoTriple` for free (`receiver_digest =
      receiver_preimage.hash()`) and reuses its `addition_record()` rather than
      restating the commitment. Written by the transaction that *creates* the
      contract UTXO, for whoever later promotes a thruput spending it.
      Deliberately not `TransparentInput`
      (`transaction/transparent_input.rs:37`), which additionally carries
      `aocl_leaf_index`: the creating transaction cannot know it, an output's
      AOCL position being fixed only by the block that confirms it and by what
      lands ahead of it there. `TransparentInput` can carry it because it is
      written by the *spender*, who already holds a membership proof. The
      promoter recovers the index instead, by matching the announced commitment
      against the confirming block's addition records.
      Publishing the preimage grants no spending authority: it lets a third
      party *complete* an authorization the forger already gave -- the
      lock-script proof was made at `Forge` time -- and the commitment equation
      pins that authorization to one thruput.
      
## Mempool

 - [ ] Index mempool-members (of either type) by: confirmed inputs (existing
      double-spend index), thruputs (new + if applicable), and outputs (new) —
      the last two are what let an arrival find its `Chain` partners.
 - [ ] On arrival, look up predecessors (`member.outputs` ⊇ `arrival.thruputs`)
      and successors (`member.thruputs` ⊆ `arrival.outputs`), and perform
      cut-through on matching pairs if the fee is large enough.
 - [ ] Conflict rules: two mempool members on the same confirmed input (already a
      conflict now), OR two successors with overlapping thruputs (new). In case
      of conflict, replicate existing policy and exit-queue construction.
 - [ ] Evict members whose predecessor was dropped: nothing re-checks
      `contains_outputs` after a block, so an orphaned successor lingers
      until the (unbuilt) TTL.
 - [ ] A member whose predecessor was *confirmed* is no longer evicted --
      its thruputs are promoted instead. `mempool.rs:1704` currently does
      the opposite, on the pre-promotion premise that a mined thruput can
      never be cut through. Blocked on the preimage source: §582
      (announcement) or §626 (wallet).
 - [ ] Promotion on advance, best-effort, chosen by preimage availability:
      (a) published in an announcement (self-perpetuating UTXOs) -- any node
      promotes; (b) held by the wallet -- only the owner promotes; (c) neither
      -- plain re-target, and the member stays unresolved until it chains or
      its TTL expires. Candidates and their `aocl_leaf_index`es both fall out of
      matching members' thruputs against the new block's addition records,
      which is a scan the mempool needs anyway to notice a thruput was
      confirmed.
 - [ ] TTL for `LinkTx`s whose thruputs never resolve (new).
 - [ ] Cast-on-demand: when a `LinkTx` can chain onto a mempool-member single-proof
      `Transaction`, `Cast` the latter in if the fee is beneficial.
      
## Tests
 - [ ] `Forge → Chain → Fix` with the real `D` throughout accepts and the
      resulting `SingleProof` verifies.
 - [ ] `Cast → Chain → Fix` with the real `D` throughout accepts and the
      resulting `SingleProof` verifies.
 - [ ] `Fix` distributivity: `Fix(Chain(A, B)) = Merge(Fix(A), Fix(B))` when
      `thruputs == []`.
 - [ ] `Advance` then `Fix` equals `Fix` on the advanced mutator set.
 - [ ] A `Fix`'d transaction enters a block. Its single proof already verifies
      against `single_proof_claim`
      (`forge_then_fix_yields_a_verifying_single_proof`,
      `cast_then_fix_yields_a_verifying_single_proof`); the block half waits on
      §Integration.
      
## Misc

- [x] Delete plan / progress tracker (TODO.md)

Sword-Smith:
In `Weld`, we should allow the `SingleProof` transaction to carry a coinbase, as that increases the likelihood that composers pick up chained transactions. Otherwise, I'm happy, apart from some AI slop comments in the code.

How can you be that good at programming, and that terrible at documentation?

Sword-Smith:
All tests fixed, and `make happy` happy.

aszepieniec:
- [ ] Teach the merge-input cache about welds, when something in the application
      actually welds. A weld conflicts with its own operand A -- it carries A's
      inputs -- so inserting it evicts A, and `have_merge_relationship`
      (`transaction_kernel.rs:176`) is what decides whether A's single proof is
      kept against the weld never being mined (`mempool.rs:1051`). It answers no
      for every weld, A's cut-through outputs being absent from it, so A is
      dropped with no copy kept and its owner must re-submit and re-prove. The
      small fix is a sibling predicate for welds: keep the inputs and
      announcements subset checks and the timestamp check, drop the ones on
      outputs. Not written yet on purpose -- the right shape depends on how the
      mempool comes to produce welds, which is §Mempool's own unbuilt half.

aszepieniec:
@Sword-Smith: FYI I signed five of your commits. You did not screw up -- I did. I had made unsigned commits and you followed up with signed ones. After rebasing to sign my own, your signatures were invalidated. So I added a signature back. Now all commits are signed. Authorship is unchanged.

Sword-Smith (approved):

PR #965, opened 2026-08-12
refactor: Make `fork_choice_rule` stable

Sword-Smith:
A few changes to `fork_choice_rule` that
1. Are consistent with the existing rule once a chain split is two deep or higher, meaning that the change is neither considered a hardfork or a softfork.
2. Still attempts to favor the composer that picks up real transactions.
3. Is stable and objective, unlike the past algorithm. Previously, the algorithm would favor the existing block (the one that came first) *if* it had inputs, and the new block if the existing tip did *not* have inputs. This was terrible since a) the first-come is sometimes favored, making the rule subjective; and b) was an unstable rule such that a node's tip could flicker between two blocks of the same height if neither of them had any inputs.

This closes the medium severity issue NPT-3 in R9.

The new fork-choice rule that I suggest is:
```rust
    pub fn fork_choice_rule<'a>(current_tip: &'a Self, incoming_block: &'a Self) -> &'a Self {
        if current_tip.hash() == incoming_block.hash() {
            // The blocks are the same. Prefer current to avoid a meaningless
            // state update.
            return current_tip;
        }

        // If blocks have different parent, prefer the one with most cumulative
        // proof of work.
        let current_cumpow = current_tip.header().cumulative_proof_of_work;
        let incoming_cumpow = incoming_block.header().cumulative_proof_of_work;
        if current_tip.header().prev_block_digest != incoming_block.header().prev_block_digest {
            if current_cumpow >= incoming_cumpow {
                return current_tip;
            }
            return incoming_block;
        }

        // Candidates have the same parent. Prefer the one with most inputs.
        let current_num_inputs = current_tip.body().transaction_kernel.inputs.len();
        let incoming_num_inputs = incoming_block.body().transaction_kernel.inputs.len();
        if current_num_inputs != incoming_num_inputs {
            if current_num_inputs > incoming_num_inputs {
                return current_tip;
            }
            return incoming_block;
        }

        // Candidates have same parent and same number of inputs. Prefer the one
        // with the most cumulative proof of work.
        if current_cumpow >= incoming_cumpow {
            current_tip
        } else {
            incoming_block
        }
    }
```

### DIFF
--- a/neptune-consensus/src/block/mod.rs
+++ b/neptune-consensus/src/block/mod.rs
@@ -1071,31 +1071,19 @@ impl Block {
     /// evaluates the following logic:
     ///  - if the height is different, prefer the block with more accumulated
     ///    proof-of-work;
-    ///  - otherwise, if exactly one of the blocks' transactions has no inputs,
-    ///    reject that one;
-    ///  - otherwise, prefer the current tip.
+    ///  - if the height is the same, prefer the current tip.
     ///
     /// This function assumes the blocks are valid and have the self-declared
     /// accumulated proof-of-work.
-    ///
-    /// This function is called exclusively in
-    /// `GlobalState::incoming_block_is_more_canonical`, which is in turn
-    /// called in two places:
-    ///  1. In `peer_loop`, when a peer sends a block. The `peer_loop` task only
-    ///     sends the incoming block to the `main_loop` if it is more canonical.
-    ///  2. In `main_loop`, when it receives a block from a `peer_loop` or from
-    ///     the `mine_loop`. It is possible that despite (1), race conditions
-    ///     arise, and they must be solved here.
     pub fn fork_choice_rule<'a>(current_tip: &'a Self, incoming_block: &'a Self) -> &'a Self {
-        if current_tip.header().height != incoming_block.header().height {
-            if current_tip.header().cumulative_proof_of_work
-                >= incoming_block.header().cumulative_proof_of_work
-            {
-                current_tip
-            } else {
-                incoming_block
-            }
-        } else if current_tip.body().transaction_kernel.inputs.is_empty() {
+        let (current, incoming) = (current_tip.header(), incoming_block.header());
+
+        // At equal height, the first-seen block is preferred. Otherwise, prefer
+        // the one with most cumulative proof of work. Preferring the first seen
+        // at equal heights disincentivizes selfish mining.
+        let incoming_wins = current.height != incoming.height
+            && incoming.cumulative_proof_of_work > current.cumulative_proof_of_work;
+        if incoming_wins {
             incoming_block
         } else {
             current_tip


aszepieniec (commented):
# Obligatory Warning

The fork choice rule is consensus-critical code. It is a key component in the mechanism that funnels individual, mutually distrusting, potentially malicious, peers into agreement on the blockchain state. Besides network splits, there are other blockchain failure modes that are equally dramatic that could result from a seemingly benign fork choice rule change. It should therefore not be modified lightly but with utmost care and caution.

# Difference

**Current:**

 - if the height is different, prefer the block with more accumulated proof-of-work;
 - otherwise, if exactly one of the blocks' transactions has no inputs, reject that one;
 - otherwise, prefer the current tip.

I scrutinized the code and believe that it *disagrees* with this policy. It's very subtle.

```rust
        } else if current_tip.body().transaction_kernel.inputs.is_empty() {
            incoming_block
        }
```

So if the tip has no inputs, and another block with the same height is received that *also* has no inputs, the code will prefer that newcomer whereas the policy does not.

This is a shortcoming and needs a fix. That said, there is more to the proposal than a minimal fix. So for the purpose of weighing the proposal's merits I will compare it to the textual description of the current policy.

**Proposal:**

 - if the **parents** are different, prefer the block with more accumulated proof-of-work;
 - otherwise, prefer the one with **the most** inputs.
 - if same parents and same number of inputs, prefer the one with the most cumulative proof of work, which is equivalent to the one with **the highest difficulty**.

Emphasis mine. I scrutinized the code and believe that it agrees with this policy -- and, moreover, that agreement is readily apparent. (Good job!)

# Current Shortcomings

## Unstable

> [The previous fork choice rule] was an unstable rule such that a node's tip could flicker between two blocks of the same height if neither of them had any inputs.

If I understand correctly, the mechanic is:
 - Your tip is block A which has 0 inputs.
 - You receive new block B of the same height and also 0 inputs. Prefer B. (Erroneously.)
 - You receive new block A (again). Prefer A. (Again erroneously.)
 - Repeat.

This is a dynamic resulting from the faulty implementation, and does not follow from the textual policy description. While it is an attack vector, it affects only miners who do not confirm any transactions with at least one input. Countermeasure until we fix it: if there are transactions in the mempool, confirm them already; if not, make one.

## Subjective

You get different answers about what the tip is based on who you ask. That possibility is a consequence of physical networks and hardware bounding propagation speed and computation. New blocks are not accepted by all nodes at the same instant. In fact "the same instant" is not even a physically tenable quality if relativity is right -- so it has no business living in a definition of what objective and subjective is.

The way we can define objectivity for consensus mechanisms is in the limit. As time passes, all parties will agree on a strictly growing prefix of history.

I think the target of this accusation is the property of the current fork choice rule that makes nodes wait until the *next* block before a race between equal-height blocks is resolved. The fork choice rule is still objective because the disparity will be resolved unambiguously eventually, but it persists for a concretely long duration. Call this duration the *subjectivity window*. It is a metric, not a boolean property. The proposed fork choice rule improves on this metric: it has a much smaller subjectivity window because it forces an immediate resolution to the race.

The change in outcome is more than merely closing the subjectivity window sooner. Phrased precisely, the distribution of limiting histories resulting from the current policy is concretely different from the distribution of limiting histories resulting from a minimal fix forcing an immediate resolution to a block race. It is because the direction in which this block race is immediately resolved in the minimally fixed case may differ from how it is eventually resolved in current policy. If that's too abstract, here is an example:

**Current:**

 - Consensus: current tip is block A.
 - Two blocks, B and C, are announced in short succession.
 - A block race occurs. Some miners mine on B; some mine on C.
 - A successor D to block B is found and disseminated. Nodes who were synced to C switch to D.
 - Resulting history: ABD.

**Minimal Fix:**

 - Consensus: current tip is block A.
 - Two blocks, B and C, are announced in short succession.
 - A block race occurs but is immediately resolved in favor of C.
 - All miners mine on C and eventually find a successor E on top of C.
 - Resulting history: ACE.

The point that I am trying to make by highlighting this difference in outcome is that reducing the subjectivity window by forcing immediate resolution is not free but has knock-on effects. In particular, the immediate resolution biases the distribution of limiting histories in favor of how the race is immediately resolved. *A priori*, it could be the only side-effect and a benign one, or neither, or anywhere in between.

# Other Properties and Metrics

I can think of a bunch of properties and metrics that I think are important for a fork choice rule to satisfy and score highly on. Let me mention the ones that I think are not relevant to this discussion briefly, and then expand on the ones that might be.

Not relevant here (probably):

 - Running time. We want a *concretely fast* algorithm for evaluating the fork choice rule and such that the running time is *independent of the inputs*.
 - Commutativity. Evaluating on (A, B) and on (B, A), where the first argument is the current state, must give the same result.
 - Universality. Everyone agrees.
 - Correlation with proof-of-work. The only way to get a systematic advantage over others is to demonstrate more work.

Probably relevant:

## Incentive to Withhold Blocks

You are a miner and you just found a block. You can broadcast it and (unless you end up losing a block race) you get to collect the block reward. But not keeping it to yourself, at least for a while, might be even more rewarding. In that case, you get to mine on a block that's ahead of the view of the competition. Provided that the block you just found ends up winning anyway, you are effectively tricking the rest of the network into wasting mining power on a stale block. The tricky part is to ensure that the block you found, wins out eventually. This is the setup behind the *selfish mining* attack.

The space of strategies is captured by a graph. Every time you find a block, you can choose to broadcast what you have or to keep extending your lead. Every time the network finds a block, you can choose to broadcast what you have and hope to win the ensuing block race (if you have the longer chain you win it outright), or you can choose to keep it secret and keep extending your lead, or you can drop your secret chain and accept the view of the network.

The profit depends on a) the strategy, which is a sequence of choices in that strategy graph; and on b) the transition probabilities of chance events. These probabilities are related to two quantities $\alpha$, the proportion of the network's mining power that you command, and $\gamma$, the proportion of the rest of the network that mines on your branch in case of a race between blocks of equal height. The probability of finding the next block yourself is $\alpha$. The probability of winning a block race between blocks of equal height is $\alpha + (1-\alpha) \cdot \gamma$. The probability of winning a block race when your chain is longer is 1.

While the maximum profit is difficult to calculate outright, it is possible to reason about it. Clearly, the maximum profit increases monotonically with both $\alpha$ and $\gamma$. Neither current nor proposed fork choice rules affect $\alpha$.

In the **current fork choice rule**, $\gamma$ is limited by network conditions. In case of a race of depth 1, nodes stick to the block they see first. So the first to broadcast the block has the advantage. Assuming the selfish miner runs a small fraction of the nodes on the network, and assuming the network graph is random, the selfish miner may expect to notice the new block halfway into its propagation time. Under the same assumptions, half of the remaining network will see the selfish miner's challenger block first. So the baseline is $\gamma = 1/4$, but is strongly dependent on network topology.

The **proposed fork choice rule** has a two features that affect this parameter and this strategy space.

1. Sibling Races Only

Whenever the parents disagree, the rule decides based on cumulative proof of work. There are no block races deeper than 1 deep. This collapses the strategy graph and makes the probability of winning a deeper block race a deterministic function of the involved branches and not of any fraction of foreign mining power. The difficulty is higher on the branch whose blocks came in closer succession, which is precondition for the attacker to be sitting on a secret branch while the rest of the network is busy catching up. So given that the attacker is sitting on a secret branch, whenever the network catches up, the attacker's branch may be expected to have a larger difficulty. Releasing that branch will win the block race outright, equivalent to $\gamma = 1$.

2. Prefer Highest Difficulty

Whenever the parents agree (and ignoring the more-inputs rule) the rule decides based on cumulative proof of work. The effect is the same, except collapsing the strategy graph at the one level the previous point left open.

Suppose you find a block before the rest of the network did. Why publish it? Based on its timestamp, it will have a higher difficulty than the block the rest of the network will find, because by assumption it finds it later than you found yours. So why publish the block at all? If you wait for the network to find and broadcast a block, and publish only then, you are guaranteed to win the block race. And you will have managed to trick the rest of the network to waste work on a block they did not know was stale.

On this metric, the proposed fork choice rule constitutes a strong degradation.

## Race Win Probability

You are a miner and a new foreign block comes in. You can either switch to the new block immediately, or you can stubbornly stay on the block you were mining on in the hope of finding the nonce and winning the resulting block race.

Whether it is rational to switch or to be stubborn depends on the expected profit. Say you represent a proportion $\alpha$ of the network's mining power, and let $\beta$ be the probability of winning a block race (the titular "race win probability").

If you switch, you win if you find the next block's nonce, which happens with probability $\alpha$. So you may expect a fraction $\alpha$ of the next block's transaction fees and block subsidy.

If you remain stubborn, you win if you guess the right nonce (probability $\alpha$) *and* win the ensuing block race (probability $\beta$). So you may expect a fraction $\alpha \times \beta$ of the *current block's* transaction fees and block subsidy. You might even bump the transaction fees by confirming fee-paying transactions broadcast in the meantime.

Stubbornness is rational if $\alpha \times \beta \times$ *current block's fees + subsidy* > $\alpha \times$ *next block's fees + subsidy*. Whether stubbornness is rational is independent of $\alpha$. If $\beta$, the probability of winning a block race, is big enough then depending on the fees and subsidy alone, all rational miners may reject the new incoming block and prefer to be stubborn. In this case the network depends on the presence of irrational miners. The way to solve this problem is to engineer $\beta$ to be small.

In the **current fork choice rule**, $\beta$ is determined by network conditions and decreases with time. Suppose a foreign miner has found a new block A and you find a competing block B of the same height and parent and broadcast it after a delay. In this case, the rest of the network has had that intervening time to accept A. If the broadcast delay of block B is much larger than the propagation delays for block A, then all of the network except you has accepted A as the new tip, and when they see block B they will reject it because A was first. So in order to win the block race, you are left entirely to your own resources, which is concretely the hash power with which you can mine on top of B. So $\beta$ decreases with time and approaches $\alpha$ in the limit.

The **proposed fork choice rule** has two features that affect this dynamic.

1. Prefer the block with more inputs.

By adding one input I can bump $\beta$ to 1. After winning the block race, the competition can do the same trick. This dynamic leads to an indefinite escalation where every current leader is challenged by a newcomer with one extra input.

One factor that could put a stop to this escalation is the cost of confirming one extra input, which is proof work, but the same proof work is required to continue mining on top of the current leader. So this one factor hurts both leader and challenger equally, and fails to stop the indefinite escalation.

The limiting factor is the number of inputs that fit in a block. A block can no longer be challenged if any block that has one more input violates the size limit. The dynamic of escalating challenges tends towards this point. Legitimate output-producing transactions are left unconfirmed in favor bogus input-consuming transactions.

On this metric, the more-inputs rule constitutes a clear degradation. 

2. Prefer the block with the larger difficulty.

Ignore the more-inputs rule. How does the preference for larger difficulty blocks affect $\beta$?

A block is rejected if the claimed difficulty is different from the one that is computed via

```rust
            difficulty_control(
                self.header().timestamp,
                previous_block.header().timestamp,
                previous_block.header().difficulty,
                network.target_block_interval(),
                previous_block.header().height,
            )
```

Note that the current block's *timestamp* is a parameter in this calculation. So I can engineer an epsilon higher difficulty – enough to win the fork choice rule with probability $\beta = 1$ – by setting the timestamp to epsilon-prime before the current leader that I am trying to challenge. There are two limiting factors pushing back against indefinite escalation:

 - a) I now have to solve a proof-of-work puzzle at difficulty epsilon-double-prime higher.
 - b) There is a hard limit on the timestamp. The block's timestamp cannot be within `Network::minimum_block_time` of its predecessor's.

Rational miners have an incentive to declare smaller timestamps than whatever 'now' is at the time of the guess. At some point the difference between now and the declared timestamp is counter-balanced by the extra cost of doing proof-of-work at difficulty plus epsilon-double-prime. So while there may be an equilibrium spoof time that does not coincide with the bound `previous_block.header().timestamp` + `Network::minimum_block_time`, taking that extremal point translates into a cost of solving the proof-of-work puzzle at a concrete difficulty, which in turn bounds $\beta$.

The relative error is `(observed block time - target block interval) / (target block interval) > -1` so the lesser clamp has no effect. Continuing with -1 for simplicity gives a PID adjustment of $1 + 2^{-4}$, which is multiplied by the old difficulty. So finding a block at a maximally time-spoofed difficulty costs a factor $1 / (1+2^{-4})$ more work, and $\beta$ is lower bounded by this number.

The good news is that once the equilibrium is reached (whether it coincides with `previous_block.header().timestamp` + `Network::minimum_block_time` or not), it is no longer rational to continue challenging the leader. Moreover, rational miners defend against rational challengers by setting the block timestamp to the optimal value from the start. So in the end, in an environment dominated by rational miners, there is no sequence of block races at the same height; instead there is just dishonest block time reporting.

On this metric, the preference for larger difficulty blocks constitutes a degradation, but one with limited distortion.

# Summary

The current fork choice rule certainly has drawbacks. The subjectivity window is concretely large. The selfish mining parameter baseline $\gamma = 1/4$ is concretely large and gameable through network-level attacks. The race win probability $\beta$ approaches the proportion of mining power under control $\alpha$, which is also concretely large for a race win probability.

The proposed fork choice rule succeeds to improve some metrics, but fails to leave the others intact. It therefore constitutes not a strict improvement, but a trade-off.

Sword-Smith:
I get from your message that there is an argument for a non-zero subjectivity window, namely when it comes to the tip. Since this disincentivizes the self mining attack. 

Also: to simplify things, we can drop the count of inputs, thus simplifying the code. The incentive to include mempool transactions will then have to come only from a) transaction fees and b) from the shortened composition path that including a mempool transaction allows.

In summary:

```rust
    pub fn fork_choice_rule<'a>(current_tip: &'a Self, incoming_block: &'a Self) -> &'a Self {
        // If blocks have different parent, prefer the one with most cumulative
        // proof of work.
        if current_tip.header().prev_block_digest != incoming_block.header().prev_block_digest
            && incoming_block.header().cumulative_proof_of_work
                > current_tip.header().cumulative_proof_of_work
        {
            return incoming_block;
        }

        // Current tip has more cumulative proof-of-work, or the candidates have
        // the same parent in which case we prefer the one that came first since
        // this disincentivizes a selfish mining attack.
        current_tip
    }
```

This candidate is whenceforth called "alpha".

Sword-Smith:
After a discussion in the office, we came to a 3rd option.

The reason to look at the height as opposed to the parent is that we believe is disincentivizes a selfish mining attack where a block of height n (`blokc_n_2`) is mined on by a mining pool while the rest of the network considers `block_n_1` the tip. `block_n_2` was discovered after `block_n_1` but `block_n_2` has a (marginally) higher cumulative proof of work, because of different timestamps. In this case the rest of the network would need to find *two* blocks before the selfish mining pool finds one, in order to win the block race. By considering heights instead of parent blocks, we believe this incentive is removed.

The drawback of this solution is that it might incentivize *longer* chain splits than one block, since the objective/in-the-limit selection rule of highest cumulative PoW is only used when the heights differ, not when the parents differ -- thus increasing the subjectivity window. However, the probability of succeeding in such an attack decreases exponentially over the chain split height.

```rust
    pub fn fork_choice_rule<'a>(current_tip: &'a Self, incoming_block: &'a Self) -> &'a Self {
        // If blocks have different height, prefer the one with most cumulative
        // proof of work.
        if current_tip.header().height != incoming_block.header().height
            && incoming_block.header().cumulative_proof_of_work
                > current_tip.header().cumulative_proof_of_work
        {
            return incoming_block;
        }

        // Current tip has more cumulative proof-of-work, or the candidates have
        // the same height in which case we prefer the one that came first since
        // this disincentivizes a selfish mining attack.
        current_tip
    }
```

This candidate is whenceforth called "beta".

Sword-Smith:
From a discussion in the office today, I think it's a tie between "alpha" and "beta". A tie-breaker could be that "beta" is closer to the current implementation.

aszepieniec:
I favor candidate beta for three reasons:
 - Perhaps not the logic but the resulting dynamic is similar to what is happening in Bitcoin because there two equal-height forks that split on the same side of the latest epoch boundary as the tips, also have the same cumulative difficulty -- thus triggering the first-come-first-serve rule and the ensuing block race.
 - The first-come-first-serve rule, at the cost of launching an ensuing block race, deprives the selfish miner of the benefit of holding a secret chain. Phrased differently, it forces them to release their blocks because otherwise they will be outpaced by the honest network, at least in expectation. The first-come-first-serve rule might or might not make the subjectivity window longer; but it certainly makes the subjectivity *visible*.
 - It is closest to the current fork choice rule, after solving the obvious flicker bug.

Sword-Smith:
Good. Then we agree.

PR #966, opened 2026-08-27
feat: Retire transactions based on announcements

Sword-Smith:
Allow transactions to be retired after a specific timestamp.

May be useful for exchanges and other actors with automated transaction initialization, to guarantee that a transaction will not be mined if it isn't mined within a specified time window.


neptune-consensus/src/block/mod.rs
```diff
@@ -1607,6 +1617,121 @@ pub(crate) mod tests {
...
+        let block1 = invalid_empty_block(&Block::genesis(network), network);
+        let now = block1.header().timestamp;
+        let retirement = block1.body().transaction_kernel.timestamp;
+        let block1 = block_with_announcements(&block1, vec![Announcement::retirement(retirement)]);
+
+        assert_eq!(
+            Err(BlockValidationError::AppendixMissingClaim),
+            block1.solo_validate(now, network).await
+        );
+    }
```
aszepieniec: The test name says a block must be accepted, but I see a block being rejected.

If a test is going to make sense only in the context of some hidden internal mechanic that is not itself being tested, then that mechanic should be documented.

neptune-consensus/src/block/mod.rs
```diff
@@ -1607,6 +1617,121 @@ pub(crate) mod tests {
...
+        );
+
+        let now = block1.header().timestamp;
+        let retired = block1.body().transaction_kernel.timestamp - Timestamp::millis(1);
+        let block1 = block_with_announcements(&block1, vec![Announcement::retirement(retired)]);
+
+        assert_eq!(
+            Err(BlockValidationError::AppendixMissingClaim),
+            block1.solo_validate(now, network).await
+        );
```
aszepieniec: Suggestion: wrap this assert, and its brothers, in a function. Call the function instead of the raw assert. Document the internal mechanic once, above the function definition.

neptune-consensus/src/transaction/announcement.rs
```diff
@@ -15,6 +16,15 @@ use crate::block::pow::LustrationStatus;
...
 use crate::transaction::transparent_input::TransparentInput;
 
+/// Flags to indicate that an announcement marks the associated transaction
+/// as retired (i.e. unmineable) after a certain timestamp.
+///
+/// Ignored prior to the activation of hardfork-delta.
+pub const RETIREMENT_FLAGS: [BFieldElement; 2] = [
+    BFieldElement::new(0x4D434D58585856u64),
+    BFieldElement::new(15725664001572566400u64),
+];
```
aszepieniec: Why are there two flags? Surely one suffices. Where did these concrete values come from?

neptune-consensus/src/transaction/announcement.rs
```diff
@@ -45,6 +55,31 @@ impl Announcement {
...
+    /// Return Some(timestamp) if the transaction holding this announcement is
+    /// set to be retired after a specific timestamp. Return `None` for all
+    /// announcements that do not indicate retirement.
+    pub fn retirement_value(&self) -> Option<Timestamp> {
+        if self.message.len() > 2
+            && self.message[0] == RETIREMENT_FLAGS[0]
+            && self.message[1] == RETIREMENT_FLAGS[1]
+        {
+            Some(Timestamp(self.message[2]))
+        } else {
```
aszepieniec, line 79: What about announcements more than 3 BFieldElements in length?

neptune-consensus/src/transaction/announcement.rs
```diff
@@ -45,6 +55,31 @@ impl Announcement {
...
+        Self::new(vec![RETIREMENT_FLAGS[0], RETIREMENT_FLAGS[1], retirement.0])
+    }
+
+    /// Return Some(timestamp) if the transaction holding this announcement is
+    /// set to be retired after a specific timestamp. Return `None` for all
+    /// announcements that do not indicate retirement.
+    pub fn retirement_value(&self) -> Option<Timestamp> {
+        if self.message.len() > 2
+            && self.message[0] == RETIREMENT_FLAGS[0]
+            && self.message[1] == RETIREMENT_FLAGS[1]
```
aszepieniec: If both flags are checked simultaneously, does that make it into a single flag of type `[BFieldElement; 2]`? In other words, the word "flag**s**" is confusing.

neptune-consensus/src/transaction/transaction_kernel.rs
```diff
@@ -127,6 +127,22 @@ impl From<RemovalRecordListUnpackError> for TransactionConfirmabilityError {
...
+    }
+
+    /// Whether the transaction's earliest retirement falls before the given
+    /// margin ahead of the given time. False for transactions that carry no
+    /// retirement announcement.
+    pub fn retires_within(&self, margin: Timestamp, now: Timestamp) -> bool {
+        self.earliest_retirement()
+            .is_some_and(|retirement| retirement < now + margin)
+    }
+
```
aszepieniec, line 146: Very confusing function. Consider inlining.

aszepieniec: If I give you the notes ahead of the meeting, that means you receive them before not after. The docstring should read "[...] falls before the given time, counted with margin." or something.

aszepieniec: This function exists to provide syntax sugar for use in `mempool.rs`. Here it is confusing, begging the questions:
 - why not pass a range or a time interval, as "within" suggests?
 - why not compute the sum before the function call?

If you want to recycle the syntax sugar, consider defining it in `mempool.rs` near to where it is used.

aszepieniec (approved):
Why two flags?

Otherwise superficial complaints.

neptune-consensus/src/transaction/announcement.rs
aszepieniec: Don't these flags belong in `neptune-primitives/src/announcement_flag.rs`? I notice that `LUSTRATION_FLAG` is singular and lives in `neptune-consensus/src/transaction/transaction_kernel.rs`.

neptune-consensus/src/block/mod.rs
Sword-Smith: I think I'll just go with `assert_ne!(...)` instead. That way the test is not too specific about internal ordering of checks.

That being said: Block validation should stay stable and shallow, and not be a flexible, abstractable, deeply nested, nimble, or DRY function. So assuming a certain ordering of check is more acceptable here than in other functions.

Sword-Smith: Solved by factoring out a function `solo_validate_without_proof`.

neptune-consensus/src/transaction/announcement.rs
Sword-Smith: I think they belong here, in `announcement.rs`. I can see `LUSTRATION_FLAG` also belonging in `announcement.rs`. `AnnouncementFlag` is actually a wallet and UTXO index abstraction, that's not used by the consensus mechanism. Wallets, and by extension the UTXO index, use these to categorize who the announcement is for.

neptune-consensus/src/transaction/announcement.rs
Sword-Smith: Good point! Let's just say it has to be exactly three elements long, longer not allowed.

neptune-consensus/src/transaction/transaction_kernel.rs
Sword-Smith: Agreed. Changing.

New interface:
```rust

    /// Whether the transaction's earliest retirement falls before the given
    /// timestamp.
    ///
    /// False for transactions that carry no retirement announcement.
    pub fn retires_before(&self, threshold: Timestamp) -> bool {
        self.earliest_retirement()
            .is_some_and(|retirement| retirement < threshold)
    }
```

Sword-Smith:
Should the block timestamp or the transaction timestamp be used in the new consensus rule? The transaction timestamp is not allowed to exceed the block timestamp. So the new rule forces `retirement <= tx_timestamp`. And `tx_timestamp <= block_timestamp`. But `tx_timestamp` can be less than `block_timestamp` by up to three days!!

That means that a transaction only becomes unmineable **three days** after it is retired by a real clock. Comparing the transaction timestamp to the **block timestamp** does not have this problem.

Ergo: Changed retirement check to be against that of the block's header's timestamp.

aszepieniec (approved):
I think it is the right approach: at the block level, enforce absence of early retirement flags. Soft fork.

Privacy tradeoff: expiring transactions are distinct from non-retiring transactions. If you know your target uses retirement, this info shrinks the sets of transactions to deanonymize.

Space tradeoff: 2 BFEs per retiring transactions. All retirement flags persist across Merges into the block transaction. So retiring transactions take up strictly more space than non-retiring transactions do and there's nothing you can do about that. But next to the on-chain announcement, in case of a payment to generation address, it's a small surcharge.

PR #967, opened 2026-08-31
Npt 198 in r11

Sword-Smith:
Expand `Chunk` encoding in a backwards compatible way. Allows for more set indices in a packed `Chunk`.

### DIFF
--- a/neptune-mutator-set/src/removal_record/chunk.rs
+++ b/neptune-mutator-set/src/removal_record/chunk.rs
@@ -808,8 +808,6 @@ mod tests {
             );
         }
 
-        /// See NPT-198 in `r11.md`.
-        ///
         /// An attacker can grind to occupy more than 2^12 indices in one chunk,
         /// thus forcing more than 12 bits to be required for the length
         /// encoding in a packed chunk.
--- a/neptune-mutator-set/src/removal_record/chunk.rs
+++ b/neptune-mutator-set/src/removal_record/chunk.rs
@@ -10,17 +10,25 @@ use thiserror::Error;
 
 use super::super::shared::CHUNK_SIZE;
 
-/// "Hard" max on the number of elements in a packed [`Chunk`].
-/// Based on the Chernoff bound, the probability of finding a [`Chunk`] with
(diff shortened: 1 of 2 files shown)


neptune-mutator-set/src/removal_record/chunk.rs
```diff
@@ -224,23 +273,41 @@ impl Chunk {
 
         let mut current = 0_u64;
         let mut width = 0_usize;
-        let indicated_length = (self.relative_indices[0] >> 20) & ((1 << 12) - 1);
+
+        // The first u12 is the length, unless its 12th bit is set, in which
+        // case the length spans the first two u12s. See [`Self::pack`].
+        let first_length_element = (self.relative_indices[0] >> 20) & ((1 << 12) - 1);
```
aszepieniec, line 279: What's the purpose of the `& ((1 << 12) - 1)`?

neptune-mutator-set/src/removal_record/chunk.rs
```diff
@@ -722,6 +808,160 @@ mod tests {
             );
         }
 
+        /// See NPT-198 in `r11.md`.
```
aszepieniec: IMO if code must link to some external document, that external document should be publicly accessible. I don't think that's the case here.

aszepieniec (approved):
You say it is backwards-compatible, but I don't think it's very hard to create a special instance that decodes/unpacks differently. I think the point is that these special instances are so rare in practice that it's safe to ignore them. How about *probabilistically* backwards-compatible?

Sword-Smith:
There's a uniqueness check added to the length indication of a packed `Chunk` that's not backwards compatible and could introduce a chain split. It was previously in a separate commit but when you merged this, you squashed the commits. I'll have to revert that.

Sword-Smith:
> There's a uniqueness check added to the length indication of a packed `Chunk` that's not backwards compatible and could introduce a chain split. It was previously in a separate commit but when you merged this, you squashed the commits. I'll have to revert that.

I fixed this problem in a follow-up commit, 1fb98797d9994ce6521cbcd0bd2bda2115832128.

So green flag from my side now.

PR #968, opened 2026-09-10
Thv/libp2p only ban observed addresses

Sword-Smith:


aszepieniec:
[refactor(peer_loop): Persist both standing decreases, and increases](https://github.com/Neptune-Crypto/neptune-core/pull/968/changes/e3cb817194463ddfc0ce7f203f3e1ee43792b3c3)
Prior to this commit, only standing decreases were persisted. Now
standing increases can also be persisted such that a well-behaving peer
can work its way out of a bad standing.

[fix: Only ban observed IPs, not declared](https://github.com/Neptune-Crypto/neptune-core/pull/968/changes/97ac1bc6eeb1981925df5698843c66642a84ed4c)
Only ban those IPs that the transport layer (TCP/UDP) sees, not the IPs
that a peer reports that it can be reached on, since the latter can be
faked and are completely unverified.

[refactor(network): Add 48-hour half life to negative peer standings](https://github.com/Neptune-Crypto/neptune-core/pull/968/changes/6af2b468a28db820d1d965c25b0e3a332b85b556)
Prevent a ban from lasting forever, instead halve negative sanctions
every 48 hours such that no one stays permanently banned.

neptune-core/src/application/network/ban.rs
```diff
@@ -198,6 +234,97 @@ mod tests {
...
+                NegativePeerSanction::DifferentGenesis,
+                SystemTime::now() - ago,
+            )),
+            latest_reward: None,
+            peer_tolerance: i32::from(tolerance),
+        }
+    }
+
+    #[proptest]
+    fn the_two_admission_gates_expire_together(
```
aszepieniec: What's an admission gate and what does it mean for a gate to expire?

aszepieniec:
> Prevent a ban from lasting forever, instead halve negative sanctions
> every 48 hours such that no one stays permanently banned.

This is a policy change. (Why) is it a good one?

aszepieniec (approved):
LGTM WRT observed IPs.

Making the sanctions impermanent needs motivating.

Sword-Smith:
> > Prevent a ban from lasting forever, instead halve negative sanctions
> > every 48 hours such that no one stays permanently banned.
> 
> This is a policy change. (Why) is it a good one?

Three reasons: 
1. On a natted network other parties can block an IP. 
2. We've had many false positives on bad peer behavior. This change will automatically clear such false positives.
3. Avoiding bad peers is mostly done to avoid DoS attacks. Adding a halflife to the negative standing means that the peer banning still achieves its purpose: Avoiding DoS.

~~But if you're worried about the negative effects on this, I'm happy to increase the half life from two days to a week.~~


Edit: After discussion in office, we agreed to keep 48 hours as the negative sanction half life.

PR #969, opened 2026-09-10
Thv/libp2p flood protection

Sword-Smith:


aszepieniec:
Conceptual approval on doing the enforcement in-crate in `source_limits.rs` instead of outsourcing it to Libp2p. Libp2p does not offer fine-grained feature support, is brittle, out of sync with its own version, and can be changed in a future version. So you don't really know what you get if you use libp2p for flood protection -- tempting though as it is.

aszepieniec:
Does 4 connections per IP not limit the number of relays passing through 1 IP to 4? Is that a worthwhile tradeoff?

neptune-core/src/application/network/actor.rs
```diff
@@ -406,6 +411,11 @@ impl NetworkActor {
                 let upnp = libp2p::upnp::tokio::Behaviour::default();
 
                 NetworkStack {
+                    connection_limits: libp2p::connection_limits::Behaviour::new(
+                        libp2p::connection_limits::ConnectionLimits::default()
+                            .with_max_pending_incoming(Some(max_pending_incoming)),
+                    ),
```
aszepieniec, line 417: Ahh, `connection_limits` was what I was trying to caution against in my earlier message :-/

neptune-core/src/application/network/source_limits.rs
```diff
@@ -0,0 +1,404 @@
+//! Per-source limits on inbound libp2p connections, enforced before any
+//! cryptography runs.
+//!
+//! Three limits, each keyed on the address a connection actually arrives
+//! from: how many connections one IP may hold, how many one prefix (`/24` for
+//! IPv4, `/64` for IPv6) may hold, and how many attempts one IP may make per
+//! minute.
+//!
+//! Relayed connections have no attributable address and pass untouched; the
+//! global pending-connection cap covers those.
```
aszepieniec: 🥳

neptune-core/src/application/network/source_limits.rs
```diff
@@ -0,0 +1,404 @@
...
+#[derive(Debug, Clone, Copy, Default)]
+pub(crate) struct SourceLimitsConfig {
+    pub(crate) max_per_ip: Option<usize>,
+    pub(crate) max_per_prefix: Option<usize>,
+    pub(crate) max_attempts_per_minute: Option<usize>,
+}
+
+/// Why an inbound connection was refused.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum Exceeded {
```
aszepieniec: I'm registering confusion and I think it is because of the name of this enum. It enumerates possible reasons why an incoming connection is refused. All such reasons have to do with limits being exceeded. (Not surprising for a file named `source_limits.rs`.) I'm not sure how `Exceeded` maps onto that purpose. I would expect a type with this name to wrap a boolean, or the degree of excess.

Suggested alternatives: `RefusalReason` `Limit` `Violation`.

neptune-core/src/application/network/source_limits.rs
```diff
@@ -0,0 +1,404 @@
...
+}
+
+/// The counting behind [`SourceLimits`], kept free of libp2p so that it can be
+/// tested in isolation.
+#[derive(Debug)]
+struct SourceCounts {
+    config: SourceLimitsConfig,
+    by_ip: HashMap<IpAddr, usize>,
+    by_prefix: HashMap<Prefix, usize>,
+    connections: HashMap<ConnectionId, IpAddr>,
```
aszepieniec, line 107: Some connections have multiple `IpAddr`s -- which one are you putting into this dictionary?

neptune-core/src/application/network/source_limits.rs
```diff
@@ -0,0 +1,404 @@
...
+            config,
+            by_ip: HashMap::new(),
+            by_prefix: HashMap::new(),
+            connections: HashMap::new(),
+            attempts: HashMap::new(),
+        }
+    }
+
+    /// Admit a pending connection from `ip`, or say why not. An admitted
+    /// connection counts until [`Self::release`]. Every attempt, admitted or
```
aszepieniec: I'm not sure what this means:

```
 An admitted connection counts until [`Self::release`]
```

aszepieniec (changes requested):
About `connection_limits`: back when I tried I came away with a headache and I was hoping to avoid depending on that crate. Is it possible to achieve what that crate is being used for in `source_limits.rs`?

If I understand correctly, DCUtR does not create a new connection but updates an existing one. The IP is not registered: at the time the connection is established, there is no attributable IP. After DCUtR, there is.

aszepieniec:
> Does 4 connections per IP not limit the number of relays passing through 1 IP to 4? Is that a worthwhile tradeoff?

The answer is no because relayed connections have no attributable IPs. Well done :partying_face:

Issue #946, opened 2026-07-10
Proof upgrading improvements


Sword-Smith:
We currently kick out `ProofCollection`-backed transactions when a new block comes since the transaction initiator is assumed to issue a new transaction. However, I would like to change this behavior for two reasons:
1. It makes the transaction flicker in-and-out of the mempool giving a bad user experience
2. The update part is cheap. Anyone performing a "raise" operation (`ProofCollection`->`SingleProof`) can just follow it up by an "update" operation. So the proof upgraders should keep outdated `ProofCollection` transactions in their mempool.

Two changes are needed:
a. Don't kick out `ProofCollection` transactions when a new block arrives
b. Let the transaction-proof upgrader upgrade unsynced ProofCollections as they can just follow this operation up with a very cheap "update" operation.

PR #975, opened 2026-09-18
feat(mempool): Admit transactions synced to a recent ancestor of the tip

codewordneptune:
**Before:** a transaction is admitted only if its removal records validate against the current tip. When a block lands during proving they often no longer do, and the wallet has to produce its ProofCollection again from scratch.

**After:** a block landing during proving no longer forces a new proof, as long as the transaction is at most 3 blocks behind the tip when it reaches the node.

The node validates such a transaction against the mutator set it was built for and checks that no input was spent since. All other admission rules apply unchanged. A proof upgrader brings it up to date afterwards, as it already does for transactions that fall behind in the mempool. Peers running this change request such transactions when notified of them. Transactions further behind are treated as today, apart from one narrow rule noted below.

## Why

A transaction is tied to the mutator set it was built against, and proving a ProofCollection takes minutes on constrained hardware. When a block lands during proving, the node checks the removal records against its new tip. If that block spent a coin with indices in the inactive part of the Bloom filter, or slid the window across one of the transaction's indices, the chunk dictionaries usually no longer validate and the transaction is refused as not confirmable. The wallet has to prove again from scratch, and can lose the same race again. Today a stale transaction gets in only by luck, when the blocks in between happened to leave its records alone.

## What

The mempool already keeps and upgrades unsynced ProofCollections (#946). What was missing is reliable admission.

The rule does not look at the proof type. It matters for ProofCollections, which wallets take minutes to produce, but a SingleProof-backed transaction built against a recent ancestor is admitted the same way and then needs only the update.

- `RecentMutatorSets` holds the mutator set after the tip and after each of its `MAX_TX_SYNC_DEPTH = 3` nearest ancestors, with the updates between them. It is owned by `LightState` and fed from every new tip. A reorganization reduces it to the new tip.
- `tx_admission::admissible` takes this window instead of the tip accumulator. A standard transaction built against a held ancestor is judged against that mutator set, then its removal records are brought forward through the intervening blocks to confirm no input was spent since. It is admitted unsynced, and the existing upgrader path takes over.
- A transaction built against anything else is judged against the tip, as before, so outside the window nothing changes. Inside the window a transaction is judged against the mutator set it names, not the tip.
- The peer loop requests notified transactions whose mutator set is within the window, instead of only the exact tip.
- Link transactions must still be synced to the tip. A transaction without inputs is only admitted at the tip, since it can never be updated.

<details>
<summary>Changes by file</summary>

- `neptune-mempool/src/recent_mutator_sets.rs` (new): `RecentMutatorSets`, `MAX_TX_SYNC_DEPTH`, `catch_up` and `CatchUpError`, with tests.
- `neptune-mempool/src/tx_admission.rs`: `admissible` takes `&RecentMutatorSets` instead of `&MutatorSetAccumulator`, which is a public API change in this crate. New variant `TxAdmissionError::SpentSinceSync(usize)`. `NotSynced` is also returned for a transaction without inputs that is not at the tip.
- `neptune-mempool/src/lib.rs`: registers the new module.
- `neptune-core/src/state/light_state.rs`: owns the window and feeds it in `update`. Its own `mutator_set_accumulator_after` field is removed; `tip_mutator_set_after()` now reads the window's tip entry, which is computed by the same call.
- `neptune-core/src/state/blockchain_state.rs`: accessor `recent_mutator_sets()`.
- `neptune-core/src/application/loops/peer_loop.rs`: both admission call sites pass the window, the `TransactionNotification` gate uses `contains`, `SpentSinceSync` maps to no sanction, and two tests are added.
- `neptune-core/src/application/json_rpc/server/service.rs`: `submit_transaction_call` passes the window, and `SpentSinceSync` maps to `SubmitTransactionError::NotConfirmable`.

</details>

Policy only, no consensus code is touched. Nodes without this change ignore announcements for such transactions without punishing the sender, so it is safe in a mixed network.

The window is derived from blocks as they become the tip, and the spent check replays those blocks' own records. Admission reads nothing from the archive, so a node holding only the current chain state can do it. Upgrading the transaction afterwards works as it does today.

The depth of 3 is one constant. How often a send falls outside the window depends on how long the wallet takes to prove, since its snapshot is exposed to new blocks for that long. Modelling block arrivals as a Poisson process at the 9.8 minute target interval, the first row is the chance that any block lands during proving, and each other cell is the chance that more blocks land than the depth covers:

| Proving time | 1 min | 3 min | 5 min |
|---|---|---|---|
| A block lands during proving | 9.7% | 26% | 40% |
| Outside the window at depth 1 | 1 in 200 | 3.8% | 9.3% |
| Outside the window at depth 2 | 1 in 6,100 | 1 in 260 | 1.5% |
| Outside the window at depth 3 | 1 in 240,000 | 1 in 3,500 | 1 in 530 |
| Outside the window at depth 4 | 1 in 12 million | 1 in 58,000 | 1 in 5,300 |
| Outside the window at depth 5 | 1 in 700 million | 1 in 1.1 million | 1 in 63,000 |

Outside the window a transaction is back to today's behaviour. Depth 3 is where the returns flatten for a typical proof, while every further block of depth adds window memory and one more block of replay that a peer can force before proof verification. Real block arrivals cluster more than this model assumes, which argues for some margin over depth 2.

## Two points for review

1. **A new kind of rejection is never sanctioned.** `SpentSinceSync` is an input spent by a block after the transaction was built. It happens honestly during block propagation, since the sender may not have seen that block. The cost is that a hostile peer can trigger it repeatedly for free: nothing counts or limits such rejections, as is already the case for `TooOld`, `Retired`, `FutureDated` and `AlreadyKnown`. Unlike those it is not cheap, since the Merkle checks on the removal records have run by then, though not the proof verification. `NotSynced` is likewise unsanctioned, and is now also returned for a transaction without inputs that is not at the tip, at the cost of a hash comparison. Should `SpentSinceSync` carry a sanction after all?
2. **Memory at very high volume.** The window holds each block's removal records unpacked, up to roughly 30 KB per input, outside the mempool's size cap. Negligible today. The oldest entry's records are already dropped since they are never read. Storing the rest packed would shrink it several-fold, but every stale transaction would then pay for unpacking up to three blocks' records, on the path that runs before proof verification and that a peer can trigger at will. That trades bounded memory for work an attacker controls, so it is left out. It can follow if memory proves to matter more.

## Testing

- `neptune-mempool`: 65 tests pass, including nine new ones for the window and an admission test covering the tip, a held ancestor, the fallback outside the window, no inputs, an input spent since, and eviction past the depth. The new tests pass with `MAX_TX_SYNC_DEPTH` set to 1, 3 and 5.
- `neptune-cash`: two new peer-loop tests. One proves a real ProofCollection, advances the tip, and checks the transaction is requested and reaches the main loop. The other checks that an unknown mutator set is ignored.
- `cargo fmt`, `cargo clippy --all-targets -- -D warnings` and `cargo doc` with warnings as errors are clean. The single-proof tests exceed the memory of the machine used, so the full suite is left to CI.

On a running regtest node built from this branch's code, a wallet built a ProofCollection-backed transaction, blocks were mined, and only then was it submitted. Coins were aged at least 160 blocks first so that they carried chunk dictionaries. SingleProof-backed transactions take the same admission path but were not exercised on the live node.

| Scenario | v0.17.0 | This branch |
|---|---|---|
| A block with a spend lands, 1 behind | Refused: `NotConfirmable` | Accepted, raised, updated, mined |
| Three blocks land, one with a spend, 3 behind | not run | Accepted, mined |
| Four empty blocks land, 4 behind, transaction X | Refused | Refused |
| Four empty blocks land, 4 behind, transaction Y | Accepted | Accepted, mined |
| The same coin is spent meanwhile, 1 behind | not run | Refused: `NotConfirmable` |

For the two rows outside the window, the wallet's exact submission was captured and the identical bytes were replayed against both versions on the same chain state, so they show the branch matching upstream there case by case.

Not included: a distinct RPC error for an input spent since the snapshot, which currently maps to the existing `NotConfirmable`.


Sword-Smith (approved):

Sword-Smith:
OK with the main points here. I'll make a few stylistic changes myself and one code change: punish double spending transactions regardless of it being tip-synced or not. With the new decay on negative peer standings, I lean on punishing more often than strictly necessary, rather than too rarely.

Sword-Smith:
Merged through
1348008e4960fb13c6a81e812bc8730011336f7a and 463aacf4b5a263185c0ee6eb917e1632fbcfe58a.

